PR #6, opened 2019-06-21
Move Openshift dependent files

Ethyling:
Move all Openshift files to a temporary directory


Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+FROM golang:1.10 AS builder
 WORKDIR /go/src/github.com/NVIDIA/gpu-operator
 COPY . .
 RUN make build
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM debian:stable-slim
```
RenaudWasTaken, line 6: Followup (commit): Can we have the same architecture as https://github.com/NVIDIA/k8s-device-plugin
So that we have an image based on openshift.

openshift-tmp/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
```
rajatchopra: Call the directory openshift instead of openshift-tmp, I would suggest. It is likely here to stay.

openshift-tmp/README.md
```diff
@@ -0,0 +1,5 @@
+# Openshift
```
rajatchopra, line 1: Call this directory openshift itself, instead of openshift-tmp. It is likely here to stay.

openshift-tmp/manifests/0200_role.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
rajatchopra, line 2: We will need a similar one for kubernetes. Right?

openshift-tmp/manifests/0400_operator.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
+kind: Deployment
```
rajatchopra, line 2: We need the deployment as well, right? Or have I missed its existence outside of openshift-tmp?

PR #41, opened 2020-03-10
Rework on Helm management

guilhem:
- create a chart for nfd as dependency
  - I'm going to push it upstream or on its own repository
- simplify configuration variables
- fix CRD management using special helm directory
- use affinity instead of nodeSelector
- use chart namespace


RenaudWasTaken:
Hello!

Thanks for your contribution! Can you:
- Move your PR to https://gitlab.com/nvidia/kubernetes/gpu-operator (the code is mirrored from there)
- Sign your work: https://help.github.com/en/github/authenticating-to-github/signing-commits

Thanks!

deployments/gpu-operator/values.yaml
```diff
@@ -7,16 +7,52 @@ replicaCount: 1
...
+          - vendor
+
 platform:
   openshift: false
 
 operator:
   repository: nvidia
   image: gpu-operator
-  tag: @OPERATOR_VERSION@
+  tag: 1.0.0
```
RenaudWasTaken, line 34: Can you leave this one to @OPERATOR_VERSION@, we need to change the CI to set it correctly before correcting this

deployments/gpu-operator/templates/serviceaccount.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: special-resource-operator
-  namespace: gpu-operator
+  namespace: {{ $.Release.Namespace }}
```
RenaudWasTaken, line 5: Just to be clear, this means the GPU Operator is by default installed in the default namespace unless the user specifies a namespace and uses `--create-namespace`

RenaudWasTaken:
By the way this is a super cool contribution! Thanks a lot for your work!

deployments/gpu-operator/values.yaml
guilhem: Is there a way to change this? (CI that update chart.yaml for example)
It's really hard to use it with this notation as git

deployments/gpu-operator/templates/serviceaccount.yaml
guilhem: its released in current or specified `--namespace`.
It's the best practice for helm charts

guilhem:
Thank you @RenaudWasTaken 
Done here: https://gitlab.com/nvidia/kubernetes/gpu-operator/-/merge_requests/80/

deployments/gpu-operator/values.yaml
RenaudWasTaken: Yes but let's do this in a followup PR

deployments/gpu-operator/templates/serviceaccount.yaml
RenaudWasTaken: Ok, I'll update the readme then

deployments/gpu-operator/values.yaml
guilhem: Yes of course :)
I update the PR on GitLab to remove the commit which change this.
(I anticipated your comment ;))

RenaudWasTaken:
Closed with 2f2a708974d08ad6eefc439068d3ff309494b5cc

PR #179, opened 2021-04-19
Mssing Docker image for node-feature-discovery from values

dignajar:
I'm using a private repository in my company and I need to set all the Docker images in the Helm values; the one in the PR is missing from the values.

```
Signed-off-by: Diego Najar <dignajar@gmail.com>
```

### DIFF
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -163,6 +163,8 @@ gfd:
   priorityClassName: system-node-critical
 
 node-feature-discovery:
+  image:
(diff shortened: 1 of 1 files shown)


shivamerla:
@dignajar Thanks for the PR. Only default override values for NFD are specified in GPU operator values.yaml. For others, we probably just need to document on what values can be changed with NFD from [here](https://github.com/NVIDIA/gpu-operator/blob/master/deployments/gpu-operator/charts/node-feature-discovery/values.yaml). Otherwise, we will always be missing some required NFD field(eg ImagePullSecrets etc).

PR #555, opened 2023-07-20
Fix reconciliation failure monitor.

montaguethomas:
Previsouly the expr was trying to include the timeframe check. The problem with this is that the last_success timestamp is only updated when the Reconcile method is triggered. It is possible to have no update or triggering of the Reconcile method for over an hour (no changes to ClusterPolicy, DaemonSets, or GPU Nodes). When this happens and a new GPU Node is added, an alert will immediately fire because the last_success timestamp hasn't updated in over an hour and the reconciliation status will change to "0"/"NotReady" as the new GPU is being proessed.

Resolve this by using the timeframe tracking of Alertmananger and only watching for reconcilitation_status not being Ready (1).

Hello!

- [You should sign your work](https://github.com/NVIDIA/gpu-operator/blob/master/CONTRIBUTING.md).

### DIFF
--- a/assets/state-operator-metrics/0400_prometheus_rule_openshift.yaml
+++ b/assets/state-operator-metrics/0400_prometheus_rule_openshift.yaml
@@ -12,8 +12,7 @@ spec:
     - alert: GPUOperatorReconciliationFailed
       expr: |
         gpu_operator_reconciliation_status != 1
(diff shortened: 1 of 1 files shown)


montaguethomas:
oh fun. gitlab. ok.

PR #674, opened 2024-02-28
Allow devicePlugin.config.default to be none.

nikp1172:
This commit allows someone not to set a clusterwide policy for GPUs It will take the default behaviour. Only if a node has selector for a particular devicePlugin config (with a given name). This config will be used

Hello!

- [You should sign your work](https://github.com/NVIDIA/gpu-operator/blob/master/CONTRIBUTING.md).


shivamerla:
@nikp1172 can you submit the PR here instead: https://gitlab.com/nvidia/kubernetes/gpu-operator

PR #721, opened 2024-05-15
Add Github Actions

cdesiniotis:


tests/holodeck.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: holodeck.nvidia.com/v1alpha1
+kind: Environment
+metadata:
+  name: HOLODECK_NAME
+  description: "end-to-end test infrastructure"
+spec:
+  provider: aws
+  auth:
+    keyName: cdesiniotis
```
ArangoGutierrez:
```suggestion
    keyName: cnt-ci
```

tests/holodeck.yaml
```diff
@@ -0,0 +1,25 @@
...
+    privateKey: HOLODECK_PRIVATE_KEY
+  instance:
+    type: g4dn.xlarge
+    region: us-west-1
+    ingressIpRanges:
+      - 0.0.0.0/0
+    image:
+      architecture: amd64
+      imageId: ami-0ce2cb35386fc22e9
```
ArangoGutierrez, line 18: Upcoming `holodeck` release won't require this line :)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,316 @@
...
+          GOLANG_VERSION=$( grep "GOLANG_VERSION ?=" versions.mk )
+          echo "GOLANG_VERSION=${GOLANG_VERSION##GOLANG_VERSION ?= }" >> $GITHUB_ENV
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GOLANG_VERSION }}
+      - run: make validate-helm-values
+
+  ### Golang checks and build ###
+  go-check:
```
ArangoGutierrez, line 73: Agree with Evan comment on splitting this file into multiple files

ArangoGutierrez (changes requested):

tests/holodeck.yaml
ArangoGutierrez: https://github.com/NVIDIA/holodeck/pull/63

cdesiniotis:
> As a general comment, does it make sense to split the actions based on scope. For example in the device plugin we have:
> 
> ```
> .github/workflows
> ├── e2e.yaml
> ├── golang.yaml
> ├── image.yaml
> └── stale.yaml
> 
> 1 directory, 4 files
> ```
> 
> Here `golang.yaml` and `image.yaml` are generally generic enough to reuse across all our components. We can add a separate file for the other validation (linting etc).

In general I agree we should split the actions and align common workflows, like `golang.yaml` and `image.yaml`, across all of our projects. And @ArangoGutierrez I know we discussed your plans for removing some of the current limitations, like not being able to run e2e tests on PRs, in the future. However, are you okay if we keep this PR as is for the time being so that we mimic the existing pipelines on Gitlab as much as possible? I am happy to re-factor this into multiple files and align with other projects once the limitations are addressed.

To prevent e2e tests from running on forks (which will likely fail anyways if AWS creds are not configured on the fork), we can consider adding the below conditional to the e2e jobs:
```
if: github.event.pull_request.head.repo.full_name == "github.com/NVIDIA/gpu-operator"
```

tests/holodeck.yaml
cdesiniotis: Updated.

cdesiniotis: @ArangoGutierrez can you add the AWS_SSH_KEY secret to this repo associated with cnt-ci?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,316 @@
...
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-test-logs
+          path: ./logs/
+          retention-days: 15
+
+  e2e-tests-nvidiadriver:
+    needs: [build-gpu-operator, build-gpu-operator-validator]
+    runs-on: ubuntu-latest
```
ArangoGutierrez, line 268:
```suggestion
    runs-on: ubuntu-latest
    if: ${{ github.event.workflow_run.conclusion == 'success' }} && ${{ github.event.workflow_run.event == 'push' }}
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,316 @@
...
+          VERSION: ${COMMIT_SHORT_SHA}
+          SUBCOMPONENT: validator
+        run: |
+          echo "${VERSION}"
+          make build-${{ matrix.dist }}
+
+  ### e2e tests ###
+  e2e-tests-containerd:
+    needs: [build-gpu-operator, build-gpu-operator-validator]
+    runs-on: ubuntu-latest
```
ArangoGutierrez, line 216:
```suggestion
    runs-on: ubuntu-latest
    if: ${{ github.event.workflow_run.conclusion == 'success' }} && ${{ github.event.workflow_run.event == 'push' }}
```

ArangoGutierrez (changes requested):
Let's add this line to ensure we only run e2e on merge for now

ArangoGutierrez:
@cdesiniotis `holodeck` is fixed, now it fails for real reasons

cdesiniotis:
> @cdesiniotis `holodeck` is fixed, now it fails for real reasons

Thanks. The e2e jobs are failing because the container images for this repo have not been made public yet. I have pinged George to make them public.

tariq1890:
> To prevent e2e tests from running on forks (which will likely fail anyways if AWS creds are not configured on the fork), we can consider adding the below conditional to the e2e jobs:
if: github.event.pull_request.head.repo.full_name == "github.com/NVIDIA/gpu-operator"

I think we can get around this by requiring manual approves by the code owners for e2e github pipelines at all times.

cdesiniotis:
All checks have passed. @ArangoGutierrez @elezar could you take another look?

For now, the Github Actions resemble our Gitlab pipelines. In a follow-up, we can look to restrict when the e2e test jobs run.

tests/holodeck.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: holodeck.nvidia.com/v1alpha1
+kind: Environment
+metadata:
+  name: HOLODECK_NAME
```
tariq1890, line 4: Can we choose a different name here? Perhaps, something along the lines of gpu-operator-e2e-env ?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,316 @@
...
+      - run: helm lint deployments/gpu-operator/
+  validate-csv:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Get Golang version
+        id: vars
+        run: |
+          GOLANG_VERSION=$( grep "GOLANG_VERSION ?=" versions.mk )
```
tariq1890, line 49: Is there no way to reuse this and eliminate the duplication?

tariq1890: I think the use top-level `env` field of the workflow.yaml to set global env vars.

In this case, I would prefer hardcoding the golang version in the top-level `env field`. While automated fetching of golang version is a plus, I feel this would be a good tradeoff for the sake of simplicity

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,316 @@
...
+    strategy:
+      matrix:
+        dist: [ubi8]
+    steps:
+      - uses: actions/checkout@v4
+        name: Check out code
+      - name: Calculate build vars
+        id: vars
+        run: |
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
```
tariq1890, line 133: Can we move this to the top level `env` field too?

Also I prefer `GITHUB_SHORT_SHA` over `COMMIT_SHORT_SHA`

cdesiniotis: We are leveraging shell [paramater expansion ](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) to extract the short sha, so  I don't believe we can define this in the top level `env` field, where we cannot leverage a shell-like syntax.

tests/holodeck.yaml
cdesiniotis: I believe the holodeck action automatically generates a name, but I could be wrong. @ArangoGutierrez what is the recommendation for naming this?

.github/workflows/ci.yaml
cdesiniotis: I am not too familiar with GH actions, so I am not sure if we can fetch the golang version once and reuse that throughout this file. 

It's a tradeoff, but I think I prefer duplicating this logic so that `versions.mk` remains the single source of truth for the Go version.

.github/workflows/ci.yaml
tariq1890: Sure, agreed that the `env` field is not an option.

Can we rename `COMMIT_SHORT_SHA` to `GITHUB_SHORT_SHA`?

cdesiniotis: I would prefer `COMMIT_SHORT_SHA` since the sha is not specific to Github, although I do not have too strong of an opinion. I believe this also aligns with the name we use in other projects.

tariq1890 (approved):

.github/workflows/ci.yaml
elezar: Ideally we would have some shared worklflow definitions that we can use to pull in logic such as this.

tests/holodeck.yaml
elezar: Since this is the same as for the device plugin, let's leave it as is for now -- pending guidance on what this should be.

To clarify, the name is set here for the github action: https://github.com/NVIDIA/holodeck/blob/58fe703797a66510af3598bfca139fbaa46f47fa/cmd/action/ci/ci.go#L58-L69

elezar (approved):
Thanks @cdesiniotis. I'm happy to start with this and iterate once we have the initial kinks ironed out.

PR #735, opened 2024-06-05
Bump NFD to v0.16.2

cdesiniotis:


deployments/gpu-operator/charts/node-feature-discovery/templates/post-delete-job.yaml
```diff
@@ -0,0 +1,94 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "node-feature-discovery.fullname" . }}-prune
+subjects:
+- kind: ServiceAccount
+  name: {{ include "node-feature-discovery.fullname" . }}-prune
+  namespace: {{ include "node-feature-discovery.namespace" .  }}
+---
+apiVersion: batch/v1
+kind: Job
```
cdesiniotis, line 52: From the v0.16.0 release notes:
```
Helm Chart

Now all nodes are cleaned up (feature labels, annotations, extended resources and taints are removed) after uninstalling NFD using a post-delete hook.
```
I suspect this post-delete hook is what is causing our e2e tests to fail. We are hitting a timeout waiting for a condition when uninstalling the GPU Operator.

tariq1890:
I have verified that this changeset fixes the e2e issue

https://github.com/kubernetes-sigs/node-feature-discovery/pull/1746

deployments/gpu-operator/charts/node-feature-discovery/templates/post-delete-job.yaml
tariq1890: This is now fixed in v0.16.1

ArangoGutierrez:
@tariq1890 could you squash the commits, so it only has the one with v0.16.1, but we don't loose @cdesiniotis contribution

tariq1890:
@ArangoGutierrez done. Please check now

ArangoGutierrez:
Looks sane, leaving the approval to @shivamerla

deployments/gpu-operator/charts/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -80,7 +94,7 @@ rules:
   - update
 {{- end }}
 
-{{- if and .Values.gc.enable .Values.gc.rbac.create (or .Values.enableNodeFeatureApi .Values.topologyUpdater.enable) }}
+{{- if and .Values.gc.enable .Values.gc.rbac.create (or (and .Values.featureGates.NodeFeatureAPI .Values.enableNodeFeatureApi) .Values.topologyUpdater.enable) }}
```
shivamerla, line 97: is `featureGates.NodeFeatureAPI` enabled by default? Otherwise we have change in behavior when enableNodeFeatureApi is set.

tariq1890: Yes. See [here](https://github.com/kubernetes-sigs/node-feature-discovery/blob/391865bbb2e486b61fbadfdea287a80a79a315ec/deployment/helm/node-feature-discovery/values.yaml#L14-L13)

deployments/gpu-operator/charts/node-feature-discovery/values.yaml
```diff
@@ -91,17 +97,17 @@ master:
...
+  resources:
+    limits:
+      memory: 4Gi
+    requests:
+      cpu: 100m
+      # You may want to use the same value for `requests.memory` and `limits.memory`. The “requests” value affects scheduling to accommodate pods on nodes.
+      # If there is a large difference between “requests” and “limits” and nodes experience memory pressure, the kernel may invoke
+      # the OOM Killer, even if the memory does not exceed the “limits” threshold. This can cause unexpected pod evictions. Memory
+      # cannot be compressed and once allocated to a pod, it can only be reclaimed by killing the pod.
+      # Natan Yellin 22/09/2022 https://home.robusta.dev/blog/kubernetes-memory-limit
```
shivamerla, line 109: This note and the default limits doesn't make sense. Limit is set at 4Gi, while request is 128Mi. Does that mean by default evictions are likely under memory pressure?

shivamerla:
@tariq1890 please audit memory requests/limits again to verify they work with scale cluster. If any requests/limits are reduced those are the ones we need to audit first.

deployments/gpu-operator/charts/node-feature-discovery/values.yaml
tariq1890: @ArangoGutierrez

tariq1890: As far as I am aware, when you have identical requests and limits, the Kubernetes scheduler prioritises the pod at schedule-time. This may be a good reason to encourage that practice and have defaults that reflect this. 
However,  It is difficult to provide sane defaults here as cluster size varies greatly in the user base. 

I think it would be consistent for us to provide default requests and limits that are consistent with the operator resource requests and limits. Perhaps, we can include those defaults in the gpu-operator chart values.yaml or just copy over the resource constraints found in NFD chart values over to the gpu-operator chart values

tariq1890: The same basis with which we provided the gpu-operator resource constraints, we use to determine the default resource constraints of NFD:

```
  resources:
    limits:
      cpu: 500m
      memory: 350Mi
    requests:
      cpu: 200m
      memory: 100Mi

```

ArangoGutierrez: is worth mention that this is just for `nfd-master` , `nfd-worker` is:
```yaml
  resources:
    limits:
      memory: 60Mi
    requests:
      cpu: 50m
      memory: 40Mi
```

ArangoGutierrez: maybe we could fine tune the NFD limits for the GPU Operator, but the master has a peak mem consumption at start time, and then it will not go over 100mi

ArangoGutierrez (changes requested):
v0.16.2 is out

cdesiniotis:
> v0.16.2 is out

Done.

deployments/gpu-operator/charts/node-feature-discovery/values.yaml
shivamerla: if we can setup priorityclassname for this, then the eviction problem can be avoided. Lets merge this and add a change to set priorityclass as "system-node-critical"

shivamerla (approved):

PR #738, opened 2024-06-06
Rename master to main

elezar:


cdesiniotis (approved):
Thanks @elezar this lgtm. After merging this, is the only remaining step to create a `main` branch and configure it as the new default branch?

PR #743, opened 2024-06-07
Add Env valueFrom function

mayooot:
The purpose of this pr is to provide `valueFrom` functionality for CRD, like: [environment-variable-expose-pod-information](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)

The main change is to the CRD definition under `api/nvidia/xxx_types.go`, from the custom struct `EnvVar` to `corev1.EnvVar`.

### Why did you add this feature?
We modified the [k8s-device-plugin](https://github.com/NVIDIA/k8s-device-plugin) code to support multiple gpu spec via valueFrom. And for the [gpu-operator](https://github.com/NVIDIA/gpu-operator) modification, we've tested it ourselves to make sure it's not a problem.

The main reason for this is that we believe that the EnvVar struct should be consistent with the K8s API instead of using a custom struct.


shivamerla:
@mayooot one of the primary reason was that `ClusterPolicy` CRD is already overloaded with many operands and close to 256KB limit. We cannot upgrade operator anymore after this limit. We need to get this information by other means and handle the translation in the code while updating the spec.

mayooot:
> @mayooot one of the primary reason was that `ClusterPolicy` CRD is already overloaded with many operands and close to 256KB limit. We cannot upgrade operator anymore after this limit. We need to get this information by other means and handle the translation in the code while updating the spec.

@shivamerla Hi, does this function support that? If yes, I can continue working on it :)

PR #746, opened 2024-06-10
[OCP] restrict RBAC perms of gpu-operator in OLM bundle

tariq1890:
This PR does the following:

- Remove the wildcard perms found in the ClusterRole defined in the OpenShift OLM bundle
- Remove the unnecessary Role as gpu-operator only needs the ClusterRole
- Parity with the helm chart 

Verified and tested on an OpenShift cluster

### DIFF
--- a/deployments/gpu-operator/templates/clusterrole.yaml
+++ b/deployments/gpu-operator/templates/clusterrole.yaml
@@ -147,10 +147,8 @@ rules:
   - list
   - watch
 - apiGroups:
-  - ""
   - coordination.k8s.io
   resources:
-  - configmaps
   - leases
   verbs:
   - get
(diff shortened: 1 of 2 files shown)


cdesiniotis (approved):

deployments/gpu-operator/templates/clusterrole.yaml
```diff
@@ -147,10 +147,8 @@ rules:
   - list
   - watch
 - apiGroups:
-  - ""
   - coordination.k8s.io
   resources:
-  - configmaps
```
shivamerla, line 153: why was this removed? did we add this RBAC somewhere else?

tariq1890: Yeah, this is a duplicate. The configMaps permission already exists in the manifest file

shivamerla (approved):

PR #747, opened 2024-06-10
Add hostRoot, driverInstallDir, and devPath options to ClusterPolicy

cdesiniotis:


controllers/object_controls.go
```diff
@@ -1213,6 +1265,36 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
+			},
+		}
+		obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, devVolume)
+
+		devVolumeMount := corev1.VolumeMount{
+			Name:      "custom-dev",
+			MountPath: "/dev-root/dev",
+		}
+		obj.Spec.Template.Spec.Containers[0].VolumeMounts = append(obj.Spec.Template.Spec.Containers[0].VolumeMounts, devVolumeMount)
+	}
```
cdesiniotis: This code only adds a volumeMount for `hostPath.DevPath` iff the path is not `hostRoot/dev` or `driverInstallDir/dev`. In other words, we add the volumeMount if the `hostPath.DevPath` is not already available inside the container.

I am wondering if it is simpler to always mount `hostPath.DevPath` to `/dev-root/dev` in the container to eliminate the need for this code. If `hostPath.DevPath` is not the empty string, we simply transform the hostPath for the volume.

deployments/gpu-operator/values.yaml
```diff
@@ -20,6 +20,21 @@ sandboxWorkloads:
...
+  # Examples include the MIG Manager and Toolkit Container which may need to
+  # stop, start, or restart systemd services
+  root: "/"
+
+  # driverInstallDir represents the root at which driver files including libraries,
+  # config files, and executables can be found.
+  driverInstallDir: "/run/nvidia/driver"
+
+  # devPath represents the path at which the NVIDIA device nodes can be found.
+  devPath: ""
```
cdesiniotis: Question -- should we explicitly set this to `/run/nvidia/driver/dev` in the values file?

validator/main.go
```diff
@@ -614,24 +664,67 @@ func assertDriverContainerReady(silent bool) error {
...
+func validateHostDriver(silent bool) error {
+	log.Info("Attempting to validate a pre-installed driver on the host")
+	command := "chroot"
+	args := []string{"/host", "nvidia-smi"}
+
+	return runCommand(command, args, silent)
+}
+
+func validateDriverContainer(silent bool) error {
+	if driverManagedByOperator {
```
cdesiniotis, line 708: Some thoughts on this concept of a "managed" driver.

Currently when the user sets `driver.enabled=false` in Clusterpolicy, I set `DRIVER_MANAGED_BY_GPU_OPERATOR` in the validator to indicate that the Operator is not managing the driver, and thus, do not wait on the `.driver-ctr-ready` file that our driver container creates as part of its readiness probe (see this MR for context on why we added this readiness file https://gitlab.com/nvidia/kubernetes/gpu-operator/-/merge_requests/580).

I think a better approach would be as follows -- in the driver validator, check if there is any driver pod running on the node. We can use the standard `app.kubernetes.io/component: nvidia-driver` label that all of our driver pods have. If there is no pod, then skip waiting on this status file.

validator/main.go
```diff
@@ -675,35 +769,65 @@ func (d *Driver) validate() error {
...
+		nvidiaDevRoot = devRootFlag
+
+		devRootCtrPath = driverRootCtrPath
+		if devRootFlag == hostRootFlag {
+			devRootCtrPath = "/host"
+		} else if devRootFlag != "" && devRootFlag != driverRootFlag {
+			// a custom dev root, which is not equal to host root or driver root,
+			// will be mounted in the operand container at /dev-root
+			devRootCtrPath = "/dev-root"
+		}
```
cdesiniotis: If we decided to always mount the devPath to `/dev-root/dev` inside of our operand containers, then this code block could simplify to 
```
devRootCtrPath = /dev-root
```

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -144,6 +146,23 @@ type OperatorSpec struct {
...
+	// Examples include the MIG Manager and Toolkit Container which may need to
+	// stop, start, or restart systemd services.
+	Root string `json:"root,omitempty"`
+
+	// DriverInstallDir represents the root at which driver files including libraries,
+	// config files, and executables can be found.
+	DriverInstallDir string `json:"driverInstallDir,omitempty"`
+
+	// DevPath represents the path at which the NVIDIA device nodes can be found.
+	DevPath string `json:"devPath,omitempty"`
```
tariq1890: I prefer to be more explicit here
```suggestion
	DeviceNodesPath string `json:"devPath,omitempty"`
```

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -144,6 +146,23 @@ type OperatorSpec struct {
...
 }
 
+// HostPathsSpec defines various paths on the host needed by GPU Operator components
+type HostPathsSpec struct {
+	// Root represents the path to the root filesystem of the host.
+	// This is used by components that need to interact with the host filesystem
+	// and as such this must be a chroot-able filesystem.
+	// Examples include the MIG Manager and Toolkit Container which may need to
+	// stop, start, or restart systemd services.
+	Root string `json:"root,omitempty"`
```
tariq1890: or `RootFileSystem`. It would be more consistent if all the fields are as explicit and suggestive as `DriverInstallDir`
```suggestion
	RootFS string `json:"rootFS,omitempty"`
```

controllers/object_controls.go
```diff
@@ -773,6 +783,48 @@ func applyCommonDaemonsetConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+// Transform the host-root volume if required for the Daemonset passed as input
+func transformHostRootVolume(obj *appsv1.DaemonSet, hostRoot string) {
+	if hostRoot == "" || hostRoot == "/" {
+		return
+	}
+
+	containsHostRootVolume := false
+	for index, volume := range obj.Spec.Template.Spec.Volumes {
+		if volume.Name == "host-root" {
+			obj.Spec.Template.Spec.Volumes[index].HostPath.Path = hostRoot
```
tariq1890: Let's move `obj.Spec.Template.Spec` to a variable and use that instead to reduce redundancy. Makes it more readable too

controllers/object_controls.go
tariq1890: This logic is fine IMO. However, my first preference is not to expose the `devPath`. So far, we only know of two possible deviceNodePath values: the host root or the driver root. If the need arises to set a completely custom device node path, then we expose the new field and add in the extra logic

controllers/object_controls.go
```diff
@@ -773,6 +783,48 @@ func applyCommonDaemonsetConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
 	return nil
 }
 
+// Transform the host-root volume if required for the Daemonset passed as input
+func transformHostRootVolume(obj *appsv1.DaemonSet, hostRoot string) {
```
tariq1890: Let's start adding test cases for `object_controls.go`. A test case that covers this new method would be a good start

controllers/object_controls.go
```diff
@@ -1213,6 +1265,36 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
+				},
+			},
+		}
+		obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, devVolume)
+
+		devVolumeMount := corev1.VolumeMount{
+			Name:      "custom-dev",
+			MountPath: "/dev-root/dev",
+		}
+		obj.Spec.Template.Spec.Containers[0].VolumeMounts = append(obj.Spec.Template.Spec.Containers[0].VolumeMounts, devVolumeMount)
```
tariq1890: Let's query the container by name instead of a hardcoded index

controllers/object_controls.go
```diff
@@ -1674,6 +1785,36 @@ func TransformMIGManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
```
tariq1890: same as previous

controllers/object_controls.go
```diff
@@ -2106,6 +2256,21 @@ func TransformValidatorComponent(config *gpuv1.ClusterPolicySpec, podSpec *corev
 					setContainerEnv(&(podSpec.InitContainers[i]), env.Name, env.Value)
 				}
 			}
+			setContainerEnv(&(podSpec.InitContainers[i]), DriverManagedByOperatorEnvName, fmt.Sprintf("%v", config.Driver.IsEnabled()))
```
```suggestion
			setContainerEnv(&(podSpec.InitContainers[i]), DriverManagedByOperatorEnvName, fmt.Sprintf("%t", config.Driver.IsEnabled()))
```

validator/main.go
tariq1890: Agreed, it is simpler. I guess it doesn't matter what the mount path is as long as the path to be mounted remains customisable

api/nvidia/v1/clusterpolicy_types.go
elezar: The issue is that this is not the `DeviceNotesPath`. It is the path to `/dev`. `DevPath` is more accurate in this case.

assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -41,7 +41,7 @@ spec:
           seLinuxOptions:
             level: "s0"
         volumeMounts:
-          - name: driver-install-path
+          - name: driver-root
```
elezar: Question: We're calling this `driver-root` because it's always the driver container?

assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -41,7 +41,7 @@ spec:
...
             mountPath: /run/nvidia/driver
```
elezar, line 49: Question (and maybe out of scope for this PR): Does it make sense to set this to `/driver-root` in the container instead?

assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -78,13 +78,17 @@ spec:
             readOnly: true
             mountPath: /bin/entrypoint.sh
             subPath: entrypoint.sh
-          - name: nvidia-run-path
-            mountPath: /run/nvidia
-            mountPropagation: Bidirectional
+          - name: toolkit-pid-file
+            mountPath: /run/nvidia/toolkit.pid
```
elezar: Does this still work as expected if the toolkit container doesn't create this file? Can the toolkit get the exclusive lock it requires?

assets/state-device-plugin/0400_configmap.yaml
```diff
@@ -9,28 +9,15 @@ data:
...
-        break
-      elif [[ -f /run/nvidia/validations/driver-ready ]]; then
-        driver_root=/run/nvidia/driver
-        container_driver_root=$driver_root
-        break
-      else
-        echo "waiting for the driver validations to be ready..."
-        sleep 5
-      fi
+    until [[ -f /run/nvidia/validations/driver-ready ]]
```
elezar, line 12: Question: Why don't we have the same logic in the toolkit container?

assets/state-device-plugin/0500_daemonset.yaml
```diff
@@ -141,10 +143,14 @@ spec:
...
-        - name: run-nvidia
+        - name: run-nvidia-validations
           hostPath:
-            path: "/run/nvidia"
-            type: Directory
+            path: "/run/nvidia/validations"
+            type: DirectoryOrCreate
+        - name: driver-root
+          hostPath:
+            path: "/run/nvidia/driver"
```
elezar, line 152: Is this modified by the operator and set to the driver install path?

controllers/object_controls.go
elezar: Since we're not modifying `Volumes` here, should this not just be:
```suggestion
			volume.HostPath.Path = hostRoot
```
at which point we don't need an explicit variable.

controllers/object_controls.go
```diff
@@ -1213,6 +1265,36 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
+		for index, volume := range obj.Spec.Template.Spec.Volumes {
+			if volume.Name == "driver-root" {
+				obj.Spec.Template.Spec.Volumes[index].HostPath.Path = config.HostPaths.DriverInstallDir
+			}
+		}
+	}
+
+	// If a custom devPath is configured which is not equal to 'hostRoot/dev' or 'driverInstallDir/dev', make it available
+	// to the toolkit container under '/dev-root/dev'
+	devRoot := filepath.Dir(config.HostPaths.DevPath)
```
elezar: Is `DevPath` `/` or `/dev`? Do we validate that it ends in `/dev`?

controllers/object_controls.go
```diff
@@ -1213,6 +1265,36 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
+	if config.HostPaths.DevPath != "" && devRoot != config.HostPaths.DriverInstallDir && devRoot != config.HostPaths.Root {
+		devVolume := corev1.Volume{
+			Name: "custom-dev",
```
elezar: Why not call this `dev-path` instead of `custom-dev`?

controllers/object_controls.go
```diff
@@ -3365,6 +3530,21 @@ func transformValidationInitContainer(obj *appsv1.DaemonSet, config *gpuv1.Clust
 					setContainerEnv(&(obj.Spec.Template.Spec.InitContainers[i]), env.Name, env.Value)
 				}
 			}
+			setContainerEnv(&(obj.Spec.Template.Spec.InitContainers[i]), DriverManagedByOperatorEnvName, fmt.Sprintf("%v", config.Driver.IsEnabled()))
+
+			// Pass env if custom driver and/or dev roots are configured with the operator
+			if config.HostPaths.DriverInstallDir != "" && config.HostPaths.DriverInstallDir != DefaultDriverInstallDir {
+				setContainerEnv(&(obj.Spec.Template.Spec.InitContainers[i]), "DRIVER_ROOT", config.HostPaths.DriverInstallDir)
```
elezar: Is `DRIVER_ROOT` the env that we're using? Should this be a constant?

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -12,6 +12,10 @@ metadata:
     "helm.sh/resource-policy": keep
   {{- end }}
 spec:
+  hostPaths:
+    root: {{ .Values.hostPaths.root | default "/" }}
+    driverInstallDir: {{ .Values.hostPaths.driverInstallDir | default "/run/nvidia/driver" }}
+    devPath: {{ .Values.hostPaths.devPath | default "" }}
```
elezar: Should the default be `/run/nvidia/driver/dev` to be explicit?

deployments/gpu-operator/values.yaml
elezar: That may make it clearer, but would mean that more variables need to be set when we change the "driver root"? What would the documentation look like for these options (assuming our two or three cases that we support).

validator/main.go
elezar: Can we guarantee that the driver validator will be started after the driver pods have been created? Are they not started simultaneously?

validator/main.go
```diff
@@ -614,24 +664,67 @@ func assertDriverContainerReady(silent bool) error {
...
+			return fmt.Errorf("failed to locate nvidia-smi: %w", err)
+		}
+		cmd := exec.Command(nvidiaSMIPath)
+		// In order for nvidia-smi to run, we need to update LD_PRELOAD to include the path to libnvidia-ml.so.1.
+		cmd.Env = setOrOverrideEnvvar(os.Environ(), "LD_PRELOAD", prependPathListEnvvar("LD_PRELOAD", driverLibraryPath))
+		if !silent {
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+		}
+		return cmd.Run()
```
elezar, line 734: This is not running in a `chroot`. Does this mean that we will end up creating the device nodes at `/dev` in the container. Is the `devPath` ever applicable in this case?

validator/main.go
```diff
@@ -641,19 +734,20 @@ func (d *Driver) validate() error {
...
-	}
-
 	driverRoot, isHostDriver, err := d.runValidation(false)
 	if err != nil {
 		log.Error("driver is not ready")
 		return err
 	}
 
-	if !disableDevCharSymlinkCreation {
+	if disableDevCharSymlinkCreation { //nolint:gocritic
```
elezar: Let's rather factor this into a function.

validator/main.go
elezar: There are two aspects here. The first is giving operand containers access to GPUs -- as is done by the NVIDIA Container Toolkit. This will always inject the device nodes at `/dev` since this is where the driver expects them.

The second is for CDI spec generation which accepts an explicit dev root -- in which case a fixed path is fine.

validator/main.go
```diff
@@ -675,35 +769,65 @@ func (d *Driver) validate() error {
...
+			devRootCtrPath = "/dev-root"
+		}
 	}
 
+	statusFileContent := strings.Join([]string{
+		fmt.Sprintf("NVIDIA_DRIVER_ROOT=%s", nvidiaDriverRoot),
+		fmt.Sprintf("DRIVER_ROOT_CTR_PATH=%s", driverRootCtrPath),
+		fmt.Sprintf("IS_HOST_DRIVER=%v", isHostDriver),
+		fmt.Sprintf("NVIDIA_DEV_ROOT=%s", nvidiaDevRoot),
+		fmt.Sprintf("DEV_ROOT_CTR_PATH=%s", devRootCtrPath),
```
elezar: Nit: Can we reorder this so that `IS_HOST_DRIVER` is first?

validator/main.go
```diff
@@ -675,35 +769,65 @@ func (d *Driver) validate() error {
...
-	return nil
+	return createStatusFileWithContent(outputDirFlag+"/"+driverStatusFile, statusFileContent)
 }
 
 // createDevCharSymlinks creates symlinks in /host-dev-char that point to all possible NVIDIA devices nodes.
 func createDevCharSymlinks(driverRoot string, isHostDriver bool) error {
 	// If the host driver is being used, we rely on the fact that we are running a privileged container and as such
 	// have access to /dev
-	devRoot := driverRoot
+	devRoot := devRootFlag
```
elezar: let's add an argument instead. We really should not be reading flags at this level.

validator/main.go
```diff
@@ -121,6 +122,10 @@ var (
 	metricsPort                   int
 	defaultGPUWorkloadConfigFlag  string
 	disableDevCharSymlinkCreation bool
+	driverManagedByOperator       bool
+	hostRootFlag                  string
+	driverRootFlag                string
+	devRootFlag                   string
```
elezar: Instead of separate flags how about the following:
```
type options struct {

	type hostPaths struct {
		root string
		driverRoot string
		devRoot string
	}
}
```
(or even a named struct for the `hostPaths`).

If we don't want to update the functions signatures to pass options instead, then we can create a global variable:
```
var hostPaths hostPathsType{}
```
and refer to this when setting the command line arguments.

assets/state-device-plugin/0500_daemonset.yaml
elezar: Update: Yes, this seems to be the case.

controllers/object_controls.go
```diff
@@ -712,6 +718,10 @@ func preProcessDaemonSet(obj *appsv1.DaemonSet, n ClusterPolicyController) error
 		return err
 	}
 
+	// transform the host-root and host-dev-char volumes if a custom host root is configured with the operator
+	transformHostRootVolume(obj, n.singleton.Spec.HostPaths.Root)
+	transformHostDevCharVolume(obj, n.singleton.Spec.HostPaths.Root)
```
elezar: Does it make sense to have a single function `transformForHostRoot` here and then call these two in that function?

controllers/object_controls.go
```diff
@@ -773,6 +783,48 @@ func applyCommonDaemonsetConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+	for index := range obj.Spec.Template.Spec.Containers {
+		setContainerEnv(&(obj.Spec.Template.Spec.Containers[index]), HostRootEnvName, hostRoot)
+	}
+}
+
+// Transform the host-dev-char volume if required for the Daemonset passed as input
+func transformHostDevCharVolume(obj *appsv1.DaemonSet, hostRoot string) {
+	if hostRoot == "" || hostRoot == "/" {
+		return
+	}
```
elezar: See my comment on adding a single `transformForHostRoot` function where we can do this check once.

controllers/object_controls.go
```diff
@@ -1213,6 +1265,36 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
 			}
 		}
 	}
+
+	// Update the driver-root volume if a custom driver install directory is configured with the operator
```
elezar: Having the driver transform logic in a separate function would help to make this more testable.

controllers/object_controls.go
```diff
@@ -1298,6 +1380,35 @@ func TransformDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
...
+	if config.HostPaths.DriverInstallDir != "" && config.HostPaths.DriverInstallDir != DefaultDriverInstallDir {
+		for index, volume := range obj.Spec.Template.Spec.Volumes {
+			if volume.Name == "driver-root" {
+				obj.Spec.Template.Spec.Volumes[index].HostPath.Path = config.HostPaths.DriverInstallDir
+			}
+		}
+	}
+
+	// If a custom devPath is configured which is not equal to 'hostRoot/dev' or 'driverInstallDir/dev', make it available
```
elezar: Are we duplicating the logic between here and the toolkit (and the mig-manager?). Is this something that is applicable to all three cases?

elezar (commented):
Thanks @cdesiniotis. It's a big PR, but I took a basic pass.

A general question is whether it makes sense to factor some of the transform logic that we are adding into functions instead of implementing them in-line in the three cases where we need them.

controllers/object_controls.go
cdesiniotis: @elezar thoughts on removing `devPath` from this PR? In our driver validator, we would look for a `/dev` filesystem at `driverInstallDir/dev` and at `hostRoot/dev` before performing the driver validation. This is similar to logic in our k8s-dra-driver. At the very least, removing `devPath` would simplify the changes in this PR and make it easier to review. If required, we can add `devPath` in a follow-up.

deployments/gpu-operator/values.yaml
cdesiniotis: Good question. I think the documentation would be limited to start with. We would only document these options for the GKE use case. Once our driver container supports custom values, then we can document the use of these options more generally and explain how these can serve as input to our driver container.

validator/main.go
cdesiniotis: While we can't strictly guarantee the driver validator will start after the driver pod initially, our driver pod always restarts all operand containers, including the toolkit and validator pods, in it's `k8s-driver-manager` initContainer: https://github.com/NVIDIA/k8s-driver-manager/blob/master/driver-manager#L591-L592.

validator/main.go
cdesiniotis: Yes, I was eluding to always mounting the device nodes to `/dev-root/dev` for the purpose of CDI spec generation.

api/nvidia/v1/clusterpolicy_types.go
cdesiniotis: I prefer `DevPath`

controllers/object_controls.go
cdesiniotis: @tariq1890 while at the moment it may be okay to remove this option, once our driver container supports customizing the "driver root" and device node path, it may be useful to explicitly set both of these paths. For example, instead of installing everything at `/run/nvidia/driver`, we can instruct the driver container to install the driver at `/opt/nvidia/driver` and device nodes at `/dev`, for example.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: No, I had changed this to `driver-root` previously to align with naming this field `driverRoot` in the ClusterPolicy API. But since we renamed it to `driverInstallDir`, I am open to renaming it. Although our operands still use the `driverRoot` name.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: I am using the exact hostPath so that dev-char symlink creation works correctly. I want to make sure the target of the symlink resolves correctly on the host. So for example, we create a symlink `/host-dev-char/195:0` --> `/run/nvidia/driver/dev/nvidia0` in the driver validator container. I guess if we transformed the target path before creating the symlink, then we could always mount the driver install directory at `/driver-root` or `/driver-install-dir`. What do you think?

assets/state-device-plugin/0400_configmap.yaml
cdesiniotis: Because the initContainer in the toolkit pod creates the `driver-ready` file.

controllers/object_controls.go
cdesiniotis: As currently implemented, I am expecting `DevPath` to be a `/dev` filesystem. It would probably be a good idea to validate that it ends in `/dev` ...

validator/main.go
cdesiniotis: You are right, we always end up creating the device nodes at `/dev` in this case. Is it possible to invoke nvidia-smi this way but have the device nodes created at `devPath`?

controllers/object_controls.go
cdesiniotis: Based on the initial set of use cases we are targeting, I believe removing `devPath` from the API is sufficient for the time being. 

The logic in the driver validator would remain the same. The only difference would be that before creating the `driver-ready` status file, we would determine the value for `devRoot`. If `driverInstallDir/dev` exists, then we set `devRoot = driverInstallDir`. Else we fallback to setting `devRoot = hostRoot`.

Below are the various use cases:

## Driver Container
Input:
```
driver:
  enabled: true
  
hostPaths:
  root: /
  driverInstallDir: /run/nvidia/driver
```
Resolved values:
```
NVIDIA_DRIVER_ROOT: /run/nvidia/driver
DRIVER_ROOT_CTR_PATH: /driver-root
NVIDIA_DEV_ROOT: /run/nvidia/driver
DEV_ROOT_CTR_PATH: /driver-root
```

## Driver Container with host driver
Input:
```
driver:
  enabled: true
  
hostPaths:
  root: /
  driverInstallDir: /run/nvidia/driver
```
Resolved values:
```
NVIDIA_DRIVER_ROOT: /
DRIVER_ROOT_CTR_PATH: /host
NVIDIA_DEV_ROOT: /
DEV_ROOT_CTR_PATH: /host
```

## GKE driver
Input:
```
driver:
  enabled: false
  
hostPaths:
  root: /
  driverInstallDir: /home/kubernetes/bin/nvidia
```
Resolved values:
```
NVIDIA_DRIVER_ROOT: /home/kubernetes/bin/nvidia
DRIVER_ROOT_CTR_PATH: /driver-root
NVIDIA_DEV_ROOT: /
DEV_ROOT_CTR_PATH: /host
```

cdesiniotis: I have removed `devPath` from the ClusterPolicy API now.

deployments/gpu-operator/values.yaml
cdesiniotis: I have removed devPath from the ClusterPolicy API now.

validator/main.go
cdesiniotis: I have removed devPath from the ClusterPolicy API now. This code has been simplified now.

controllers/object_controls.go
cdesiniotis: I have removed devPath from the ClusterPolicy API now. Consequently, this code has been removed.

controllers/object_controls.go
cdesiniotis: I have removed devPath from the ClusterPolicy API now. Consequently, this code has been removed.

controllers/object_controls.go
cdesiniotis: I have removed devPath from the ClusterPolicy API now. Consequently, this code has been removed.

deployments/gpu-operator/templates/clusterpolicy.yaml
cdesiniotis: I have removed devPath from the ClusterPolicy API now. Consequently, this code has been removed.

validator/main.go
cdesiniotis: Done.

controllers/object_controls.go
cdesiniotis: I have removed devPath from the ClusterPolicy API now. Consequently, this code has been removed.

api/nvidia/v1/clusterpolicy_types.go
cdesiniotis: Updated to `RootFS` as suggested.

controllers/object_controls.go
cdesiniotis: Makes sense. I have added a `transformForHostRoot` function.

controllers/object_controls.go
cdesiniotis: Added test cases for the host root transforms.

controllers/object_controls.go
cdesiniotis: Updated as suggested.

controllers/object_controls.go
cdesiniotis: Done.

controllers/object_controls.go
cdesiniotis: Done.

controllers/object_controls.go
cdesiniotis: Done

controllers/object_controls.go
cdesiniotis: I have not moved this to a constant since we are now only setting this specific environment variable once in `transformForDriverInstallDir()`.

validator/main.go
cdesiniotis: Done

validator/main.go
cdesiniotis: I have removed the devRootFlag.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: To reduce churn, I have kept the `driver-root` name. If preferred, I can rename this volume / volumeMount to `driver-install-dir`  in a follow-up.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: These are good questions. With my limited testing, the toolkit container appears to acquire the exclusive lock just fine, even after restarting the pod or uninstalling / re-installing the operator. But definitely something I can look into more.

Note, I decided to remove the `/run/nvidia` mount to avoid mounting the driver install directory twice -- once at `/driver-root` in the container and again at `/run/nvidia/driver`. If preferred, I can reintroduce the original `/run/nvidia` mount, or we can update the toolkit container to create the pid file at another path, like `/run/nvidia/toolkit/toolkit.pid`.

validator/main.go
cdesiniotis: I can look into this in a follow-up.

controllers/object_controls.go
```diff
@@ -773,6 +785,73 @@ func applyCommonDaemonsetConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+			containsHostRootVolume = true
+			break
+		}
+	}
+
+	if !containsHostRootVolume {
+		return
+	}
+
+	for index := range obj.Spec.Template.Spec.InitContainers {
```
cdesiniotis, line 815: Reading this again, I can simplify by iterating over all containers and adding this envvar if the container has a volumeMount named "host-root"

assets/state-container-toolkit/0500_daemonset.yaml
klueska: Yes, let's make it match the field names in the `HostPathsSpec` struct.

validator/find.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// getDevRoot returns the dev root associated with the root.
+// If the root is not a dev root, this defaults to "/".
+func (r root) getDevRoot() string {
+	if r.isDevRoot() {
+		return string(r)
+	}
+	return "/"
+}
```
klueska, line 81: I was surprised to not see a field for `devPath` in `HostPathsSpec`, but then I recalled a conversation about not needing to specify it if we assume that it is either only ever at:
1. `<DriverInstallDir>/dev`; OR
1. `/dev`

It looks like this is the logic that is implementing this assumption.

assets/state-container-toolkit/0500_daemonset.yaml
elezar: Doing so as a follow up is fine though.

assets/state-container-toolkit/0500_daemonset.yaml
elezar: I believe we'd be able to create the symlink regardless of whether the target exists or not, so as lon as `NVIDIA_DEV_ROOT` (or `NVIDIA_DEV_PATH`) refers to the correct path *on the host* the symlinks should work as expected.

In the medium term we could consider moving this out of the toolkit container anyway ... since it really should be a postcondition of the driver container.

Let's leave it as is for now though.

assets/state-container-toolkit/0500_daemonset.yaml
elezar: I'll defer to you on this. The one thing to note is that the we would not be able to remove the file here https://github.com/NVIDIA/nvidia-container-toolkit/blob/ac90b7963d24ba27cf5d17de85e7b17c9d1ed184/tools/container/nvidia-toolkit/run.go#L289-L296

What about mounting (as `DirectoryOrCreate`) `/run/nvidia/toolkit` and creating the pid file at `/run/nvidia/toolkit/toolkit.pid`. This may require (minor) changes to the toolkit container. The one issue is that the toolkit container does remove `/run/nvidia/toolkit` which would fail if `toolkit.pid` exists there.

I have created https://github.com/NVIDIA/nvidia-container-toolkit/pull/544 that should address this and should be compatible with this change if you update the mount spec to `/run/nvidia/toolkit` as a folder.

assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -96,17 +100,18 @@ spec:
           configMap:
             name: nvidia-container-toolkit-entrypoint
             defaultMode: 448
-        - name: nvidia-run-path
+        - name: toolkit-pid-file
```
elezar:
```suggestion
        - name: toolkit-root
```

assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -96,17 +100,18 @@ spec:
...
           hostPath:
-            path: /run/nvidia
-            type: DirectoryOrCreate
+            path: /run/nvidia/toolkit.pid
+            type: FileOrCreate
```
elezar:
```suggestion
            path: /run/nvidia/toolkit
            type: DirectoryOrCreate
```

assets/state-device-plugin/0400_configmap.yaml
elezar: Why is the `driver-ready` file not created by the driver validator container? Any init contianer for a component that depends on the driver being ready should wait on this file.

assets/state-mig-manager/0600_daemonset.yaml
```diff
@@ -88,10 +90,14 @@ spec:
...
+      - name: run-nvidia-validations
         hostPath:
-          path: "/run/nvidia"
-          type: Directory
+          path: "/run/nvidia/validations"
+          type: DirectoryOrCreate
+      - name: driver-root
+        hostPath:
+          path: "/run/nvidia/driver"
+          type: DirectoryOrCreate
```
elezar: Out of scope: This is begining to look like the kind of thing we can add to each component in code. Not sure what the best practice is in this case.

controllers/object_controls.go
```diff
@@ -773,6 +785,73 @@ func applyCommonDaemonsetConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+	}
+
+	podSpec := obj.Spec.Template.Spec
+	for _, volume := range podSpec.Volumes {
+		if volume.Name == "driver-root" {
+			volume.HostPath.Path = driverInstallDir
+		}
+	}
+
+	for i, ctr := range podSpec.InitContainers {
```
elezar, line 853: Should this only be done if we use the volume above?

controllers/object_controls.go
```diff
@@ -773,6 +785,73 @@ func applyCommonDaemonsetConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+		if ctr.Name == "driver-validation" {
+			setContainerEnv(&(podSpec.InitContainers[i]), "DRIVER_ROOT", driverInstallDir)
```
elezar: I may have asked somewhere. What is the difference between `DRIVER_ROOT` and `NVIDIA_DRIVER_ROOT`?

controllers/object_controls.go
```diff
@@ -2106,6 +2185,7 @@ func TransformValidatorComponent(config *gpuv1.ClusterPolicySpec, podSpec *corev
 					setContainerEnv(&(podSpec.InitContainers[i]), env.Name, env.Value)
 				}
 			}
+			setContainerEnv(&(podSpec.InitContainers[i]), DriverManagedByOperatorEnvName, fmt.Sprintf("%t", config.Driver.IsEnabled()))
```
elezar: It's confusing that has to be set here as well as in `transformValidationInitContainer` since this seems to be looping over the init containers for the validator already.

controllers/transforms_test.go
elezar, line 1: Nice!

validator/driver.go
```diff
@@ -0,0 +1,72 @@
...
+			driverRootCtrPath: "/host",
+			devRoot:           hostRoot,
+			devRootCtrPath:    "/host",
+		}
+	}
+
+	// For drivers not installed directly on the host, devRoot will
+	// either be driverRoot or hostRoot.
+	var devRoot, devRootCtrPath string
+	devRoot = root(driverRoot).getDevRoot()
```
elezar: This is running in a container, correct. Is `driverRoot` here the container driver root or the host driver root?

validator/main.go
```diff
@@ -318,6 +320,27 @@ func main() {
...
+			Name:        "host-root",
+			Value:       "/",
+			Usage:       "root path of the underlying host",
+			Destination: &hostRootFlag,
+			EnvVars:     []string{"HOST_ROOT"},
+		},
+		&cli.StringFlag{
+			Name:        "driver-root",
+			Value:       defaultDriverRoot,
+			Usage:       "root path of a containerized NVIDIA driver installation",
```
elezar: Is this the path on the host or in the container?

validator/main.go
elezar: Sure. Let's do it as a follow-up.

elezar (commented):
Thanks @cdesiniotis. I think we're getting there.

I still have some comments / questions, but these are mostly informative.

validator/main.go
```diff
@@ -318,6 +320,27 @@ func main() {
 			Destination: &disableDevCharSymlinkCreation,
 			EnvVars:     []string{"DISABLE_DEV_CHAR_SYMLINK_CREATION"},
 		},
+		&cli.BoolFlag{
```
tariq1890: I think we should avoid adding this env var to the validator container. I would much prefer if we had logic within the validator to introspect the owner of the pod via the kube api and then determine if the driver is gpu-operator managed or not. Adding an env var to function as a boolean flag does not seem right to me

validator/main.go
```diff
@@ -591,14 +616,26 @@ func runCommandWithWait(command string, args []string, sleepSeconds int, silent
...
+	if len(prepend) == 0 {
+		return os.Getenv(envvar)
 	}
+	current := filepath.SplitList(os.Getenv(envvar))
+	return strings.Join(append(prepend, current...), string(filepath.ListSeparator))
+}
 
-	return driverContainerRoot, false
+// setOrOverrideEnvvar adds or updates an envar to the list of specified envvars and returns it.
+func setOrOverrideEnvvar(envvars []string, key, value string) []string {
```
tariq1890: I think `set` more or less implies that it will override the value if it has already been set previously
```suggestion
func setEnvVar(envvars []string, key, value string) []string {
```

validator/main.go
tariq1890: Ah, I just read your note. Looks like we are on the same page. Can we make the change in this PR instead of doing it in a follow-up?

cdesiniotis: I agree, and I suggested this approach here: https://github.com/NVIDIA/gpu-operator/pull/747#discussion_r1633942404

assets/state-device-plugin/0400_configmap.yaml
```diff
@@ -9,28 +9,15 @@ data:
...
     done
-
-    export NVIDIA_DRIVER_ROOT=$driver_root
-    echo "NVIDIA_DRIVER_ROOT=$NVIDIA_DRIVER_ROOT"
-
-    export CONTAINER_DRIVER_ROOT=$container_driver_root
-    echo "CONTAINER_DRIVER_ROOT=$CONTAINER_DRIVER_ROOT"
+    
+    set -o allexport
+    cat /run/nvidia/validations/driver-ready
```
Dragoncell, line 19: Does it also export the custom toolkit installation path from values in toolkit to the device plugin for nvidia-ctk tool ?

cdesiniotis: No. But we set the `NVIDIA_CTK_PATH` variable here for the device-plugin in the clusterpolicy controller: https://github.com/NVIDIA/gpu-operator/blob/main/controllers/object_controls.go#L1284.

controllers/object_controls.go
tariq1890: Thanks!

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: Thanks @elezar. I prefer to create the pid file at `/run/nvidia/toolkit/toolkit.pid` instead, so let's go ahead with the change you are making to the toolkit container.

validator/driver.go
cdesiniotis: Yes this code is running in container.

Admittedly this code is a bit hard to follow. To explain using an example, let's say GPU Operator is configured with the below values:
```
hostPaths:
  rootFS: "/"
  driverInstallDir: "/run/nvidia/driver"
```
The driver-validator container will have the rootFS mounted at `/host` and the driverInstallDir mounted at `/run/nvidia/driver`.

At this point in the code, `driverRoot` will be `/run/nvidia/driver` which is both the path to the driver root in the validator container and on the host.

validator/find.go
cdesiniotis: Correct. We decided to omit `devPath` from `HostPathsSpec` to start with. If a use case presents itself where device nodes need to exist at some other location, we can introduce this field to the API.

validator/main.go
cdesiniotis: As currently implemented, it happens to be both the path on the host AND in the driver-validation container, due to the way the volumeMount is configured for the daemonset.

It may be good to split this into `driver-root` and `driver-root-ctr-path` fields for clarity (even if they are the same values to begin with).

validator/find.go
```diff
@@ -0,0 +1,109 @@
...
+	return "", fmt.Errorf("error locating %q", name)
+}
+
+// resolveLink finds the target of a symlink or the file itself in the
+// case of a regular file.
+// This is equivalent to running `readlink -f ${l}`.
+func resolveLink(l string) (string, error) {
+	resolved, err := filepath.EvalSymlinks(l)
+	if err != nil {
+		return "", fmt.Errorf("error resolving link '%v': %v", l, err)
```
tariq1890:
```suggestion
		return "", fmt.Errorf("error resolving link '%v': %w", l, err)
```

validator/main.go
```diff
@@ -641,69 +725,77 @@ func (d *Driver) validate() error {
 		return err
 	}
 
-	// delete host driver status file is already present
-	err = deleteStatusFile(outputDirFlag + "/" + hostDriverStatusFile)
+	driverInfo, err := d.runValidation(false)
 	if err != nil {
+		log.Error("driver is not ready")
```
tariq1890: Let's log the error in this line too
```suggestion
		log.Error("driver is not ready")
```

validator/main.go
```diff
@@ -641,69 +725,77 @@ func (d *Driver) validate() error {
...
+			"",
+			"This bug impacts container runtimes configured with systemd cgroup management enabled.",
+			"To disable the symlink creation, set the following envvar in ClusterPolicy:",
+			"",
+			"    validator:",
+			"      driver:",
+			"        env:",
+			"        - name: DISABLE_DEV_CHAR_SYMLINK_CREATION",
+			"          value: \"true\""}, "\n")
+		return fmt.Errorf("%v\n\n%s", err, msg)
```
tariq1890:
```suggestion
		return fmt.Errorf("%w\n\n%s", err, msg)
```

validator/main.go
```diff
@@ -641,69 +725,77 @@ func (d *Driver) validate() error {
...
+	return d.createStatusFile(driverInfo)
+}
 
-	statusFile := driverStatusFile
-	if isHostDriver {
-		statusFile = hostDriverStatusFile
-	}
+func (d *Driver) createStatusFile(driverInfo driverInfo) error {
+	statusFileContent := strings.Join([]string{
+		fmt.Sprintf("IS_HOST_DRIVER=%v", driverInfo.isHostDriver),
```
tariq1890:
```suggestion
		fmt.Sprintf("IS_HOST_DRIVER=%t", driverInfo.isHostDriver),
```

validator/main.go
```diff
@@ -614,24 +651,71 @@ func assertDriverContainerReady(silent bool) error {
...
+func (d *Driver) runValidation(silent bool) (driverInfo, error) {
+	err := validateHostDriver(silent)
+	if err == nil {
+		log.Info("Detected a pre-installed driver on the host")
+		return getDriverInfo(true, "/host", hostRootFlag), nil
+	}
+
+	err = validateDriverContainer(silent)
+	if err != nil {
+		return driverInfo{}, err
```
tariq1890, line 761: Why not return nil?
```suggestion
		return nil, err
```

assets/state-device-plugin/0400_configmap.yaml
cdesiniotis: I have updated the container-toolkit startup script to have the same logic.

validator/main.go
cdesiniotis: Following discussions with @tariq1890, I have decided to update this PR with the approach I suggested. Instead of passing a boolean flag to the validator, we are now inferring this information at runtime by checking for the presence of a GPU Operator-owned driver daemonset.

validator/main.go
cdesiniotis: Done.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: I have gone ahead and updated the name of this volume to `driver-install-dir` throughout to align with the name of the field in ClusterPolicy.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: I have updated this now. We now mount `/run/nvidia/toolkit` as `DirectoryOrCreate` in the toolkit container. We also set the `TOOLKIT_PID_FILE=/run/nvidia/toolkit/toolkit.pid` envvar explicitly.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: Done.

controllers/object_controls.go
cdesiniotis: Yeah. I have updated this so that we only iterate over the container if the volume exists above.

controllers/object_controls.go
cdesiniotis: For clarity, I have updated the envvars for the driver-validator to `DRIVER_INSTALL_DIR` and `DRIVER_INSTALL_DIR_CTR_PATH` to align with the name of the field in ClusterPolicy. `DRIVER_INSTALL_DIR` represents the path on the host, while `DRIVER_INSTALL_DIR_CTR_PATH` represents the path mounted in the driver-validator container.

validator/driver.go
cdesiniotis: For clarity, I have updated the driver validator to have two flags -- `driver-install-dir` and `driver-install-dir-ctr-path`. I have updated this code to be 
```
devRoot = root(driverInstallDirCtrPath).getDevRoot()
```
so that it is clear we are searching for the devRoot in the container.

validator/find.go
cdesiniotis: Done

validator/main.go
cdesiniotis: For clarity, I have updated the driver validator to have two flags -- `driver-install-dir` and `driver-install-dir-ctr-path`.

validator/main.go
cdesiniotis: Done

validator/main.go
cdesiniotis: Done

validator/main.go
cdesiniotis: Done.

validator/main.go
cdesiniotis: Done.

assets/state-container-toolkit/0500_daemonset.yaml
cdesiniotis: Done.

controllers/object_controls.go
cdesiniotis: This has now been removed.

assets/state-container-toolkit/0200_role.yaml
```diff
@@ -12,3 +12,9 @@ rules:
   - use
   resourceNames:
   - privileged
+- apiGroups:
```
elezar, line 15: I was not aware of the toolkit inspecting daemonsets? Was thsi change intentional?

assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -67,6 +71,8 @@ spec:
           value: "management.nvidia.com/gpu"
         - name: NVIDIA_VISIBLE_DEVICES
           value: "void"
+        - name: TOOLKIT_PID_FILE
```
elezar, line 74: For reference. This requires https://github.com/NVIDIA/nvidia-container-toolkit/pull/544

assets/state-container-toolkit/0200_role.yaml
elezar: Is this because the validator runs here too? I thought we changed the logic to just look for the ready file?

validator/driver.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// For drivers not installed directly on the host, devRoot can either be
+	// hostRoot or driverInstallDir
+	var devRoot, devRootCtrPath string
+	devRoot = root(driverInstallDirCtrPath).getDevRoot()
+	if devRoot == "/" {
+		devRoot = hostRoot
+		devRootCtrPath = "/host"
+	} else {
+		devRoot = driverInstallDir
```
elezar, line 61: Just so that I'm clear, this is only triggered for managed drivers?

validator/main.go
elezar: My understanding is that `nvidia-smi` expects the device nodes created at `/dev` and that there isn't a way to control this.

validator/main.go
```diff
@@ -641,69 +770,78 @@ func (d *Driver) validate() error {
...
-		statusFile = hostDriverStatusFile
-	}
+func (d *Driver) createStatusFile(driverInfo driverInfo) error {
+	statusFileContent := strings.Join([]string{
+		fmt.Sprintf("IS_HOST_DRIVER=%t", driverInfo.isHostDriver),
+		fmt.Sprintf("NVIDIA_DRIVER_ROOT=%s", driverInfo.driverRoot),
+		fmt.Sprintf("DRIVER_ROOT_CTR_PATH=%s", driverInfo.driverRootCtrPath),
+		fmt.Sprintf("NVIDIA_DEV_ROOT=%s", driverInfo.devRoot),
+		fmt.Sprintf("DEV_ROOT_CTR_PATH=%s", driverInfo.devRootCtrPath),
+	}, "\n") + "\n"
```
elezar, line 808: nit: is this something that we could hang off the `driverInfo` type?

elezar (approved):
I'm happy to move forward with this as is and then address any findings from testing as follow-ups.

assets/state-container-toolkit/0200_role.yaml
cdesiniotis: Yes, it is because the validator runs here as an init container. We still need to run the driver validation here since the operator-validator pod runs with the `nvidia` runtime class, and thus, needs the toolkit pod to run first before it can run and start performing any validations.

validator/driver.go
cdesiniotis: No. This logic applies to any driver not installed on the host directly at `/`. So this can be the operator managed driver or an unmanaged driver, like GKEs daemonset.

cdesiniotis: The `driverInstallDir` that is referenced here will correspond to what is set in ClusterPolicy as `hostPaths.driverInstallDir`.

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -12,6 +12,9 @@ metadata:
     "helm.sh/resource-policy": keep
   {{- end }}
 spec:
+  hostPaths:
+    rootFS: {{ .Values.hostPaths.rootFS | default "/" }}
```
tariq1890: Is the default needed here?

Won't the default value in the values.yaml suffice?

validator/main.go
```diff
@@ -591,14 +622,26 @@ func runCommandWithWait(command string, args []string, sleepSeconds int, silent
...
+func prependPathListEnvvar(envvar string, prepend ...string) string {
+	if len(prepend) == 0 {
+		return os.Getenv(envvar)
 	}
+	current := filepath.SplitList(os.Getenv(envvar))
+	return strings.Join(append(prepend, current...), string(filepath.ListSeparator))
+}
 
-	return driverContainerRoot, false
+// setOrOverrideEnvvar adds or updates an envar to the list of specified envvars and returns it.
```
tariq1890:
```suggestion
// setEnvVar adds or updates an envar to the list of specified envvars and returns it.
```

validator/main.go
```diff
@@ -614,24 +657,110 @@ func assertDriverContainerReady(silent bool) error {
...
+		return false, fmt.Errorf("error listing daemonsets: %w", err)
+	}
+
+	for i := range dsList.Items {
+		ds := dsList.Items[i]
+		owner := meta_v1.GetControllerOf(&ds)
+		if owner == nil {
+			continue
+		}
+		if owner.Kind == nvidiav1.ClusterPolicyCRDName || owner.Kind == nvidiav1alpha1.NVIDIADriverCRDName {
```
tariq1890: Makes this check more robust.
```suggestion
		if owner.APIVersion == "nvidia.com" && (owner.Kind == nvidiav1.ClusterPolicyCRDName || owner.Kind == nvidiav1alpha1.NVIDIADriverCRDName) {
```

validator/find.go
```diff
@@ -0,0 +1,109 @@
...
+	return "", fmt.Errorf("error locating %q", name)
+}
+
+// resolveLink finds the target of a symlink or the file itself in the
+// case of a regular file.
+// This is equivalent to running `readlink -f ${l}`.
+func resolveLink(l string) (string, error) {
+	resolved, err := filepath.EvalSymlinks(l)
+	if err != nil {
+		return "", fmt.Errorf("error resolving link '%v': %w", l, err)
```
tariq1890:
```suggestion
		return "", fmt.Errorf("error resolving link '%s': %w", l, err)
```

validator/main.go
cdesiniotis:
```
cannot use nil as driverInfo value in return statement
```

validator/main.go
cdesiniotis: Wouldn't API version be `nvidia.com/v1` or `nvidia.com/v1alpha1`?

cdesiniotis: What about?
```
strings.HasPrefix(owner.APIVersion, "nvidia.com/") && . . .
```

tariq1890: That works!

deployments/gpu-operator/templates/clusterpolicy.yaml
cdesiniotis: Yeah, the default value in values.yaml should suffice. I have updated this.

validator/main.go
cdesiniotis: Done.

validator/main.go
cdesiniotis: Done.

validator/find.go
cdesiniotis: Done.

tariq1890 (approved):

validator/find.go
```diff
@@ -0,0 +1,109 @@
...
+// getDriverLibraryPath returns path to `libnvidia-ml.so.1` in the driver root.
+// The folder for this file is also expected to be the location of other driver files.
+func (r root) getDriverLibraryPath() (string, error) {
+	librarySearchPaths := []string{
+		"/usr/lib64",
+		"/usr/lib/x86_64-linux-gnu",
+		"/usr/lib/aarch64-linux-gnu",
+		"/lib64",
+		"/lib/x86_64-linux-gnu",
+		"/lib/aarch64-linux-gnu",
```
jgehrcke, line 36: Trying to catch up with history -- did we have thoughts about adding `/usr/lib` here?

validator/find.go
```diff
@@ -0,0 +1,109 @@
...
+	return libraryPath, nil
+}
+
+// getNvidiaSMIPath returns path to the `nvidia-smi` executable in the driver root.
+func (r root) getNvidiaSMIPath() (string, error) {
+	binarySearchPaths := []string{
+		"/usr/bin",
+		"/usr/sbin",
+		"/bin",
+		"/sbin",
```
jgehrcke, line 53: Did we maybe think about `/usr/local/bin` here? I am mostly curious about if there were maybe any counter-arguments. Thinking about some of that in the context of https://github.com/NVIDIA/k8s-dra-driver-gpu/issues/605.

PR #749, opened 2024-06-11
[CC Mgr VGPU Device Mgr] move pods access permissions from ClusterRole to Role

tariq1890:


assets/state-vgpu-device-manager/0210_clusterrole.yaml
```diff
@@ -14,8 +14,6 @@ rules:
   - ""
   resources:
   - nodes
-  - pods
-  - pods/eviction
```
cdesiniotis, line 18: I believe access to `pods/eviction` was never needed in the first place. We only watch for pods (our operands) to be deleted before applying a vGPU device configuration.

cdesiniotis (approved):

assets/state-vgpu-device-manager/0210_clusterrole.yaml
tariq1890: Yes, I had checked the vgpu device manager code to confirm this

PR #750, opened 2024-06-11
[RBAC] move namespace-scoped resource permissions to Roles

tariq1890:


bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -714,29 +742,52 @@ spec:
...
+          - update
+          - patch
+          - delete
         - apiGroups:
-          - apiextensions.k8s.io
+          - ""
           resources:
-          - customresourcedefinitions
+          - configmaps
+          - endpoints
```
cdesiniotis, line 788: Question -- do you know why we only specify `endpoints` in the OLM bundle but not in the chart?

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -601,6 +601,7 @@ spec:
           - patch
           - update
           - watch
+          - deletecollection
```
cdesiniotis, line 604: Question -- why is this needed now?

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -631,13 +632,31 @@ spec:
...
+          - imagestreams
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - monitoring.coreos.com
+          resources:
+          - servicemonitors
+          - prometheusrules
```
cdesiniotis, line 647: Question -- do we really need these at the cluster level?

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
tariq1890: I just added it so that it is consistent with the helm charts

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
tariq1890: Good question, when I was testing it on OpenShift, the gpu-operator controller errored out saying that it lacked permissions to access endpoints. Adding this back resolved the error

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
tariq1890: Yeah, we seem to need it for OpenShift :(. When I moved it to Roles, the gpu-operator controller would fail

cdesiniotis (approved):

PR #758, opened 2024-06-14
feat: vfio-manager graphics mode

jojimt:
Changes vfio-manager script to bind the auxiliary device to vfio-pci.
@cdesiniotis @zvonkok PTAL

### DIFF
--- a/assets/state-vfio-manager/0400_configmap.yaml
+++ b/assets/state-vfio-manager/0400_configmap.yaml
@@ -94,6 +94,12 @@ data:
 
         echo "unbinding device $gpu"
         unbind_from_driver $gpu
+        #for graphics mode, we need to unbind the auxiliary device as well
+        aux_dev=$(get_graphics_aux_dev "$gpu")
+        if [ "$aux_dev" != "NONE" ]; then
+          echo "gpu $gpu is in graphics mode aux_dev $aux_dev"
+          unbind_from_driver "$aux_dev"
+        fi
     }
 
     unbind_all() {
@@ -106,12 +112,8 @@ data:
         done
     }
 
-    bind_device() {
+    bind_pci_device() {
         local gpu=$1
-
-        if ! is_nvidia_gpu_device $gpu; then
-            return 0
-        fi
 
         if ! is_bound_to_vfio $gpu; then
           unbind_from_other_driver $gpu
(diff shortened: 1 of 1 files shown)


assets/state-vfio-manager/0400_configmap.yaml
```diff
@@ -123,6 +125,35 @@ data:
         fi
     }
 
+    get_graphics_aux_dev() {
+        local gpu=$1
+        device_class_file=$(readlink -f "/sys/bus/pci/devices/$gpu/class")
+        device_class=$(cat "$device_class_file")
+        if [ "$device_class" == "0x030000" ]; then
```
cdesiniotis: Some questions for my understanding:

Will there ever be a scenario where the GPU is configured in graphics mode, but no auxilary device is present? Also, does device class `0x03000` always indicate a GPU is in graphics mode?

jojimt: Per https://docs.nvidia.com/grid/13.0/grid-gpumodeswitch-user-guide/index.html#compute-and-graphics-mode, the answer to the second question is yes. Not sure about the first question. But the script will gracefully handle this case, by doing nothing.

cdesiniotis:
```
Classcode | VGA Controller | This classcode indicates to OS that the GPU can function as a primary display device.
```
I believe this answers my second question. Thanks.

> Not sure about the first question. But the script will gracefully handle this case, by doing nothing.

It is not clear to me that we handle this case gracefully. If we find a GPU that is a VGA controller (class 0x030000), we always return the output of `aux_dev=$(ls "/sys/bus/pci/devices/$gpu" | grep consumer | awk -Fconsumer:pci: '{print $2}')` regardless if there is a valid auxiliary device or not. We then call `bind_pci_device "$aux_dev"` with this string. I am not sure what the behavior of `bind_pci_device` is for invalid input.

jojimt: Right, let me push a fix to handle this case.

zvonkok: I am curious how this will change once we enable SR-IOV and we have VFs. I haven't checked how the VFs are enumerated regarding BDF in graphics mode. Will the `consumer` link change or not, I suppose not because what `consumer` means is that a `pci` device with BDF xxxx:xx:xx.1 uses the parent device that we're currently inspecting. 
```
consumer:pci:0000:19:00.1 -> ../../../../../virtual/devlink/pci:0000:19:00.0--pci:0000:19:00.1
```

zvonkok: Can we check that we have a BDF that is returned if anything is returned?

zvonkok: Or do a check if we have a multi function device if we're not using sriov?

jojimt: Added a check to verify /sys/bus/pci/devices/$aux_dev/ exists.

assets/state-vfio-manager/0400_configmap.yaml
```diff
@@ -123,6 +125,44 @@ data:
...
+          if ls "/sys/bus/pci/devices/$gpu" | grep consumer >& /dev/null; then
+            aux_dev=$(ls "/sys/bus/pci/devices/$gpu" | grep consumer | awk -Fconsumer:pci: '{print $2}')
+            if [ "$device_class" == "" ]; then
+              echo "NONE"
+              return
+            fi
```
cdesiniotis: Why are we checking for an empty device class string here? Couldn't we check this before checking the `consumer` link?

assets/state-vfio-manager/0400_configmap.yaml
```diff
@@ -123,6 +125,44 @@ data:
         fi
     }
 
+    get_graphics_aux_dev() {
+        local gpu=$1
+        device_class_file=$(readlink -f "/sys/bus/pci/devices/$gpu/class")
+        device_class=$(cat "$device_class_file")
+        if [ "$device_class" == "0x030000" ]; then
```
cdesiniotis: For better readability, I would suggest inverting this if statement and returning early if the device is not a VGA controller:
```
if [ "$device_class" != "0x030000" ]; then
  echo "NONE"
  return
fi

if ls "/sys/bus/pci/devices/$gpu" | grep consumer >& /dev/null; then
   <snip>
fi
<snip>

```

assets/state-vfio-manager/0400_configmap.yaml
jojimt: You caught a copy/paste error. Meant to check for empty aux_dev.

cdesiniotis (approved):

PR #760, opened 2024-06-14
Add dependabot config

cdesiniotis:


elezar (approved):
LGTM.

I squashed your last two commits (DCO was failing) and changed the order of the dependabot entries.

.github/dependabot.yml
```diff
@@ -0,0 +1,36 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gomod"
```
elezar, line 6: @cdesiniotis one thing to consider is whether we want update for the latest `release-*` branch too?

cdesiniotis:
Thank you @elezar!

.github/dependabot.yml
cdesiniotis: Not opposed to it. @tariq1890 what do you think?

PR #780, opened 2024-06-24
[MIG] add support for H200 141GB

tariq1890:


assets/state-mig-manager/0400_configmap.yaml
```diff
@@ -145,37 +145,75 @@ data:
           mig-devices:
             "1g.12gb": 7
 
+      # H200-141GB
```
cdesiniotis: For readability, can we group all of the H200-141GB configurations together? For example: 
```
# H200-141GB
all-1g.18gb: 
 < snip >
 
all-1g.18gb.me:
  < snip >
.
.

all-7g.141gb:
  < snip > 
```

assets/state-mig-manager/0400_configmap.yaml
```diff
@@ -201,9 +239,24 @@ data:
...
+      # H200-141GB, H100-96GB, GH200, H100 NVL, H800 NVL, H100-80GB, H800-80GB, A800-40GB, A800-80GB, A100-40GB, A100-80GB, A30-24GB, PG506-96GB
       all-balanced:
+        # H200 141GB
+        - device-filter: ["0x233510DE"]
+          devices: all
+          mig-enabled: true
+          mig-devices:
+            "1g.18gb": 2
+            "2g.35gb": 1
+            "3g.71gb": 1
```
cdesiniotis, line 213: This may not be a valid configuration. The total memory adds up to 142GB. We may need to limit this to just one `1g.18gb` instance instead of two.

assets/state-mig-manager/0400_configmap.yaml
tariq1890: The thing is I've followed the arrangement of the configs that was already in place. 


I do agree that it's better to group the configs by product . Can I do this in a follow up PR?

tariq1890: In the follow up MR I can do re-arrange the whole config file

assets/state-mig-manager/0400_configmap.yaml
tariq1890: I had the same thought. But when you look at the [config doc](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#h200-profiles) you'll see that there are 2 instances of "3g.71gb". That too adds up to 142GB

assets/state-mig-manager/0400_configmap.yaml
cdesiniotis: Since none of the H200-141GB profiles are applicable to any other SKU (as far as I am aware), I would prefer if we grouped all the H200-141GB configs at the bottom, before the `all-balanced` config, as part of this PR. Then, we can do any additional cleanup in a follow-up. 

The arrangement of the configs gets a bit tricky when some configs / profiles are common between multiple SKUs and some are not.

assets/state-mig-manager/0400_configmap.yaml
cdesiniotis: Ah okay, thanks for pointing me to the official table. Just make sure to verify this config can be applied correctly.

assets/state-mig-manager/0400_configmap.yaml
tariq1890: Sure, I will make the change

tariq1890: @cdesiniotis Please check now

cdesiniotis (approved):

PR #788, opened 2024-06-26
bump helm client to v0.12.10

tariq1890:
This fixes some CVEs that were found in the `helm.sh/helm` dependency

P.S: I am not sure why dependabot isn't raising PRs for most of the gpu-operator deps


cdesiniotis:
> P.S: I am not sure why dependabot isn't raising PRs for most of the gpu-operator deps

I believe dependabot is currently configured to only update our k8s.io dependencies:
```
    groups:
      k8sio:
        patterns:
          - k8s.io/*
        exclude-patterns:
          - k8s.io/klog/*
```

tariq1890:
@cdesiniotis I think that's only the group settings. The stanza however enables all of the go dep bumps. The same config works fine in the other repos

cdesiniotis (approved):

PR #789, opened 2024-06-27
Create driver-ready file atomically

elezar:
This change ensures that the `driver-ready` file is created atomically. This prevents operands from loading an empty file and not applying the relevant settings.

### DIFF
--- a/validator/main.go
+++ b/validator/main.go
@@ -864,16 +864,23 @@ func createStatusFile(statusFile string) error {
 }
 
 func createStatusFileWithContent(statusFile string, content string) error {
-	f, err := os.Create(statusFile)
+	dir := filepath.Dir(statusFile)
+	tmpFile, err := os.CreateTemp(dir, filepath.Base(statusFile)+".*.tmp")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


validator/main.go
```diff
@@ -864,16 +864,23 @@ func createStatusFile(statusFile string) error {
...
 	if err != nil {
-		return fmt.Errorf("unable to write contents of status file %s: %s", statusFile, err)
+		return fmt.Errorf("failed to write temporary status file: %w", err)
 	}
+	defer func() {
+		_ = os.Remove(tmpFile.Name())
+	}()
 
+	if err := os.Rename(tmpFile.Name(), statusFile); err != nil {
+		return fmt.Errorf("error moving temporary file to '%v': %w", statusFile, err)
```
tariq1890, line 882:
```suggestion
		return fmt.Errorf("error moving temporary file to '%s': %w", statusFile, err)
```

cdesiniotis:
Moving to #790

validator/main.go
cdesiniotis: I have address this comment in #790

PR #793, opened 2024-06-27
standardise object hash generation across the repo

tariq1890:
This change removes the need for the `mitchellh/hashstructure` module and sticks to using one hash generator method for the whole project.

We are also using an old version of `mitchellh/hashstructure` and we'd be better off with lesser module imports in general


cdesiniotis (approved):
Thanks @tariq1890!

Issue #802, opened 2024-07-02
Allow management of mig-configs for migManager via values.yaml like it's implemented for devicePlugin


frittentheke:
### 1. Quick Debug Information
* OS/Version(e.g. RHEL8.6, Ubuntu22.04):
* Kernel Version:
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker):
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS):
* GPU Operator Version:

### 2. Issue or feature description
The Helm chart currently does allow to reference the name of a ConfigMap with custom `mig-configs` via https://github.com/NVIDIA/gpu-operator/blob/8f81b78a55ed609d1553eec3ac573f0478de41e7/deployments/gpu-operator/values.yaml#L361, but it cannot manage the content of this ConfigMap, the actual mig-configs.

In comparison a custom devicePlugin configuration and the content of the corresponding ConfigMap can be managed right via the Helm chart, see https://github.com/NVIDIA/gpu-operator/blob/8f81b78a55ed609d1553eec3ac573f0478de41e7/deployments/gpu-operator/values.yaml#L291

When using Helm, having resources not templated and managed and having some configuration outside of the `values.yaml` is somewhat of an anti-pattern.

PR #803, opened 2024-07-02
Allow passing custom mig-parted configmap data in helm chart 

frittentheke:
Closes: #802

### DIFF
--- a/deployments/gpu-operator/templates/mig_config.yaml
+++ b/deployments/gpu-operator/templates/mig_config.yaml
@@ -1,0 +1,14 @@
+{{- if and (.Values.migManager.config.create) (not (empty .Values.migManager.config.data)) }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.migManager.config.name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "gpu-operator.labels" . | nindent 4 }}
+data:
+  config.yaml: |
+    version: v1
+    mig-configs:
+{{- .Values.migManager.config.data | nindent 6 }}
+{{- end }}
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -283,7 +283,7 @@ devicePlugin:
   config:
     # Create a ConfigMap (default: false)
     create: false
-    # ConfigMap name (either exiting or to create a new one with create=true above)
+    # ConfigMap name (either existing or to create a new one with create=true above)
     name: ""
     # Default config name within the ConfigMap
     default: ""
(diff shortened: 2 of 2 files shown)


shivamerla (approved):
lgtm

deployments/gpu-operator/values.yaml
```diff
@@ -358,9 +358,42 @@ migManager:
...
+  #         mig-enabled: true
+  #         mig-devices:
+  #           "3g.40gb": 1
+  #           "4g.40gb": 1
   config:
-    name: "default-mig-parted-config"
     default: "all-disabled"
+    # Create a ConfigMap (default: false)
+    create: false
+    # ConfigMap name (either exiting or to create a new one with create=true above)
```
cdesiniotis: nit: typo
```suggestion
    # ConfigMap name (either existing or to create a new one with create=true above)
```

deployments/gpu-operator/values.yaml
```diff
@@ -358,9 +358,42 @@ migManager:
...
+    name: "default-mig-parted-config"
```
cdesiniotis, line 394: Let's make the default name empty to align with the values for the device-plugin. 
```suggestion
    name: ""
```

deployments/gpu-operator/templates/mig_config.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if and (.Values.migManager.config.create) (not (empty .Values.migManager.config.data)) }}
```
tariq1890, line 1: Won't this suffice?
```suggestion
{{- if and (.Values.migManager.config.create) ( .Values.migManager.config.data) }}
```

cdesiniotis:
@frittentheke can you amend the commit message to better indicate the change introduced? My suggestion: `Allow passing custom mig-parted configmap data in helm chart`.

deployments/gpu-operator/values.yaml
tariq1890: @cdesiniotis I am unsure if we should do this, as we deploy this configmap by default when the mig-manager is enabled.

tariq1890: Ah ok, I do see [here](https://github.com/NVIDIA/gpu-operator/blob/7068fec931e78af2994f82a6ecfea55cbf737b48/controllers/object_controls.go#L1744) that we inject `default-mig-parted-config` when it's unspecified

Please disregard my comment

deployments/gpu-operator/values.yaml
frittentheke: I copied this typo over from the existing comment :see_no_evil: 
I shall fix them both.

frittentheke:
@shivamerla @cdesiniotis @tariq1890 

Thanks a bunch for your review and feedback!
I just pushed another revision fixing your remarks - PTAL.

tariq1890 (approved):

frittentheke:
@tariq1890 @shivamerla there is are two e2e tests failing due to a missing ssh_key:
> ssh key not provided

Could you kindly have a look into this? Would love to have this PR succeed in tests to be ready to merge.

deployments/gpu-operator/values.yaml
cdesiniotis: @frittentheke I do not see this comment being addressed. Can you update this as I have suggested? The `name` field here should only be used when configuring a custom ConfigMap, so defaulting to `default-mig-parted-config` does not make sense.

cdesiniotis (changes requested):
@frittentheke requesting one more change before we can merge this.

cdesiniotis:
> @tariq1890 @shivamerla there is are two e2e tests failing due to a missing ssh_key:
> 
> > ssh key not provided
> 
> Could you kindly have a look into this? Would love to have this PR succeed in tests to be ready to merge.

@frittentheke this is expected as our e2e tests require that your Github repo has valid AWS credentials configured. We can merge this PR without these tests passing.

klueska:
I haven't looked closely, but does this mimic what we do for the device plugin configs?

cdesiniotis:
> I haven't looked closely, but does this mimic what we do for the device plugin configs?

Yes, this mimics what we have for the device-plugin.

frittentheke:
> @frittentheke requesting one more change before we can merge this.


@cdesiniotis This part is actually unchanged to how it was / is prior to this PR and I thought this was fine as it is the same name used by the default config map - see comment by  https://github.com/NVIDIA/gpu-operator/pull/803/files/3e655d7704070c2cc9d7e8afad4bbf08bb774bf0#r1676570719

I gladly change it to some other name if you want me to.

cdesiniotis:
> @cdesiniotis This part is actually unchanged to how it was / is prior to this PR 

I overlooked this in a prior review. Thanks for pointing this out. I can update the name of this field in a follow-up. Thanks for the contribution!

cdesiniotis:
I have opened https://github.com/NVIDIA/gpu-operator/pull/849 to align the `config.name` and `config.data` fields with the device-plugin config.

PR #808, opened 2024-07-03
Configure NVIDIA_CDI_HOOK_PATH envvar in mig-manager 

cdesiniotis:
This change goes hand-in-hand with https://github.com/NVIDIA/mig-parted/pull/90 to support custom install directories configured with the toolkit-container.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -148,6 +148,8 @@ const (
 	CDIEnabledEnvName = "CDI_ENABLED"
 	// NvidiaCTKPathEnvName is the name of the envvar specifying the path to the 'nvidia-ctk' binary
 	NvidiaCTKPathEnvName = "NVIDIA_CTK_PATH"
+	// NvidiaCDIHookPathEnvName is the name of the envvar specifying the path to the 'nvidia-cdi-hook' binary
+	NvidiaCDIHookPathEnvName = "NVIDIA_CDI_HOOK_PATH"
 	// CrioConfigModeEnvName is the name of the envvar controlling how the toolkit container updates the cri-o configuration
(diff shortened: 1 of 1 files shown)


controllers/object_controls.go
```diff
@@ -1371,7 +1371,7 @@ func TransformDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), DeviceListStrategyEnvName, "envvar,cdi-annotations")
 		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIAnnotationPrefixEnvName, "nvidia.cdi.k8s.io/")
 		if config.Toolkit.IsEnabled() {
-			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCTKPathEnvName, filepath.Join(config.Toolkit.InstallDir, "toolkit/nvidia-ctk"))
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCDIHookPathEnvName, filepath.Join(config.Toolkit.InstallDir, "toolkit/nvidia-cdi-hook"))
```
shivamerla, line 1374: does this break backward compatiblity? i.e gpu-operator used with existing device-plugin?

cdesiniotis: Technically yes... I can revert this if desired.

tariq1890 (approved):

controllers/object_controls.go
shivamerla: Given that this is only for the case of CDI enabled, this change seems fine. We rollout device-plugin updates along with operator upgrades with Helm/K8s.

PR #810, opened 2024-07-03
Driver persistence changes

JunAr7112:
Hello!

- [You should sign your work](https://github.com/NVIDIA/gpu-operator/blob/main/CONTRIBUTING.md).

### DIFF
--- a/config/samples/nvidia_v1alpha1_nvidiadriver.yaml
+++ b/config/samples/nvidia_v1alpha1_nvidiadriver.yaml
@@ -6,9 +6,9 @@ spec:
   # use pre-compiled packages for NVIDIA driver installation.
   usePrecompiled: false
   driverType: gpu
-  repository: nvcr.io/nvidia
+  repository: "1837582943" #nvcr.io/nvidia
   image: driver
-  version: "550.90.07"
+  version: "550.54.17"
   imagePullPolicy: IfNotPresent
   imagePullSecrets: []
   nodeSelector: {}
@@ -18,6 +18,8 @@ spec:
     useHostMofed: false
   gds:
     enabled: false
+  persist: true
+  installDirectory: '/opt/nvidia/driver'
   # Private mirror repository configuration
   repoConfig:
     name: ""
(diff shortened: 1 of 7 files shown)


config/samples/nvidia_v1alpha1_nvidiadriver.yaml
```diff
@@ -6,9 +6,9 @@ spec:
   # use pre-compiled packages for NVIDIA driver installation.
   usePrecompiled: false
   driverType: gpu
-  repository: nvcr.io/nvidia
+  repository: "1837582943" #nvcr.io/nvidia
```
shivamerla: please revert this, this was required only for testing using your private image.

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
```diff
@@ -6,9 +6,9 @@ spec:
...
   image: driver
-  version: "550.90.07"
+  version: "550.54.17"
```
shivamerla: this too

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
```diff
@@ -18,6 +18,8 @@ spec:
     useHostMofed: false
   gds:
     enabled: false
+  persist: true
```
shivamerla: we don't want to enable this by default but only when CDI is enabled. Need to handle this in the code to error out accordingly.

manifest.diff
```diff
@@ -0,0 +1,51 @@
+diff --git a/manifests/state-driver/0500_daemonset.yaml b/manifests/state-driver/0500_daemonset.yaml
+index 8ceb7820c..8716d147d 100644
+--- a/manifests/state-driver/0500_daemonset.yaml
++++ b/manifests/state-driver/0500_daemonset.yaml
+@@ -205,6 +205,12 @@ spec:
+         # always use runc for driver containers
+         - name: NVIDIA_VISIBLE_DEVICES
+           value: void
++      {{- if .Driver.Spec.PersistDriver }}
++        - name: RESTARTS_ENABLED
```
shivamerla: update this env to `PERSIST_DRIVER`

manifests/state-driver/0500_daemonset.yaml
```diff
@@ -205,6 +205,12 @@ spec:
         # always use runc for driver containers
         - name: NVIDIA_VISIBLE_DEVICES
           value: void
+      {{- if .Driver.Spec.PersistDriver }}
+        - name: RESTARTS_ENABLED
```
shivamerla: same here. update env.

PR #813, opened 2024-07-05
[validator] address edge case where nvidia-smi on host is empty

cdesiniotis:
This restores the logic that was accidentally removed in https://github.com/NVIDIA/gpu-operator/commit/c46ade78beebc4ae3dd7de934974f5871baa0bd3

### DIFF
--- a/validator/main.go
+++ b/validator/main.go
@@ -693,6 +693,13 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 
 func validateHostDriver(silent bool) error {
 	log.Info("Attempting to validate a pre-installed driver on the host")
+	fileInfo, err := os.Lstat("/host/usr/bin/nvidia-smi")
+	if err != nil {
+		return fmt.Errorf("no 'nvidia-smi' file present on the host: %w", err)
+	}
+	if fileInfo.Size() == 0 {
+		return fmt.Errorf("empty 'nvidia-smi' file found on the host")
+	}
 	command := "chroot"
 	args := []string{"/host", "nvidia-smi"}
 


tariq1890 (dismissed):

validator/main.go
```diff
@@ -693,6 +693,9 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 
 func validateHostDriver(silent bool) error {
 	log.Info("Attempting to validate a pre-installed driver on the host")
+	if fileInfo, err := os.Lstat("/host/usr/bin/nvidia-smi"); err == nil && fileInfo.Size() == 0 {
```
tariq1890: So if the file is not found, we should throw an error. This logic would be in an `else if` clause

cdesiniotis: Updated.

validator/main.go
```diff
@@ -693,6 +693,13 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 
 func validateHostDriver(silent bool) error {
 	log.Info("Attempting to validate a pre-installed driver on the host")
+	fileInfo, err := os.Lstat("/host/usr/bin/nvidia-smi")
+	if err != nil {
+		return fmt.Errorf("no 'nvidia-smi' file present on the host")
```
tariq1890: Let's log the error too.

cdesiniotis: A separate log statement or append the error message to the error that gets returns here?

tariq1890: the latter. Let's wrap the error

cdesiniotis: Done.

tariq1890 (approved):

PR #818, opened 2024-07-10
upgrade golang version to v1.22.5

tariq1890:
This version bump address the following CVE:

https://nvd.nist.gov/vuln/detail/CVE-2024-24791

### DIFF
--- a/validator/versions.mk
+++ b/validator/versions.mk
@@ -16,4 +16,4 @@
 include $(CURDIR)/../versions.mk
 
 CUDA_SAMPLES_VERSION ?= 11.7.1
-GOLANG_VERSION ?= 1.22.2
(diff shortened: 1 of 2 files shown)


tariq1890:
@cdesiniotis  I don't think it is required. The toolchain directive only seems to take effect if the version of the go runtime in use is lesser than the version specified in the `toolchain` directive

> A toolchain directive declares a suggested Go toolchain to use with a module. The suggested Go toolchain’s version cannot be less than the required Go version declared in the go directive. The toolchain directive only has an effect when the module is the main module and the default toolchain’s version is less than the suggested toolchain’s version.

reference: https://go.dev/ref/mod

cdesiniotis (approved):

PR #825, opened 2024-07-11
add prometheus CR permissions back to gpu-operator clusterrole

tariq1890:
This PR fixes a regression introduced by https://github.com/NVIDIA/gpu-operator/pull/750

It seems that the Clusterpolicy controller's informers expect to be able list and watch  ServiceMonitors in the cluster scope when the `ServiceMonitor` CRD is installed.

To reproduce this, install the latest gpu-operator helm chart on a cluster that already has the kube-prometheus stack installed


cdesiniotis (approved):
Thanks @tariq1890 !

Issue #710, opened 2024-04-30
Allow adding custom labels to the "gpu-operator" ServiceMonitor


peihsuant:
### feature description
There is a ServiceMonitor named `gpu-operator` that was automatically created and is owned by the ClusterPolicy. I would like to add custom labels to it so the Prometheus instance with a serviceMonitorSelector can scrape it.

Thanks.

cdesiniotis:
@peihsuant this is already possible today with the `dcgmExporter.serviceMonitor.additionalLabels` field in clusterpolicy.

peihsuant:
Hi @cdesiniotis, what I need to modify is the servicemonitor of `gpu-operator` not `dcgmExporter`, thanks.

cdesiniotis:
@peihsuant apologies, I misread the description. I see we do not have any fields for configuring the `gpu-operator` ServiceMonitor, unlike for `dcgmExporter`. 

If you are interested in working on this, PRs against our gitlab repository are always welcome. https://gitlab.com/nvidia/kubernetes/gpu-operator

csauoss:
@cdesiniotis I would like to give this a try if its still needed.

cdesiniotis:
@csauoss yes this is open. PRs are welcome here https://github.com/NVIDIA/gpu-operator/pulls

csauoss:
@cdesiniotis thank you, but should I create a MR in gitlab(since documentation mentions it) or a PR here is fine as well?

csauoss:
@cdesiniotis I created [MR 1099 in gitlab](https://gitlab.com/nvidia/kubernetes/gpu-operator/-/merge_requests/1099) for this issue for now.  Please review it when you get a chance.

cdesiniotis:
@csauoss we have recently migrated to GitHub and now perform development here. Can you open you PR against https://github.com/NVIDIA/gpu-operator?

> (since documentation mentions it)

Can you point me to where our documentation states this?

csauoss:
Thanks @cdesiniotis! sounds good, will PR here in that case. Here in [CONTRIBUTING.md](https://github.com/NVIDIA/gpu-operator/blob/main/CONTRIBUTING.md#-source-code) file is where I saw references to Gitlab.

csauoss:
[...]

PR #850, opened 2024-07-18
changes to allow custom labels for ServiceMonitor

csauoss:
This pr addresses issue [710](https://github.com/NVIDIA/gpu-operator/issues/710) to allow passing custom labels for gpu-operator's ServiceMonitor.


shivamerla:
changes look good, @csauoss can you rebase please.

csauoss:
> changes look good, @csauoss can you rebase please.

@shivamerla Thanks for taking a look. pulled in changes from main and did the rebase.

csauoss:
@shivamerla can you please review it again when you get a chance? thanks!

ramesius:
@csauoss @shivamerla It would be great to see this land, and updates on getting this into `main`?

csauoss:
@ramesius thanks for bumping this. I have rebased it and is ready for review again. Hoping for @shivamerla @cdesiniotis to take a look when they get a chance.

rahulait:
Re-opening as it was autoclosed due to inactivity. This is a relevant PR and we want to take a look and see if we can get this in.

csauoss:
@rahulait thanks for the update. I will rebase this as well soon.

rahulait:
Thanks @csauoss. I'll try to get this prioritized and provide feedback so that you don't have to keep on rebasing it every now and then.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -984,51 +948,12 @@ type DCGMExporterSpec struct {
 	// Optional: ServiceMonitor configuration for NVIDIA DCGM Exporter
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="ServiceMonitor configuration for NVIDIA DCGM Exporter"
-	ServiceMonitor *DCGMExporterServiceMonitorConfig `json:"serviceMonitor,omitempty"`
+	ServiceMonitor *ServiceMonitorConfig `json:"serviceMonitor,omitempty"`
```
rajathagasthya, line 951: Renaming the struct is a breaking change for consumers importing this type in their Go code. A type alias could work for backward compatibility:

```go
type DCGMExporterServiceMonitorConfig = ServiceMonitorConfig
```

csauoss: good point, thanks! done.

csauoss:
@rahulait @rajathagasthya addressed the comment and have rebased the PR. Please take a look when you get a chance.

controllers/object_controls.go
```diff
@@ -4945,30 +4972,21 @@ func ServiceMonitor(n ClusterPolicyController) (gpuv1.State, error) {
...
-			obj.Spec.Endpoints[0].RelabelConfigs = relabelConfigs
+	if n.stateNames[state] == "state-operator-metrics" {
+		serviceMonitor := n.singleton.Spec.Operator.ServiceMonitor
+		// if ServiceMonitor CRD is missing, assume prometheus is not setup and ignore CR creation
+		if !serviceMonitorCRDExists {
+			logger.V(1).Info("ServiceMonitor CRD is missing, ignoring creation of CR for operator-metrics")
+			return gpuv1.Ready, nil
 		}
+		obj.Spec.NamespaceSelector.MatchNames = []string{obj.Namespace}
+		serviceMonitorCustomEdits(serviceMonitor, obj)
```
rahulait: We should call this only if serviceMonitor is not nil.

csauoss: done

controllers/object_controls.go
```diff
@@ -4945,30 +4972,23 @@ func ServiceMonitor(n ClusterPolicyController) (gpuv1.State, error) {
...
-		if serviceMonitor.Relabelings != nil {
-			relabelConfigs := make([]promv1.RelabelConfig, len(serviceMonitor.Relabelings))
-			for i, relabel := range serviceMonitor.Relabelings {
-				if relabel != nil {
-					relabelConfigs[i] = *relabel
-				}
-			}
-			obj.Spec.Endpoints[0].RelabelConfigs = relabelConfigs
+		obj.Spec.NamespaceSelector.MatchNames = []string{obj.Namespace}
+		if serviceMonitor != nil {
```
rahulait: Should we also check serviceMonitor.IsEnabled() here and only then call serviceMonitorCustomEdits() like we do for dcgm-exporter?

controllers/object_controls.go
```diff
@@ -4945,30 +4972,23 @@ func ServiceMonitor(n ClusterPolicyController) (gpuv1.State, error) {
...
+			serviceMonitorCustomEdits(serviceMonitor, obj)
 		}
 	}
```
rahulait, line 5018: Also, lets add unittest for it like we have for dcgm-exporter

controllers/object_controls.go
csauoss: yeah can add if its needed. It wasn't there for original code and I just moved around the common code.

rahulait: If someone sets something like
```
serviceMonitor:
  enabled: false
  honorLabels: true
```
then we should not enable it by default. We should also keep it disabled. So I think its better to keep things consistent.

csauoss: sounds good

csauoss: @rahulait just a heads-up though, the state-operator-metric was hardcoded to be enabled by default https://github.com/NVIDIA/gpu-operator/blob/main/controllers/state_manager.go#L1142-L1143 unlike state-dcgm-exporter and thats why there was no "serviceMonitor.IsEnabled" check. This change is redundant unless we make  changes in state_manager.go.

rahulait: Ah ok, I think in that case we can leave it like as it is currently. Thanks for pointing it out. Maybe we can add a log there saying its always enabled, so disabling it doesn't take any effect. @tariq1890 any thoughts on this?

controllers/object_controls.go
```diff
@@ -4896,6 +4896,33 @@ func crdExists(n ClusterPolicyController, name string) (bool, error) {
 	return true, nil
 }
 
+func serviceMonitorCustomEdits(desiredState *gpuv1.ServiceMonitorConfig, currentState *promv1.ServiceMonitor) {
```
```suggestion
func applyServiceMonitorCustomEdits(desiredState *gpuv1.ServiceMonitorConfig, currentState *promv1.ServiceMonitor) {
```

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -136,6 +136,44 @@ const (
...
+	// Enabled indicates if ServiceMonitor is deployed
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Enable deployment of ServiceMonitor"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// Interval which metrics should be scraped from. If not specified Prometheus’ global scrape interval is used.
+	// Supported units: y, w, d, h, m, s, ms
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Interval which metrics should be scraped from"
```
```suggestion
	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Interval at which metrics should be scraped"
```

rahulait:
@csauoss after further internal discussions, we are thinking of nesting this field under metrics. So something like:
```
type OperatorSpec struct {
	DefaultRuntime Runtime `json:"defaultRuntime,omitempty"`
	RuntimeClass string `json:"runtimeClass,omitempty"`
	InitContainer InitContainerSpec `json:"initContainer,omitempty"`
	Labels map[string]string `json:"labels,omitempty"`
	Annotations map[string]string `json:"annotations,omitempty"`
	Metrics *OperatorMetricsSpec `json:"metrics,omitempty"`
	UseOpenShiftDriverToolkit *bool `json:"use_ocp_driver_toolkit,omitempty"`
}

type OperatorMetricsSpec struct {
	ServiceMonitor *ServiceMonitorConfig `json:"serviceMonitor,omitempty"`
}
```
So then, it will become something like:
```
operator:
  metrics:
    serviceMonitor:
      enabled: true
      interval: 15s
      honorLabels: true
```
We won't touch for dcgm-exporter as its already there in public. That way, in future, if we have to add more things for metrics of operator, we can add them in that section.

I can also do that change if you want me to add it to your current changes. If you can squash your commits, rebase your branch on top of main of gpu-operator, I can fix that implementation and push a commit for the same.

csauoss:
> type OperatorMetricsSpec struct {
> 	ServiceMonitor *ServiceMonitorConfig `json:"serviceMonitor,omitempty"`
> }

@rahulait sounds good, thank you! I just squashed my commits and branch is all set for you. Let me know if there is anything else I can do.

controllers/object_controls.go
tariq1890: > Maybe we can add a log there saying its always enabled, so disabling it doesn't take any effect.

I am fine leaving it as-is

tariq1890:
@csauoss  We sincerely appreciate your patience and responsiveness here! I've added a few minor review comments, but we are pretty close to merging this :)

controllers/object_controls.go
tariq1890: Since enabled is no-op, let's not expose it in the helm chart template. We can also call out in the comments that it will be ignored

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -46,6 +46,9 @@ spec:
     {{- if .Values.operator.use_ocp_driver_toolkit }}
     use_ocp_driver_toolkit: {{ .Values.operator.use_ocp_driver_toolkit }}
     {{- end }}
+    {{- if .Values.operator.serviceMonitor }}
+    serviceMonitor: {{ toYaml .Values.operator.serviceMonitor | nindent 6 }}
```
tariq1890: Let's add example values in the `values.yaml`. Those values can be commented out in that file

rahulait: updated.

api/nvidia/v1/clusterpolicy_types.go
rahulait: updated.

controllers/object_controls.go
rahulait: updated.

controllers/object_controls.go
rahulait: Yup, added the same and removed from helm template.

csauoss:
> @csauoss We sincerely appreciate your patience and responsiveness here! I've added a few minor review comments, but we are pretty close to merging this :)

@tariq1890 thank you, happy to contribute and look forward to getting this merged!

bundle/manifests/nvidia.com_clusterpolicies.yaml
```diff
@@ -681,25 +681,23 @@ spec:
...
-                          for NVIDIA DCGM Exporter
                         type: boolean
                       honorLabels:
                         description: HonorLabels chooses the metric’s labels on collisions
                           with target labels.
                         type: boolean
                       interval:
                         description: |-
-                          Interval which metrics should be scraped from NVIDIA DCGM Exporter. If not specified Prometheus’ global scrape interval is used.
+                          Interval which metrics should be scraped from. If not specified Prometheus’ global scrape interval is used.
```
tariq1890:
```suggestion
                          Interval at which metrics should be scraped. If not specified, Prometheus’ global scrape interval is used.
```

rahulait: Thanks, updated. It was coming from the comment in api spec. Updated it to have this reflected in all places.

tariq1890 (approved):

Issue #761, opened 2024-06-14
 /sys/module/firmware_class/parameters/path: Read-only file system -- nvidia-vgpu-manager-daemonset


davidhwua:
### 1. Quick Debug Information
* OS/Version:Ubuntu22.04
* Kernel Version: Linux  5.15.0-112-generic
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker): nvida
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): k8s 1.80 
* GPU Operator Version: gpu-operator-v24.3.0v24.3.0 
* Tesla T4
### 2. Issue or feature description

When I install the vgpu-manager, the error appears to be unable to write custom firmware "/run/nvidia/driver/lib/firmware/module/firmware" to GPU server /sys/module/firmware_class/parameters/path.I manually write this line to  /run/nvidia/driver/lib/firmware/module/firmware was passed,can somebody help me how to pass this error?

### 3. Steps to reproduce the issue
`helm install --wait --generate-name       -n gpu-operator --create-namespace       nvidia/gpu-operator      \
         --set sandboxWorkloads.enabled=true      \
         --set vgpuManager.enabled=true   \
          --set toolkit.enabled=true \
          --set vgpuManager.repository=registry.xxxx.com/    \
           --set vgpuManager.image=vgpu-manager \
           --set vgpuManager.version=550.54.10 \
           --set vgpuManager.imagePullSecrets={***}  --timeout 3600s`

### 4. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)
Pods
`root@k8s-gpu1:~# getpod
NAME                                                              READY   STATUS     RESTARTS       AGE
gpu-operator-1718359653-node-feature-discovery-gc-644fbf54zsjcm   1/1     Running    0              19m
gpu-operator-1718359653-node-feature-discovery-master-7f5f4dg2q   1/1     Running    0              19m
gpu-operator-1718359653-node-feature-discovery-worker-cnblz       1/1     Running    0              19m
gpu-operator-1718359653-node-feature-discovery-worker-vhtxp       1/1     Running    0              19m
gpu-operator-66567bfffb-k9ltp                                     1/1     Running    0              19m
nvidia-sandbox-device-plugin-daemonset-48zr6                      0/1     Init:1/2   0              18m
[...]

PR #865, opened 2024-07-24
[vgpu-manager] mount necessary directories so that GSP firmware can be used

cdesiniotis:
This fixes https://github.com/NVIDIA/gpu-operator/issues/761. This change is also required when enabling OpenRM with the vGPU host driver.

### DIFF
--- a/manifests/state-driver/0500_daemonset.yaml
+++ b/manifests/state-driver/0500_daemonset.yaml
@@ -572,6 +572,10 @@ spec:
           - name: host-os-release
             mountPath: /host-etc/os-release
             readOnly: true
+          - name: firmware-search-path
+            mountPath: /sys/module/firmware_class/parameters/path
+          - name: nv-firmware
+            mountPath: /lib/firmware
       {{- end }}
       volumes:
         - name: run-nvidia
--- a/assets/state-vgpu-manager/0210_clusterrole.yaml
+++ b/assets/state-vgpu-manager/0210_clusterrole.yaml
@@ -14,7 +14,6 @@ rules:
   - ""
   resources:
   - nodes
-  - pods
   verbs:
   - get
   - list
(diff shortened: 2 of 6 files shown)


tariq1890 (dismissed):

cdesiniotis:
@tariq1890 I have updated our NVIDIADriver templates to include the fw mounts as well.

manifests/state-driver/0500_daemonset.yaml
```diff
@@ -572,6 +572,12 @@ spec:
           - name: host-os-release
             mountPath: /host-etc/os-release
             readOnly: true
+        {{- if eq .Driver.Spec.DriverType "vgpu-host-manager" }}
```
tariq1890: Don't we need the same if conditional block for the `volumes` as well?

cdesiniotis: No, we always add these volumeMounts in the main container, so no need for a conditional in the `volumes` section. Only for `vgpu-host-manager` do we add this volume mount to the driver-toolkit container.

tariq1890: Understood. In that case, why make it a conditional at all? There shouldn't be any harm having the driver toolkit mount this volume at all times just like the main container, yes?

cdesiniotis: I don't think there would be any side effects, but would need to verify.

tariq1890: I prefer to limit the "if conditional blocks" as much as possible. Moreover, this would also be consistent with how we mount volumes in the main container

cdesiniotis: Removed the conditional.

tariq1890: Thanks!

tariq1890 (approved):

PR #890, opened 2024-08-01
controller-runtime cache should only list-watch resources in the operator namespace

tariq1890:
The controller-runtime `Manager ` object exposes the `cache.Options` struct which contains fields used to set the caching strategy of the controller.

By default, the controller's cache lists and watches namespaced-resources from all namespaces. With this change, we set the caching strategy to only list and watch for K8s resources that are created in the gpu-operator's namespace. 

Note: This change only affects Kubernetes resources that have the namespace-scope. The cluster-scoped resources like Nodes, ClusterPolicy, NVIDIADriver will continue to be cached as before

### DIFF
--- a/cmd/gpu-operator/main.go
+++ b/cmd/gpu-operator/main.go
@@ -27,6 +27,7 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	"go.uber.org/zap/zapcore"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 
 	apiconfigv1 "github.com/openshift/api/config/v1"
 	apiimagev1 "github.com/openshift/api/image/v1"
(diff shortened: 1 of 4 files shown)


cdesiniotis (approved):

cmd/gpu-operator/main.go
```diff
@@ -104,13 +105,23 @@ func main() {
 		Port: 9443,
 	})
 
+	operatorNamespace := os.Getenv("OPERATOR_NAMESPACE")
+	openshiftNamespace := "openshift"
+	cacheOptions := cache.Options{
+		DefaultNamespaces: map[string]cache.Config{
+			operatorNamespace:  {},
+			openshiftNamespace: {},
```
cdesiniotis, line 115: nit: can we add a comment denoting that this is needed for getting / listing `ImageStream` resources in the `openshift` namespace?

tariq1890: For sure!

cdesiniotis (approved):

PR #937, opened 2024-08-17
update base image to the ubi9 variant

tariq1890:


docker/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN set -eux; \
     case "${arch##*-}" in \
         x86_64 | amd64) ARCH='amd64' ;; \
         ppc64el | ppc64le) ARCH='ppc64le' ;; \
-        aarch64) ARCH='arm64' ;; \
+        aarch64 | arm64) ARCH='arm64' ;; \
```
cdesiniotis, line 28: Question -- In what case(s) is arch set to `arm64` instead of `aarch64`?

docker/Dockerfile
```diff
@@ -97,7 +97,7 @@ COPY deployments/gpu-operator/charts/node-feature-discovery/crds/nfd-api-crds.ya
 # Install / upgrade packages here that are required to resolve CVEs
 ARG CVE_UPDATES
 RUN if [ -n "${CVE_UPDATES}" ]; then \
-        yum update -y ${CVE_UPDATES} && \
+        dnf update -y ${CVE_UPDATES} && \
```
cdesiniotis, line 100: Question -- just curious, does `yum` not exist in the ubi9 images?

tests/gpu-pod.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   containers:
     - name: cuda-vector-add
       # https://catalog.ngc.nvidia.com/orgs/nvidia/teams/k8s/containers/cuda-sample
-      image: "nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.7.1-ubuntu20.04"
+      image: "nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda12.5.0"
```
cdesiniotis: Why was this change included in this PR?

validator/versions.mk
```diff
@@ -15,5 +15,5 @@
 # Include global variables from gpu-operator project root
 include $(CURDIR)/../versions.mk
 
-CUDA_SAMPLES_VERSION ?= 11.7.1
+CUDA_SAMPLES_VERSION ?= 12.5.0
```
cdesiniotis: Same question as before -- why was this change included in this PR? If it is not required, can we make this change in a separate PR?

docker/Dockerfile
tariq1890: In some machines the `uname -m` does return arm64. This is the case for the Macbooks based on the M1/M2 chips

docker/Dockerfile
tariq1890: It does, I made this change because I thought we should be consistent with our yum/dnf usage in the Dockerfile. Moreover, `dnf` is the recommended pkg manager tool over `yum`

validator/Makefile
```diff
@@ -87,7 +87,7 @@ $(BUILD_TARGETS): build-%:
 		--build-arg VERSION="$(VERSION)" \
 		--build-arg GIT_COMMIT="$(GIT_COMMIT)" \
 		--build-arg BUILDER_IMAGE="$(BUILDER_IMAGE)" \
-		--build-arg CUDA_SAMPLE_IMAGE=nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda$(CUDA_SAMPLES_VERSION)-$(*) \
```
tariq1890, line 90: @cdesiniotis . If you see here, the cuda samples container image reference resolves to `nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.7.1-ubi8`.  As we are no longer using the `ubi8` tag, I thought it would be preferable to use a cuda-sample image that has a tag without `ubi8`.  Also, this way, the cuda-sample image has nothing to do with the base distribution we use to build the gpu-operator image.

validator/versions.mk
tariq1890: See [here](https://github.com/NVIDIA/gpu-operator/pull/937/files#r1722645761)

tests/gpu-pod.yaml
tariq1890: See [here](https://github.com/NVIDIA/gpu-operator/pull/937/files#r1722645761)

docker/Dockerfile
cdesiniotis: Got it!

validator/Makefile
cdesiniotis: Sure, I understand wanting to move to an image tag that is distro-agnostic. I am more concerned with the CUDA_SAMPLES_VERSION bump. As long as all active driver branches are compatible with 12.5.0, then I am fine with this change.

tariq1890: Yes, that should be the case. Since the e2e tests cover both the 535 and 550 drivers, we can confirm that the new cuda samples image is working for the active driver branches  we support

tariq1890:
@cdesiniotis I have reverted the CUDA_SAMPLES image bump and rebased the PR

.github/workflows/ci.yaml
```diff
@@ -237,9 +237,9 @@ jobs:
           LOWERCASE_REPO_OWNER=$(echo "${GITHUB_REPOSITORY_OWNER}" | awk '{print tolower($0)}')
           echo "LOWERCASE_REPO_OWNER=$(echo "${GITHUB_REPOSITORY_OWNER}" | awk '{print tolower($0)}')" >> $GITHUB_ENV
 
-          echo "OPERATOR_VERSION=${COMMIT_SHORT_SHA}-ubi8" >> $GITHUB_ENV
+          echo "OPERATOR_VERSION=${COMMIT_SHORT_SHA}-ubi9" >> $GITHUB_ENV
```
elezar, line 240: Note: Since we only provide one flavour of image we could just drop the `-ubi9` suffix.

docker/Dockerfile
```diff
@@ -97,7 +97,7 @@ COPY deployments/gpu-operator/charts/node-feature-discovery/crds/nfd-api-crds.ya
 # Install / upgrade packages here that are required to resolve CVEs
 ARG CVE_UPDATES
```
elezar, line 98: In general, I would recommend removing this block entirely. We have largely stopped using this envvar across our projects -- especially since we've streamlined our waiver process.

validator/Dockerfile
```diff
@@ -87,7 +87,7 @@ LABEL vsc-ref=${GIT_COMMIT}
 # Install / upgrade packages here that are required to resolve CVEs
 ARG CVE_UPDATES
```
elezar, line 88: Same here on removing the block.

validator/Makefile
```diff
@@ -41,8 +41,8 @@ OUT_IMAGE_TAG = $(OUT_IMAGE_VERSION)-$(DIST)
 OUT_IMAGE = $(OUT_IMAGE_NAME):$(OUT_IMAGE_TAG)
 
 ##### Public rules #####
-DISTRIBUTIONS := ubi8 ubuntu20.04
-DEFAULT_PUSH_TARGET := ubi8
+DISTRIBUTIONS := ubi9 ubuntu20.04
```
elezar: Question: Do we need two flavours of validator image?

.github/workflows/ci.yaml
tariq1890: Agreed. I am assuming there is a rationale behind the ubi8 suffix? If it's safe enough to drop it, I am all for the change

docker/Dockerfile
tariq1890: Yes, agreed . I think this block was added as a stopgap to quickly fix CVEs at the time

validator/Makefile
tariq1890: I don't think so. I think they are all the same images

docker/Dockerfile
tariq1890: I can make this change in a follow up PR

validator/Makefile
tariq1890: It looks like we only ever push the tags with ubi8 suffix to the NGC prod registry

cdesiniotis: Yeah we don't need to two flavors. Let's just stick to ubi.

.github/workflows/ci.yaml
tariq1890: @cdesionitis @shivamerla thoughts on this?

tariq1890: @cdesiniotis @shivamerla thoughts on this?

shivamerla: yeah, safe to drop `ubi` suffix as we don't reference those tags in the Helm charts or OLM.

tariq1890: @elezar Looks like I'll have to make changes in multiple places to drop the DIST suffix. I'd like to make that change in a follow-up PR. 

I'd like to limit the changeset to ubi9 upgrade in this PR

cdesiniotis (approved):

PR #952, opened 2024-08-26
update operator initContainer and vfioManager images to 12.6.0

tariq1890:


shivamerla:
digests in the CSV are bumped as well?

tariq1890:
Thanks for pointing that out Shiva. I will update the CSV files

cdesiniotis (approved):
We should really look into automating these image bumps..

PR #955, opened 2024-08-27
update dependabot gomod update frequency to daily

tariq1890:


.github/dependabot.yml
```diff
@@ -24,7 +23,7 @@ updates:
     schedule:
       interval: "daily"
     labels:
-      - toolchain-dependencies
```
tariq1890, line 27: `toolchain-dependencies` does not exist as label, so reusing an existing label instead

cdesiniotis (approved):

PR #964, opened 2024-08-29
drop the DIST (ubi9) suffix in image tags

tariq1890:


shivamerla (approved):
lgtm

Makefile
```diff
@@ -278,13 +276,9 @@ else
 # We skip this for the development release
 DEVEL_RELEASE_IMAGE_VERSION ?= devel
 ifneq ($(strip $(VERSION)),$(DEVEL_RELEASE_IMAGE_VERSION))
-push-$(DEFAULT_PUSH_TARGET): push-short
+push-$(DEFAULT_PUSH_TARGET):
```
cdesiniotis: What is the purpose of this now? Can we remove this?

validator/Makefile
```diff
@@ -68,13 +66,9 @@ validator:
 # We skip this for the development release
 DEVEL_RELEASE_IMAGE_VERSION ?= devel
 ifneq ($(strip $(VERSION)),$(DEVEL_RELEASE_IMAGE_VERSION))
-push-$(DEFAULT_PUSH_TARGET): push-short
+push-$(DEFAULT_PUSH_TARGET):
```
cdesiniotis: Same as above comment. Can this be removed?

Makefile
tariq1890: Sure, I am open to removing this. I only kept it as I haven't changed the make target names, only the names of the  tags that are generated as output of these make jobs

cdesiniotis: We can remove this. The actual make target `push-ubi9` is defined in `multi-arch.mk` / `native.mk`.

validator/Makefile
tariq1890: Done

Makefile
tariq1890: Done, please check now

cdesiniotis (approved):

PR #969, opened 2024-08-30
add make target to automate sync'ing of generated crds into helm and olm artifacts

tariq1890:


deployments/gpu-operator/templates/upgrade_crd.yaml
```diff
@@ -86,7 +86,7 @@ spec:
           - /bin/sh
           - -c
           - >
-            kubectl apply -f /opt/gpu-operator/nvidia.com_clusterpolicies_crd.yaml;
+            kubectl apply -f /opt/gpu-operator/nvidia.com_clusterpolicies.yaml;
```
shivamerla, line 89: these file names have to be updated in the helm chart as well, otherwise it will end up copying duplicate file.

tariq1890: Yes, I have updated it in the helm chart. The PR diff highlights it as a rename in the helm chart dir

PR #976, opened 2024-09-05
add gpu driver 560.35.03

tariq1890:


tests/scripts/.definitions.sh
```diff
@@ -17,7 +17,7 @@ TERRAFORM="terraform -chdir=${TERRAFORM_DIR}"
 : ${LOG_DIR:="/tmp/logs"}
 : ${PROJECT:="$(basename "${PROJECT_DIR}")"}
 : ${TEST_NAMESPACE:="test-operator"}
-: ${TARGET_DRIVER_VERSION:="535.183.06"}
+: ${TARGET_DRIVER_VERSION:="560.35.03"}
```
cdesiniotis, line 20: Question: Do we want to always test the default driver version in the operator helm chart?

tariq1890: This is the target driver version used for the driver upgrade tests in our e2e suite

tariq1890: The default driver version for the e2e tests is taken from the values.yaml which is R550 currently

cdesiniotis: Ah, got it.

cdesiniotis (approved):

PR #978, opened 2024-09-06
disable privileged mode for toolkit-validation init containers

tariq1890:
The `privileged: true` was required as the host mounts for the toolkit-validation containers were previously "Bidirectional". However, after this commit 2f156d6528882c68f9c9a13702771cc785afb0a7, the toolkit-validation init-container host mounts have since been changed to "HostToContainer" propagation.

In line with the principle of least privilege, we disable the privileged: true as they "HostToContainer" propagation don't necessitate privileged mode


shivamerla (approved):
lgtm

PR #984, opened 2024-09-10
configure crun as the low-level runtime to prioritise when using CRI-O

tariq1890:


controllers/object_controls.go
```diff
@@ -1218,6 +1222,14 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
 		}
 	}
 
+	// As crun is the default low-level-runtime of CRI-O, we configure the toolkit to favour crun when running the
+	// toolkit setup and initialising the nvidia-container-runtime wrapper binary.
+	// Specify the full paths of the crun and runc executables as the CRI-O installed binaries are not available in the PATH
+	if runtime == gpuv1.CRIO.String() {
+		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCtrRuntimeRuntimesEnvName,
+			"/usr/libexec/crio/crun,/usr/libexec/crio/runc,crun,runc")
```
tariq1890, line 1230: UPDATE: It looks like this assumption does not hold good for ALL cri-o deployments.

 In cases of cri-o on OpenShift: 
 
 1. The low-level runtime binaries are installed in the following locations:

          
          /usr/bin/runc
          /usr/bin/crun
          
2. `runc` is the default runtime as opposed to `crun` in vanilla CRI-O

PR #986, opened 2024-09-11
[RBAC cleanup] move namespaced resources to Role from ClusterRole

tariq1890:
Since limited the caching of resources to the `gpu-operator` and `openshift` namespace, we should now be good to move all namespaced resources to roles from ClusterRole. The GPU Operator should no longer need Cluster-level permissions for most (if not all) of the namespaced resources it interacts with


cdesiniotis (approved):

tariq1890:
Tested and Validated in OpenShift

PR #1006, opened 2024-09-23
vsphere e2e tests setup

shivakunv:


ArangoGutierrez:
Thanks for this @shivakunv but as discussed on the team meeting we are moving away from Vspehre

PR #1007, opened 2024-09-23
Support for the `DevicePluginCDIDevices` feature

jfroy:
This patch adds support for the `DevicePluginCDIDevices` feature gate by adding `spec.operator.useDevicePluginCDIDevicesFeature` to `ClusterPolicy`.  When this field is set, the operator sets the `DEVICE_LIST_STRATEGY` device plug-in environment variable to `cdi-cri`.


jfroy:
@cdesiniotis @elezar

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -148,6 +148,9 @@ type OperatorSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="On OpenShift, enable DriverToolkit image to build and install driver modules"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	UseOpenShiftDriverToolkit *bool `json:"use_ocp_driver_toolkit,omitempty"`
+
+	// UseDevicePluginCDIDevicesFeature indicates if the device plug-in should be configured to use the CDI devices feature
```
elezar, line 152: One question on the UX for this: Should this be under the `cdi` object in the cluster policy?

controllers/object_controls.go
```diff
@@ -1388,7 +1388,11 @@ func TransformDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 	// update env required for CDI support
 	if config.CDI.IsEnabled() {
 		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), DeviceListStrategyEnvName, "envvar,cdi-annotations")
+		if config.Operator.DevicePluginCDIDevicesFeatureEnabled() {
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), DeviceListStrategyEnvName, "cdi-cri")
+		} else {
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), DeviceListStrategyEnvName, "envvar,cdi-annotations")
+		}
 		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIAnnotationPrefixEnvName, "nvidia.cdi.k8s.io/")
```
elezar, line 1406: This block is also not relevant when using `cdi-cri`.

api/nvidia/v1/clusterpolicy_types.go
jfroy: It's an option that only affects the operator, but is related to CDI. I don't know to which it binds more strongly. So I picked one. It's easy enough to change that in the PR, but of course hard after. Maybe other folks can chime in. Ultimately I will defer to you and your team.

controllers/object_controls.go
jfroy: Do you mean to say that `useDevicePluginCDIDevicesFeature` should be stronger than `cdi.enabled`? The current design and implementation is that `useDevicePluginCDIDevicesFeature` does nothing unless `cdi.enabled` is set.

jfroy:
@cdesiniotis I imagine https://github.com/NVIDIA/gpu-operator/pull/1285 obsoletes this PR?

jfroy:
> @cdesiniotis I imagine #1285 obsoletes this PR?

Ah wait, no, it doesn't use "native CDI" but instead relies on annotations. I'll comment more in the internal document about this.

cdesiniotis:
@jfroy I've updated #1285 to use the CRI instead of annotations, so yes, if we proceed with #1285 it will obsolete this PR. See the discussion here: https://github.com/NVIDIA/gpu-operator/pull/1285#discussion_r1963633036

jfroy:
This is no longer needed. 🚀

PR #1019, opened 2024-09-28
set RUNTIME_CONFIG and RUNTIME_SOCKET envars to support new toolkit versions

tariq1890:
This PR includes setting of additional environment variables in container-toolkit asset for forward compatibility with upcoming toolkit releases 

Reference: https://github.com/NVIDIA/nvidia-container-toolkit/pull/704/files

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1291,6 +1291,7 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		configEnvvarName = "CRIO_CONFIG"
 	}
 
+	setContainerEnv(mainContainer, "RUNTIME_CONFIG", DefaultRuntimeConfigTargetDir+sourceConfigFileName)
 	setContainerEnv(mainContainer, configEnvvarName, DefaultRuntimeConfigTargetDir+sourceConfigFileName)
 
 	volMountConfigName := fmt.Sprintf("%s-config", runtime)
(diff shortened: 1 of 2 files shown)


controllers/object_controls.go
```diff
@@ -1309,6 +1310,9 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		sourceSocketFileName := path.Base(runtimeSocketFile)
 		// set envvar for runtime socket
 		var socketEnvvarName string
+		if runtime == gpuv1.Containerd.String() || runtime == gpuv1.Docker.String() {
```
cdesiniotis: We do not need this conditional. At this point, we know `runtimeSocketFile` is not the empty string, so we can unconditionally set the `RUNTIME_SOCKET` envvar.

tariq1890: Ah you're right, thanks

controllers/object_controls.go
```diff
@@ -1309,6 +1310,7 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		sourceSocketFileName := path.Base(runtimeSocketFile)
 		// set envvar for runtime socket
 		var socketEnvvarName string
+		setContainerEnv(mainContainer, "RUNTIME_SOCKET", DefaultRuntimeSocketTargetDir+sourceSocketFileName)
```
cdesiniotis: minor nit: can we move this so both `setContainerEnv()` calls are made in consecutive lines? E.g. move this to line 1319

cdesiniotis (approved):

controllers/object_controls.go
tariq1890: Sure

tariq1890: Done, please check now

PR #1055, opened 2024-10-22
Make the IMEX nodes config file available to GFD

cdesiniotis:


tariq1890 (approved):

assets/gpu-feature-discovery/0500_configmap.yaml
```diff
@@ -0,0 +1,25 @@
...
+      echo "waiting for the driver validations to be ready..."
+      sleep 5
+    done
+
+    set -o allexport
+    cat /run/nvidia/validations/driver-ready
+    . /run/nvidia/validations/driver-ready
+
+    # DRIVER_ROOT_CTR_PATH will either be /host or /driver-root
+    export GFD_IMEX_NODES_CONFIG_FILE=${DRIVER_ROOT_CTR_PATH}/etc/nvidia-imex/nodes_config.cfg
```
tariq1890: Actually, setting this envar seems to be a scope creep: GFD_IMEX_NODES_CONFIG_FILE

tariq1890: This should be an implementation detail contained within the GFD application

ArangoGutierrez (approved):
This makes me think back on Zvonko idea os getting rid of the assets folder and move to helm charts for all sub components.

assets/gpu-feature-discovery/0500_configmap.yaml
ArangoGutierrez: https://github.com/NVIDIA/k8s-device-plugin/blob/main/cmd/gpu-feature-discovery/main.go#L93. ??

tariq1890: This was how I envisioned it. 

So that envar looks okay to me, but I think we should have the logic to construct the part of the IMEX config file from the `DRIVER_ROOT`. The GFD_IMEX_CONFIG_FILE envar if specified would just override the resolved config path from `DRIVER_ROOT`

cdesiniotis: Are you suggesting that `DRIVER_ROOT` be introduced as an environment variable to GFD. And GFD would search for the IMEX config file at this root?

cdesiniotis: Is this discussion out of scope for this PR?

elezar: As I mentioned in the meeting, I think we could implement the "discovery" logic in GFD to find this file in either `/host` or `/driver-root` by default. Basically converting our config file command line argument to a slice instead of s single string.

Something like https://github.com/NVIDIA/k8s-device-plugin/pull/1009

cdesiniotis: Ah okay I see what you meant now... I am open to pushing this logic to GFD. One advantage being that if the IMEX config file is not in the standard location (/etc/nvidia-imex), one could override the GFD environment variable and point it to the non-standard path under `/host` (since we have read-only access to the entire host root fs).

tariq1890: > Are you suggesting that DRIVER_ROOT be introduced as an environment variable to GFD. And GFD would search for the IMEX config file at this root?

yes

shivamerla: There is no alternative to make this file accessible? Can this file be added in the generated CDI spec for injecting into all management containers?

cdesiniotis: I had a similar thought about adding this file to the management CDI spec as well. That may work, but it would only cover the CDI case, which is not the default yet...

cdesiniotis: I have removed setting `GFD_IMEX_NODES_CONFIG_FILE` in the entrypoint script. The logic for searching for this file at `DRIVER_ROOT_CTR_PATH` is now part of the GFD code introduced here: https://github.com/NVIDIA/k8s-device-plugin/pull/1009

tariq1890: Thanks Chris!

tariq1890 (approved):

shivamerla (approved):

PR #1080, opened 2024-10-29
mount imex nodes config

guptaNswati:


assets/state-driver/0500_daemonset.yaml
```diff
@@ -335,3 +337,7 @@ spec:
           hostPath:
             path: /run/nvidia/driver/lib/firmware
             type: DirectoryOrCreate
+        - name: host-imex-config
+	  hostPath:
+            path: /etc/nvidia-imex/nodes_config.cfg
+	    type: File
```
cdesiniotis: We can't unconditionally mount this file as most systems don't have IMEX installed, and thus, this file won't exist. I am also opposed to configuring this volume as `FileOrCreate`, as that would cause this file to be unnecessarily created on all GPU nodes. 

We may want to consider the approach we are leveraging in GFD. That is, mount the host's `/etc` in an init container, and if `/host/etc/nvidia-imex/nodes_config.cfg` exists, copy this file to an emptyDir volume that is shared with the main driver container. Unlike in GFD, we can mount this shared emptyDir volume to `/etc/nvidia-imex` in the driver container.

guptaNswati: Yes make sense.

assets/state-driver/0500_daemonset.yaml
```diff
@@ -88,6 +88,25 @@ spec:
             - name: run-mellanox-drivers
               mountPath: /run/mellanox/drivers
               mountPropagation: HostToContainer
+        - name: check-imex-config
```
tariq1890, line 91: Why does this have to be a separate sidecar container?

Can we not add this logic to the main driver container scripts?

cdesiniotis:
Closing as this is no longer needed.

PR #1091, opened 2024-10-31
Bump k8s-device-plugin, gpu-feature-discovery and node-feature-discovery versions

tariq1890:


deployments/gpu-operator/Chart.yaml
```diff
@@ -19,6 +19,6 @@ keywords:
 
 dependencies:
   - name: node-feature-discovery
-    version: v0.16.5
+    version: v0.16.6
```
ArangoGutierrez, line 22: ++

ArangoGutierrez (approved):
NFD V0.16.6 ++

deployments/gpu-operator/values.yaml
```diff
@@ -242,7 +242,7 @@ devicePlugin:
   enabled: true
   repository: nvcr.io/nvidia
   image: k8s-device-plugin
-  version: v0.17.0-rc.1-ubi9
+  version: v0.17.0-ubi9
```
elezar, line 245: Not critical, but do we want to drop the `-ubi9` suffix?

elezar (approved):

deployments/gpu-operator/values.yaml
tariq1890: Sure, I can make a note of that. I see that we are no longer maintaining distro variants. So there will only a single tag for every release going forward?

PR #1095, opened 2024-10-31
Revert "Revert "Use NV-GHA runners""

cdesiniotis:
This reverts commit 0efdd6d12f70db4e261ea45f3c18d3346049925c.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,15 +15,9 @@
 name: CI
 
 on:
-  pull_request:
-    types:
-      - opened
-      - synchronize
-    branches:
-      - main
-      - release-*
   push:
     branches:
+      - "pull-request/[0-9]+"
       - main
       - release-*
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,15 +15,9 @@
...
-  pull_request:
-    types:
-      - opened
-      - synchronize
-    branches:
-      - main
-      - release-*
   push:
     branches:
+      - "pull-request/[0-9]+"
```
tariq1890, line 20: This change still doesn't make sense to me.  How would this constraint even be satisfied when the PR branches aren't named this way?

cdesiniotis: Me too. @ArangoGutierrez do you know why this change was initially made?

elezar: When the copy-pr bot is enabled it creates a branch named `pull-request/<PR_ID>` (for example https://github.com/NVIDIA/gpu-operator/tree/pull-request/1095 for this PR).

The bot was enabled in https://github.com/NVIDIA/gpu-operator/pull/1068.

ArangoGutierrez (approved):

PR #1111, opened 2024-11-09
enable hostPID in mps-control-daemon

tariq1890:
See [here](https://github.com/NVIDIA/k8s-dra-driver/pull/115) for context

Additionally, as our gpu drivers run in the host's pid namespace, this change would allow for viewing the mps server and workload processes when querying nvidia-smi from the driver context. This is useful for demo and debugging purposes.


cdesiniotis:
LGTM

elezar (approved):

PR #1115, opened 2024-11-15
[mig-manager] add support for H200 NVL

tariq1890:


cdesiniotis (approved):

cdesiniotis:
@tariq1890 we should also make the same change in mig-parted.

PR #1137, opened 2024-11-25
[WIP] Get rid of kubectl depepdency in container image

tobiasgiese:
We should not use the kubectl binary inside our containers because of potential CVEs.
To get rid of the binary we can use the crdutil from the k8s-operator-libs.


go.mod
```diff
@@ -169,14 +169,19 @@ require (
...
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/kustomize/api v0.17.2 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.17.1 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
+
+// DROP BEFORE MERGE!
+// Implements https://github.com/NVIDIA/k8s-operator-libs/pull/58
+// This is only for testing.
+replace github.com/NVIDIA/k8s-operator-libs => github.com/tobiasgiese/k8s-operator-libs v0.0.0-20241125092837-e8a080621717
```
tobiasgiese, line 187: must be dropped before merging, just as a reminder

deployments/gpu-operator/values.yaml
```diff
@@ -77,6 +77,7 @@ operator:
   use_ocp_driver_toolkit: false
   # cleanup CRD on chart un-install
   cleanupCRD: false
+  imageCleanupCRD: bitnami/kubectl:1.31
```
tobiasgiese, line 80: We can use the upstream `bitnami/kubectl` image here as we don't need the CRDs in the container for the cleanup task.

tariq1890: The cleanest approach would be to add a `DeleteCRD` method to the k8s-operator-libs. I think we can revisit this PR after that. Thoughts?

tobiasgiese: Good call, was also thinking about this. Let me add a command to delete CRDs to the cmdutil

Makefile
```diff
@@ -86,6 +86,11 @@ gpu-operator:
 	CGO_ENABLED=0 GOOS=$(GOOS) \
 		go build -ldflags "-s -w -X $(VERSION_PKG).gitCommit=$(GIT_COMMIT) -X $(VERSION_PKG).version=$(VERSION)" -o gpu-operator ./cmd/gpu-operator/...
 
+# Build apply-crds binary
+apply-crds:
+	CGO_ENABLED=0 GOOS=$(GOOS) \
+		go build -ldflags "-s -w -X $(VERSION_PKG).gitCommit=$(GIT_COMMIT) -X $(VERSION_PKG).version=$(VERSION)" -o apply-crds ./cmd/apply-crds/...
+
```
elezar, line 93: In other projects we using the following:
```
CMDS := $(patsubst ./cmd/%/,%,$(sort $(dir $(wildcard ./cmd/*/))))
CMD_TARGETS := $(patsubst %,cmd-%,$(CMDS))

ifneq ($(PREFIX),)
cmd-%: COMMAND_BUILD_OPTIONS = -o $(PREFIX)/$(*)
endif
ifneq ($(shell uname),Darwin)
EXTLDFLAGS = -Wl,--export-dynamic -Wl,--unresolved-symbols=ignore-in-object-files
else
EXTLDFLAGS = -Wl,-undefined,dynamic_lookup
endif
BUILDFLAGS = -ldflags "-s -w '-extldflags=$(EXTLDFLAGS)' -X $(CLI_VERSION_PACKAGE).gitCommit=$(GIT_COMMIT) -X $(CLI_VERSION_PACKAGE).version=$(CLI_VERSION)"
build:
	go build $(BUILDFLAGS) ./...

cmds: $(CMD_TARGETS)
$(CMD_TARGETS): cmd-%:
	go build $(BUILDFLAGS) $(COMMAND_BUILD_OPTIONS) $(MODULE)/cmd/$(*)
```

to auto-generate `make cmd-*` targets for any binary that we place in `cmd`. Would that be a useful addition here?

deployments/gpu-operator/values.yaml
elezar: There may be other requirements in terms of using a thirdparty image. Implementing the required logic in code may be simpler in the long run.

tobiasgiese:
In favor of https://github.com/NVIDIA/gpu-operator/pull/1802

deployments/gpu-operator/templates/upgrade_crd.yaml
```diff
@@ -83,13 +83,12 @@ spec:
...
-          - >
-            kubectl apply -f /opt/gpu-operator/nvidia.com_clusterpolicies.yaml;
-            kubectl apply -f /opt/gpu-operator/nvidia.com_nvidiadrivers.yaml;
+          - /usr/bin/apply-crds
+          args:
+            - --crds-file=/opt/gpu-operator/nvidia.com_clusterpolicies.yaml
+            - --crds-file=/opt/gpu-operator/nvidia.com_nvidiadrivers.yaml
         {{- if .Values.nfd.enabled }}
-            kubectl apply -f /opt/gpu-operator/nfd-api-crds.yaml;
+            - --crds-file=/opt/gpu-operator/nfd-api-crds.yaml
```
elezar, line 91: [bookmark]

PR #1143, opened 2024-11-28
Add DRA driver for IMEX

cdesiniotis:


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -841,6 +843,60 @@ type SandboxDevicePluginSpec struct {
 	Env []EnvVar `json:"env,omitempty"`
 }
 
+// DRADriverSpec defines the properties for the NVIDIA DRA Driver deployment
+// TODO: add 'controller' and 'kubeletPlugin' structs to allow for per-component configuration
```
elezar, line 847: One question: Should we expose `controller` and `kubeletPlugin` as concepts to the user? These seem like internal details and including them here couples the operator and the DRA driver implementation more tightly.

assets/state-dra-driver/0200_clusterrole.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nvidia-dra-driver
+rules:
+# TODO: restrict RBAC for DRA driver
```
elezar: Has this been done upstream yet?

assets/state-dra-driver/0400_deviceclass-imex.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: resource.k8s.io/v1alpha3
```
elezar: I thought we were using the beta API?

assets/state-dra-driver/0500_deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+          key: node-role.kubernetes.io/master
+          operator: Exists
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/control-plane
+          operator: Exists
+      containers:
+      - name: controller
+        image: "FILLED BY THE OPERATOR"
+        imagePullPolicy: IfNotPresent
+        command: ["nvidia-dra-controller", "-v", "6"]
```
elezar: Should we be setting the verbosity here?

assets/state-dra-driver/0600_configmap.yaml
```diff
@@ -0,0 +1,34 @@
...
+    until [[ -f /run/nvidia/validations/driver-ready ]]
+    do
+      echo "waiting for the driver validations to be ready..."
+      sleep 5
+    done
+    
+    set -o allexport
+    cat /run/nvidia/validations/driver-ready
+    . /run/nvidia/validations/driver-ready
+    # TODO: add an alias for DRIVER_ROOT_CTR_PATH in the k8s-dra-driver and remove the below export
```
elezar: Created https://github.com/NVIDIA/k8s-dra-driver/issues/209

assets/state-dra-driver/0700_daemonset.yaml
```diff
@@ -0,0 +1,130 @@
...
+        - name: MASK_NVIDIA_DRIVER_PARAMS
+          value: "false"
+        - name: NVIDIA_VISIBLE_DEVICES
+          value: void
+        - name: CDI_ROOT
+          value: /var/run/cdi
+        - name: NVIDIA_MIG_CONFIG_DEVICES
+          value: all
+        - name: DEVICE_CLASSES
+          value: imex
```
elezar: Is it expected that the `DEVICE_CLASSES` for the `controller` and the `plugin` match?

assets/state-dra-driver/0700_daemonset.yaml
```diff
@@ -0,0 +1,130 @@
...
+      containers:
+      - name: plugin
+        image: "FILLED BY THE OPERATOR"
+        imagePullPolicy: IfNotPresent
+        command: ["/bin/bash", "-c"]
+        args:
+          - /bin/entrypoint.sh
+        env:
+        - name: MASK_NVIDIA_DRIVER_PARAMS
+          value: "false"
```
elezar: Question: Can a user still override this if required?

assets/state-dra-driver/0700_daemonset.yaml
```diff
@@ -0,0 +1,130 @@
...
+          mountPath: /driver-root
+          mountPropagation: HostToContainer
+          readOnly: true
+        - name: host-root
+          mountPath: /host
+          readOnly: true
+        - mountPath: /var/lib/kubelet/plugins_registry
+          name: plugins-registry
+        - mountPath: /var/lib/kubelet/plugins
+          mountPropagation: Bidirectional
```
elezar: Question: Is bidirectional mount propagation really needed in the plugins folder?

assets/state-dra-driver/0700_daemonset.yaml
```diff
@@ -0,0 +1,130 @@
...
+          path: /
+      - name: plugins-registry
+        hostPath:
+          path: /var/lib/kubelet/plugins_registry
+      - name: plugins
+        hostPath:
+          path: /var/lib/kubelet/plugins
+      - name: cdi
+        hostPath:
+          path: /var/run/cdi
```
elezar:
```suggestion
      - name: cdi
        hostPath:
          path: /var/run/cdi
          type: DirectoryOrCreate
```

controllers/object_controls.go
```diff
@@ -147,6 +148,8 @@ const (
 	NvidiaCtrRuntimeCDIPrefixesEnvName = "NVIDIA_CONTAINER_RUNTIME_MODES_CDI_ANNOTATION_PREFIXES"
 	// CDIEnabledEnvName is the name of the envvar used to enable CDI in the operands
 	CDIEnabledEnvName = "CDI_ENABLED"
+	// NvidiaCTKHookPathEnvName is the name of the envvar specifying the path to the 'nvidia-ctk' binary
```
elezar: Comment is incorrect.

elezar: Note we shouldn't need the `nvidia-ctk` path in addition to the `nvidia-cdi-hook` path. They can be used interchangeably.

controllers/object_controls.go
```diff
@@ -1539,6 +1543,55 @@ func TransformSandboxDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
 	return nil
 }
 
+// TransformDRADriverPlugin transforms nvidia-dra-driver-plugin daemonset with required config as per ClusterPolicy
+func TransformDRADriverPlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) error {
```
elezar: Out of scope for this PR: How much merit is there in refactoring these `Transform*` functions to strip out the common logic that is performed for all containers?

controllers/object_controls.go
```diff
@@ -1539,6 +1543,55 @@ func TransformSandboxDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
...
+	}
+	// set/append environment variables for device-plugin container
+	if len(config.DRADriver.Env) > 0 {
+		for _, env := range config.DRADriver.Env {
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
+		}
+	}
+
+	if config.Toolkit.IsEnabled() {
+		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCTKPathEnvName, filepath.Join(config.Toolkit.InstallDir, "toolkit/nvidia-ctk"))
```
elezar, line 1589: Let's rather update the driver to also use the `nvidia-cdi-hook` path. The following should be sufficient:
```suggestion
		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCTKPathEnvName, filepath.Join(config.Toolkit.InstallDir, "toolkit/nvidia-cdi-hook"))
```

elezar: Created https://github.com/NVIDIA/k8s-dra-driver/issues/210 as a follow-up.

controllers/object_controls.go
```diff
@@ -3663,6 +3716,48 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
+
+	// set image pull secrets
+	if len(config.ImagePullSecrets) > 0 {
+		addPullSecrets(&obj.Spec.Template.Spec, config.ImagePullSecrets)
+	}
+
+	return nil
+}
+
+func transformDeployment(obj *appsv1.Deployment, n ClusterPolicyController) error {
```
elezar, line 3739: Question: Why is this not a function defined on a `ClusterPolicyController`?

controllers/object_controls.go
```diff
@@ -3663,6 +3716,48 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
+	logger := n.logger.WithValues("Deployment", obj.Name)
+	transformations := map[string]func(*appsv1.Deployment, ClusterPolicyController) error{
```
elezar: Does `supportedTransformations` or `requiredTransformations` make more sense as a variable name here?

controllers/object_controls.go
```diff
@@ -3663,6 +3716,48 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
+		"nvidia-dra-driver-controller": TransformDRADriverController,
+	}
+
+	t, ok := transformations[obj.Name]
+	if !ok {
+		logger.Info(fmt.Sprintf("No transformation for Deployment '%s'", obj.Name))
```
elezar: We already add the deployment to the logger. Do we need to print it again?

controllers/object_controls.go
```diff
@@ -3663,6 +3716,48 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
+		return nil
+	}
+
+	// apply per operand Deployment config
+	err := t(obj, n)
+	if err != nil {
+		logger.Error(err, "Failed to apply transformation", "resource", obj.Name)
```
elezar: Could we drop the `resource` `obj.Name`?  We already add this when constructing the logger.

controllers/object_controls.go
```diff
@@ -3663,6 +3716,48 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
+		return err
+	}
+
+	return nil
```
elezar:
```suggestion
	return t(obj, n)
```

controllers/object_controls.go
elezar, line 3739: As a summary of my individual comments below what about rewriting this as:
```
func transformDeployment(obj *appsv1.Deployment, n ClusterPolicyController) error {
	logger := n.logger.WithValues("Deployment", obj.Name)
	switch obj.Name {
	case "nvidia-dra-driver-controller":
		return TransformDRADriverController(obj, n)
	default:
		logger.Info("No transformation for object")
		return nil
	}
}
```

controllers/object_controls.go
```diff
@@ -3663,6 +3716,48 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
+	// set image pull secrets
+	if len(config.ImagePullSecrets) > 0 {
+		addPullSecrets(&obj.Spec.Template.Spec, config.ImagePullSecrets)
+	}
+
+	return nil
+}
+
+func transformDeployment(obj *appsv1.Deployment, n ClusterPolicyController) error {
+	logger := n.logger.WithValues("Deployment", obj.Name)
```
elezar: Note that in the calling function we also add the namespace information to the logger.

controllers/object_controls.go
```diff
@@ -3682,6 +3777,12 @@ func Deployment(n ClusterPolicyController) (gpuv1.State, error) {
 		return gpuv1.Disabled, nil
 	}
 
+	err := transformDeployment(obj, n)
+	if err != nil {
```
elezar:
```suggestion
	if err := transformDeployment(obj, n); err != nil {
```

controllers/resource_manager.go
```diff
@@ -23,15 +23,15 @@ import (
 	"sort"
 	"strings"
 
+	secv1 "github.com/openshift/api/security/v1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	nodev1 "k8s.io/api/node/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	drav1 "k8s.io/api/resource/v1alpha3"
```
elezar: Should we not be using the beta API?

assets/state-dra-driver/0500_deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+        env:
+        - name: DEVICE_CLASSES
```
guptaNswati: are we only doing imex? not mig? we need to do some error handling in case when imex daemon is not running. rn  we always expect the imex daemon to be running.

assets/state-dra-driver/0700_daemonset.yaml
guptaNswati: We can set from one of these gpu, mig and imex and plugin will automatically pick it up 
- name: DEVICE_CLASSES
   value: {{ .Values.deviceClasses | join "," }}
   
   here, it seems we are only doing IMEX by default.

assets/state-dra-driver/0400_deviceclass-imex.yaml
guptaNswati: or we need to enable these feature gates in init config https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#enabling-dynamic-resource-allocation

```
$ cat kubeadm-init-config.yaml
apiVersion: kubeadm.k8s.io/v1beta4
kind: ClusterConfiguration
apiServer:
  extraArgs:
  - name: "feature-gates"
     value: "DynamicResourceAllocation=true"
  - name: "runtime-config"
     value: "resource.k8s.io/v1alpha3=true"
controllerManager:
  extraArgs:
  - name: "feature-gates"
     value: "DynamicResourceAllocation=true"
scheduler:
  extraArgs:
  - name: "feature-gates"
     value: "DynamicResourceAllocation=true"
---
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
featureGates:
  DynamicResourceAllocation: true
```

assets/state-dra-driver/0700_daemonset.yaml
```diff
@@ -0,0 +1,130 @@
...
+        configMap:
+          name: nvidia-dra-driver-kubelet-plugin-entrypoint
+          defaultMode: 448
+      - name: run-nvidia-validations
+        hostPath:
+          path: "/run/nvidia/validations"
+          type: DirectoryOrCreate
+      - name: driver-install-dir
+        hostPath:
+          path: "/run/nvidia/driver"
```
guptaNswati: we are still doing host managed drivers for imex, so we need override this path to `/` with --set driver.enabled=false

deployments/gpu-operator/values.yaml
```diff
@@ -294,6 +294,17 @@ devicePlugin:
     # MPS root path on the host
     root: "/run/nvidia/mps"
 
+draDriver:
+  enabled: true
+  repository: ghcr.io/nvidia
```
guptaNswati, line 299: should we mirror it to nvcr.io similar to other images and may be retag with semvar versioning

assets/state-dra-driver/0200_clusterrole.yaml
cdesiniotis: Not yet, see https://github.com/NVIDIA/k8s-dra-driver/blob/main/deployments/helm/k8s-dra-driver/templates/clusterrole.yaml

Issue tracker: https://github.com/NVIDIA/cloud-native-team/issues/134

assets/state-dra-driver/0400_deviceclass-imex.yaml
cdesiniotis: Can you point me to the beta API? I only see `v1alpha3` here: https://pkg.go.dev/k8s.io/api@v0.31.3/resource/v1alpha3

assets/state-dra-driver/0500_deployment.yaml
cdesiniotis: I don't have a strong opinion. I just copied the flags passed in the upstream DRA helm chart: https://github.com/NVIDIA/k8s-dra-driver/blob/32805fec62d6b3269e75ddb0eddeaa5c4d214564/deployments/helm/k8s-dra-driver/templates/controller.yaml#L55

assets/state-dra-driver/0700_daemonset.yaml
cdesiniotis: Yes. The user can override this variable in Clusterpolicy with `draDriver.env`.

assets/state-dra-driver/0700_daemonset.yaml
cdesiniotis: I am not sure. I copied this from the upstream DRA helm chart: https://github.com/NVIDIA/k8s-dra-driver/blob/32805fec62d6b3269e75ddb0eddeaa5c4d214564/deployments/helm/k8s-dra-driver/templates/kubeletplugin.yaml#L99

@klueska do you know if this is required?

assets/state-dra-driver/0500_deployment.yaml
cdesiniotis: The scope of this PR is to only enable IMEX. 

I am not sure what type of error handling you are envisioning, but I believe we need to ensure that the DRA driver daemonset only ever gets scheduled on nodes that are in an IMEX domain. As is, my PR deploys the DRA driver daemonset on all GPU nodes, regardless if they are in an IMEX domain. I can look into updating the nodeAffinity to leverage the IMEX domain label that GFD adds.

assets/state-dra-driver/0700_daemonset.yaml
cdesiniotis: We mount both `/run/nvidia/driver` and the host root `/` into the container, to account for both the driver container and host-managed driver scenarios. In the case of host-managed drivers, we set the `DRIVER_CTR_ROOT_PATH` envvar to `/host` (this is the container path) during the container entrypoint script.

deployments/gpu-operator/values.yaml
cdesiniotis: Yes, this is a placeholder for now. We will update this once we publish the DRA driver image to nvcr.io

api/nvidia/v1/clusterpolicy_types.go
cdesiniotis: I don't believe we should at this point in time. But I can see where having the ability to control the `controller` / `kubeletPlugin` configuration independently could be useful. For example, one may need to bump the `cpu` / `mem` resources for the controller (and not the plugin) to account for larger sized clusters. Open to continue discussion on this and enumerate the list of fields we want to expose in Clusterpolicy.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -841,6 +843,60 @@ type SandboxDevicePluginSpec struct {
...
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Image pull secrets"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:io.kubernetes:Secret"
+	ImagePullSecrets []string `json:"imagePullSecrets,omitempty"`
+
+	// Optional: Define resources requests and limits for each pod
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Resource Requirements"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:resourceRequirements"
+	Resources *ResourceRequirements `json:"resources,omitempty"`
```
cdesiniotis, line 885: Question -- should we include the `resources`, `args`, and `env` fields at this point in time?

assets/state-dra-driver/0400_deviceclass-imex.yaml
cdesiniotis: My mistake, I was looking at the release tag, and not main. I can pull the latest `v1beta1` api package from main, or just wait to update this once K8s 1.32 is out. The release calendar shows 12/11 as the release date. https://www.kubernetes.dev/resources/release/

controllers/object_controls.go
cdesiniotis: There is definitely some merit. We currently apply some common transforms for all DaemonSets here before calling the individual `Transform*` functions: https://github.com/NVIDIA/gpu-operator/blob/58b19546c08722c5c633243c2656df3c902e06a7/controllers/object_controls.go#L718-L729

We could strip out more, but I believe the main issue is that each `Transform*` function is reading configuration from a different data type. E.g. `TransformDRADriverPlugin` reads from a struct of type `DRADriverSpec` while `TransformDriver` reads from a struct of type `DriverSpec`.

controllers/object_controls.go
cdesiniotis: This is much cleaner. Updated as you suggested.

cdesiniotis: No particular reason other than following the precedent we have with similar transform functions we have in this file. Is there a good reason for defining this as a method instead?

assets/state-dra-driver/0700_daemonset.yaml
cdesiniotis: Updated.

controllers/object_controls.go
cdesiniotis: Updated the comment. Will remove this once https://github.com/NVIDIA/k8s-dra-driver/issues/210 is complete.

assets/state-dra-driver/0400_deviceclass-imex.yaml
cdesiniotis: Updated to use v1beta1.

assets/state-dra-driver/0500_deployment.yaml
cdesiniotis: I have updated the nodeAffinity such that the IMEX DRA driver kubelet plugin only gets scheduled on nodes labeled with `nvidia.com/gpu.imex-domain`

assets/state-dra-driver/0200_clusterrole.yaml
cdesiniotis: I see @guptaNswati  is working on restricting the RBAC upstream: https://github.com/NVIDIA/k8s-dra-driver/pull/219. I will pull in those changes once they are finalized.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -54,7 +54,7 @@ type ClusterPolicySpec struct {
 	// DevicePlugin component spec
 	DevicePlugin DevicePluginSpec `json:"devicePlugin"`
 	// DRADriver component spec
-	DRADriver DRADriverSpec `json:"draDriver"`
+	IMEXDRADriver IMEXDRADriverSpec `json:"imexDRADriver"`
```
guptaNswati, line 57: This approach seems better.

api/nvidia/v1/clusterpolicy_types.go
cdesiniotis: Since the IMEX DRA Driver consists of a controller and kubeletPlugin, it feels as if I should update this so that users can configure each component independently.

```
controller:
  resources: {}
  env: []
  
kubeletPlugin:
  resources: {}
  env: []
```

controllers/resource_manager.go
cdesiniotis: This is updated now.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1872,6 +1931,15 @@ func (p *DevicePluginSpec) IsEnabled() bool {
 	return *p.Enabled
 }
 
+// IsEnabled returns true if IMEX DRA Driver is enabled through gpu-operator
+func (d *IMEXDRADriverSpec) IsEnabled() bool {
+	if d.Enabled == nil {
+		// default is true if not specified by user
+		return true
```
empovit, line 1938: You've made it enabled by default. Does it mean that IMEX is common/expected to be widely used?

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -841,6 +843,60 @@ type SandboxDevicePluginSpec struct {
 	Env []EnvVar `json:"env,omitempty"`
 }
 
+// IMEXDRADriverSpec defines the properties for the NVIDIA IMEX DRA Driver deployment
+// TODO: add 'controller' and 'kubeletPlugin' structs to allow for per-component configuration
+type IMEXDRADriverSpec struct {
```
empovit, line 848: If the DRA driver will eventually manage other resources, not just IMEX channels, it will be difficult to move away from the IMEX-specific settings for a DRA image - because of backward compatibility. If it's something that's going to evolve (I just don't know), I'd probably have a general DRA config (controller image, etc.), and under that a section that would control what kinds of resources (classes?) and how the DRA driver could manage. IMEX channels in this case.

cdesiniotis:
Closing in favor of https://github.com/NVIDIA/gpu-operator/pull/1541

PR #1155, opened 2024-12-05
Fix OUT_IMAGE_VERSION set for the validator image in ngc release jobs

cdesiniotis:


elezar (approved):

elezar:
@cdesiniotis do we need to backport this? In which cases does this break things?

tariq1890 (approved):

tariq1890:
@

tariq1890:
@elezar We only need this if we plan on publishing patch releases from release-24.9 or earlier.

PR #1171, opened 2024-12-14
fix: use namespace prefix in gpu operator metrics

googs1025:
- use namespace prefix in gpu operator metrics

### DIFF
--- a/controllers/operator_metrics.go
+++ b/controllers/operator_metrics.go
@@ -61,25 +61,31 @@ const (
 
 	driverAutoUpgradeEnabled  = 1
 	driverAutoUpgradeDisabled = 0
+
+	// operatorMetricsNamespace is the name of the namespace used for the GPU Operator metrics.
+	operatorMetricsNamespace = "gpu_operator"
 )
 
-func initOperatorMetrics(n *ClusterPolicyController) *OperatorMetrics {
+func initOperatorMetrics() *OperatorMetrics {
 	m := &OperatorMetrics{
 		gpuNodesTotal: promcli.NewGauge(
 			promcli.GaugeOpts{
-				Name: "gpu_operator_gpu_nodes_total",
-				Help: "Number of nodes with GPUs",
+				Namespace: operatorMetricsNamespace,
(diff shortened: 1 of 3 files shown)


controllers/operator_metrics.go
```diff
@@ -61,25 +61,31 @@ const (
...
 )
 
-func initOperatorMetrics(n *ClusterPolicyController) *OperatorMetrics {
+func initOperatorMetrics() *OperatorMetrics {
 	m := &OperatorMetrics{
 		gpuNodesTotal: promcli.NewGauge(
 			promcli.GaugeOpts{
-				Name: "gpu_operator_gpu_nodes_total",
-				Help: "Number of nodes with GPUs",
+				Subsystem: gpuOperatorSystemName,
```
tariq1890: I am reading the godoc [here](https://github.com/prometheus/client_golang/blob/main/prometheus/metric.go#L69), and it looks like `Namespace` might be a better fit than `Subsystem`. Can we designate `"gpu-operator"` as the Namespace and update this PR (& commit) accordingly?

tariq1890:
Thanks you for your contribution @googs1025! I have just provided one review comment

controllers/operator_metrics.go
```diff
@@ -61,25 +61,31 @@ const (
 
 	driverAutoUpgradeEnabled  = 1
 	driverAutoUpgradeDisabled = 0
+
+	// gpuOperatorSystemName is the name of the subsystem used for the GPU Operator metrics.
+	gpuOperatorSystemName = "gpu_operator"
```
tariq1890:
```suggestion
	operatorMetricsNamespace = "gpu_operator"
```

controllers/operator_metrics.go
googs1025: Indeed.looking back to docs, `namespace` tend to refer to different components, and `subsystem` refer to different modules within some component.

cdesiniotis (approved):

tariq1890:
@googs1025 can you amend the commit as well?

PR #1180, opened 2024-12-20
Changes required for WSL2

cdesiniotis:


validator/main.go
```diff
@@ -693,6 +693,11 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 
 func validateHostDriver(silent bool) error {
 	log.Info("Attempting to validate a pre-installed driver on the host")
+	if fileInfo, err := os.Lstat("/host/usr/lib/wsl/lib/nvidia-smi"); err == nil && fileInfo.Size() != 0 {
```
elezar: Does it make sense to use the `info` package from `go-nvlib` here instead (for example https://github.com/NVIDIA/k8s-device-plugin/blob/44b263c37e13370f857ef0691b8ab8ff02135c0d/internal/plugin/factory.go#L129).

This will allow for consistent detection across components.

cdesiniotis: I agree with the overall notion of aligning the detection across components, so I have added a commit to use the `info` package as you have suggested.

However, the usage of the platform resolver in the `driver-validator` does not work as expected and returns `unknown` as the platform. This is because the `libdxcore.so` library is not in the ldcache, and likewise, neither is `libnvidia-ml.so`. The platform resolver works in the device-plugin because the NVIDIA Container Toolkit takes care of injecting the `libdxcore.so` library into the container and updating the ldcache. 

Let me know if you have any suggestions to get the platform detection working in the `driver-validator`, otherwise I will revert the latest commit I added in favor of the original logic.

elezar: Looking at the implementiong in the info package, I don't think there's a way to get this working with the current implementation. I think for the time being, let's revert the proposed change and stick with the existing implementation.

cdesiniotis: I have reverted the commit.

elezar (approved):

validator/main.go
```diff
@@ -215,6 +215,8 @@ const (
 	CCCapableLabelKey = "nvidia.com/cc.capable"
 	// appComponentLabelKey indicates the label key of the component
 	appComponentLabelKey = "app.kubernetes.io/component"
+	// wslNvidiaSMIPath indicates the path to the nvidia-smi binary on WSL
+	wslNvidiaSMIPath = "/usr/lib/wsl/lib/nvidia-smi"
```
cdesiniotis, line 219: @tariq1890 as requested, I have created a constant for this path.

cdesiniotis:
@tariq1890 PTAL

tariq1890 (approved):

PR #1199, opened 2025-01-13
add new field kernelModuleType to the API Spec

tariq1890:
This PR adds a new field `kernelModuleType` to the ClusterPolicy and NVIDIADriver APIs. Once this PR is merged, the `useOpenKernelModules` field be rendered as no-op. 

The `kernelModuleType` field supports three values:

- "auto" - Defer to the gpu-operator/driver-container to choose the recommended kernel module type based on the gpu devices and driver version used. This is the recommended option for most users.

- "open" - Choose the OpenRM Kernel Modules

- "proprietary" - Choose the Proprietary Modules


deployments/gpu-operator/values.yaml
```diff
@@ -137,7 +137,11 @@ driver:
     deployDefaultCR: true
     driverType: gpu
     nodeSelector: {}
-  useOpenKernelModules: false
+  kernelModuleType: "auto"
```
cdesiniotis, line 140: Let's also update the example ClusterPolicy CR in the OLM bundle: https://github.com/NVIDIA/gpu-operator/blob/55cb4844409c62c84d2f53160702e5ee3b52f3ce/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L44

tariq1890: Ah yes, thanks for calling that out

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -50,6 +49,8 @@ metadata:
                   "force": false,
                   "timeoutSeconds": 300
                 },
+                "kernelModuleType": "auto",
+                "maxParallelUpgrades": 1,
```
cdesiniotis, line 53: This line was added by accident?

tariq1890: ah yes, thanks for catching that!

cdesiniotis (approved):

PR #1211, opened 2025-01-19
Remove defaultRuntime as it's no longer used.

meskinemohamed:
The defaultRuntime field is no longer used as the runtime detection, so we don't need the field anymore.

### DIFF
--- a/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+++ b/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
@@ -29,7 +29,6 @@ metadata:
           },
           "spec": {
             "operator": {
-              "defaultRuntime": "crio",
               "use_ocp_driver_toolkit": true,
               "initContainer": {
               }
--- a/config/samples/v1_clusterpolicy.yaml
+++ b/config/samples/v1_clusterpolicy.yaml
@@ -4,7 +4,6 @@ metadata:
   name: gpu-cluster-policy
 spec:
   operator:
-    defaultRuntime: crio
     runtimeClass: nvidia
     initContainer: {}
     use_ocp_driver_toolkit: true
--- a/deployments/gpu-operator/templates/clusterpolicy.yaml
+++ b/deployments/gpu-operator/templates/clusterpolicy.yaml
@@ -16,9 +16,6 @@ spec:
     rootFS: {{ .Values.hostPaths.rootFS }}
     driverInstallDir: {{ .Values.hostPaths.driverInstallDir }}
   operator:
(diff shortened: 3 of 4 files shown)


tariq1890:
Thank you for your contribution @meskinemohamed ! 

Indeed that `containerd` is a better default, however, `defaultRuntime` is a field we no longer use in the controller-logic. We glean the runtime from K8s Node Info instead. In light of that, can we remove[ this conditional block](https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/templates/clusterpolicy.yaml#L20) as a part of this PR?

meskinemohamed:
Thank you @tariq1890 I was not aware of this information. I made the necessary changes in my PR in my last commit.

tariq1890:
Thank you @meskinemohamed ! Please squash your commit history and I can then trigger the CI pipelines

tariq1890 (approved):

cdesiniotis:
@meskinemohamed thanks for the contribution. Would you mind also removing this field in a few other places?
CR sample: https://github.com/NVIDIA/gpu-operator/blob/0014db48f224c6b0449375efd598961046bb1dd7/config/samples/v1_clusterpolicy.yaml#L7
OLM bundle: https://github.com/NVIDIA/gpu-operator/blob/0014db48f224c6b0449375efd598961046bb1dd7/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L32

bundle/manifests/nvidia.com_clusterpolicies.yaml
```diff
@@ -1523,14 +1523,6 @@ spec:
                       set by external tools to store and retrieve arbitrary metadata. They are not
                       queryable and should be preserved when modifying objects.
                     type: object
-                  defaultRuntime:
```
tariq1890: @meskinemohamed We can't remove this as it is a change in the CRD's API spec. This will break the deployments of users

deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
```diff
@@ -2393,4 +2393,4 @@ spec:
     served: true
     storage: true
     subresources:
-      status: {}
+      status: {}
```
tariq1890: we need the EOL here. Can you add a new line at the end of the file and then run `make check` to sync the CRDs?

pro-tip: you can restore this file to what it was in the `main` branch, and then run `make check` to sync the changes in other locations

tariq1890 (approved):

cdesiniotis (approved):

PR #1245, opened 2025-01-31
RBAC improvements

cdesiniotis:


tariq1890 (approved):

cdesiniotis:
@tariq1890 I modified one of my commits since your last review, but I forgot to remove auto-merge. PTAL at the additional cleanup I did for pods in the clusterrole. For pod eviction / drain, we only need create permission on the pods/eviction resource.

PR #1254, opened 2025-02-03
[olm] add perms for pods and pods/eviction in operator Role

cdesiniotis:
This change aligns the gpu-operator Role object across our helm chart and OLM bundle


cdesiniotis:
cc @empovit

tariq1890 (approved):

PR #1275, opened 2025-02-14
[Openshift] hide deprecated useOpenKernelModules in the UI

tariq1890:


bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -296,6 +296,12 @@ spec:
           path: imagePullPolicy
           x-descriptors:
             - 'urn:alm:descriptor:com.tectonic.ui:imagePullPolicy'
+        - displayName: UseOpenKernelModules
```
cdesiniotis, line 299: Question -- probably out of scope for this change, but do we want to add an entry here for the new `KernelModuleType` field?

tariq1890: Yes, I think we could. My understanding is that these fields already show up by default, but I am not opposed to adding an entry to the CSV

cdesiniotis: They do show up by default, but I believe without a description. Not a high priority.

cdesiniotis (approved):

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
cdesiniotis: One other question -- how did you update the CSV file? Is there a make target for that? Or is it manual?

cdesiniotis (approved):

PR #1284, opened 2025-02-18
bump go and golangci-lint versions, lint fixes

tariq1890:
Bumps the golang and golangci-lint version. It also fixes the new gosimple lint issue from the golangci-lint update. 

This PR is needed to unblock the gitlab CI builds. See [here](https://gitlab.com/nvidia/kubernetes/gpu-operator/-/pipelines/1672377793)

For context, go [here](https://staticcheck.dev/docs/checks/#S1009)

### DIFF
--- a/versions.mk
+++ b/versions.mk
@@ -19,8 +19,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= v24.9.2
 
-GOLANG_VERSION ?= 1.23.5
+GOLANG_VERSION ?= 1.23.6
 
-GOLANGCI_LINT_VERSION ?= v1.63.4
+GOLANGCI_LINT_VERSION ?= v1.64.5
 
 GIT_COMMIT ?= $(shell git describe --match="" --dirty --long --always 2> /dev/null || echo "")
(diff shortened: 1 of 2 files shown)


versions.mk
```diff
@@ -19,8 +19,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= v24.9.2
 
-GOLANG_VERSION ?= 1.23.5
+GOLANG_VERSION ?= 1.23.6
 
-GOLANGCI_LINT_VERSION ?= v1.63.4
+GOLANGCI_LINT_VERSION ?= v1.64.5
```
ArangoGutierrez, line 24: is there a way we can move this to a place that Dependabot will take care of the bump? maybe a `tools.go` under hack folder?

ArangoGutierrez (approved):
non blocking comment

versions.mk
tariq1890: This is [where](https://github.com/NVIDIA/gpu-operator/blob/43ec0296a94d2c73aa5030a7ae31afc9c5cbdd08/.github/workflows/ci.yaml#L87) we ultimately consume the version. Is it possible to have dependabot bump the version here?

PR #1285, opened 2025-02-19
Use native CDI in container runtimes when supported

cdesiniotis:


tests/scripts/enable-cdi.sh
```diff
@@ -0,0 +1,12 @@
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source ${SCRIPT_DIR}/.definitions.sh
+
+# Import the check definitions
+source ${SCRIPT_DIR}/checks.sh
+
+kubectl patch clusterpolicy/cluster-policy --type='json' -p='[{"op": "replace", "path": "/spec/cdi/enabled", "value": "true"}]'
```
tariq1890:
```suggestion
kubectl patch clusterpolicy/cluster-policy --type='json' -p='[{"op": "replace", "path": "/spec/cdi/enabled", "value": true}]'
```

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1632,19 +1632,19 @@ type VGPUDevicesConfigSpec struct {
...
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
-	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Enable CDI as a mechanism for making GPUs accessible to containers"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Enable CDI as the mechanism for making GPUs accessible to containers"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	Enabled *bool `json:"enabled,omitempty"`
 
-	// Default indicates whether to use CDI as the default mechanism for providing GPU access to containers.
+	// Deprecated: This field is no longer used. Setting cdi.enabled=true will configure CDI as the default mechanism for making GPUs accessible to containers.
```
tariq1890, line 1643: Can we add the marker to hide this from the UI?

```
'urn:alm:descriptor:com.tectonic.ui:hidden'
```

controllers/state_manager.go
```diff
@@ -725,27 +729,26 @@ func (n *ClusterPolicyController) getRuntime() error {
...
+			runtime = rt
+		} else if rt != runtime {
+			n.logger.Error(nil, "Different runtimes on different worker nodes is not supported")
+			return fmt.Errorf("different runtimes on different worker nodes is not supported")
 		}
-		runtime = rt
-		if runtime == gpuv1.Containerd {
-			// default to containerd if >=1 node running containerd
-			break
+		if runtime == gpuv1.Containerd && semver.Compare(version, "v1.7.0") < 0 {
```
tariq1890, line 747: Is it possible to query the CRI API and see if CDI is enabled ? or a similar check? That would be a more robust check IMO. It's possible that we may have forks of containerd running (rehashes based on a vanilla containerd with different versioning). This check wouldn't yield the expected result in those cases

elezar: I think there was some talk about exposing this for use in DRA too, but it may not officially be part of the CRI. 

If I recall correctly, runtimes could send this information in the CRI status: https://github.com/containerd/containerd/blob/6f652853f01ef9ba340a860c2f39edf1701102d1/internal/cri/server/status.go#L34 and https://github.com/cri-o/cri-o/blob/02f3400b358159265d28a37df61be430404925e9/server/runtime_status.go#L15

I would be surprised if the `cdi-enabled` field "magically" propagates there.

controllers/object_controls.go
```diff
@@ -180,6 +180,8 @@ const (
 	// DriverInstallDirCtrPathEnvName is the name of the envvar used by the driver-validator to represent the path
 	// of the driver install dir mounted in the container
 	DriverInstallDirCtrPathEnvName = "DRIVER_INSTALL_DIR_CTR_PATH"
+	// NvidiaRuntimeSetAsDefaultEnvName is the name of the toolkit container env for configuring NVIDIA Container Runtime as the default runtime
+	NvidiaRuntimeSetAsDefaultEnvName = "NVIDIA_RUNTIME_SET_AS_DEFAULT"
```
elezar, line 184: Question. Out of scope: Should we sort these constants to allow us to find them more easily?

controllers/object_controls.go
```diff
@@ -1195,6 +1197,25 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
 	return envVars
 }
 
+func transformToolkitForCDI(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) {
+	if !config.CDI.IsEnabled() {
+		return
+	}
+
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCtrRuntimeCDIPrefixesEnvName, "nvidia.cdi.k8s.io/")
```
elezar: Note: This is would mean that native CDI support for containers will not work when using `cdi-annotations` in the device plugin.

elezar: OK. I see that we set this value based on whether the runtime supports CDI in the device plugin. Does that mean that we need to optionally set that here too.

Note that I think if the runtime supports CDI and it is enabled, we would want to avoid also responding to the annotations in the toolkit.

controllers/object_controls.go
```diff
@@ -1352,6 +1366,27 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 	return nil
 }
 
+func transformDevicePluginForCDI(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) {
+	if !config.CDI.IsEnabled() {
+		return
+	}
+
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), DeviceListStrategyEnvName, "envvar,cdi-annotations")
```
elezar: Question: Under which conditions could we drop `envvar` and / or include `cdi-cri` here?

controllers/object_controls.go
```diff
@@ -2392,13 +2420,15 @@ func getRuntimeClass(config *gpuv1.ClusterPolicySpec) string {
...
 }
 
-func setRuntimeClass(podSpec *corev1.PodSpec, runtime gpuv1.Runtime, runtimeClass string) {
-	if runtime == gpuv1.Containerd {
-		if runtimeClass == "" {
-			runtimeClass = DefaultRuntimeClass
-		}
-		podSpec.RuntimeClassName = &runtimeClass
+func setRuntimeClass(podSpec *corev1.PodSpec, n ClusterPolicyController, runtimeClass string) {
+	if !n.singleton.Spec.CDI.IsEnabled() && n.runtime != gpuv1.Containerd {
```
elezar, line 2429: Are we not also expecting to use a runtimeclass for `cri-o` here? Also, why do we only do this if CDI is enabled?

controllers/state_manager.go
```diff
@@ -161,10 +161,11 @@ type ClusterPolicyController struct {
 	openshift        string
 	ocpDriverToolkit OpenShiftDriverToolkit
 
-	runtime        gpuv1.Runtime
-	hasGPUNodes    bool
-	hasNFDLabels   bool
-	sandboxEnabled bool
+	runtime            gpuv1.Runtime
+	runtimeSupportsCDI bool
```
elezar, line 165: Is `gpuv1.Runtime` just a string. Does it make sense to update that type with `SupportsCDI` and `Version` menbers instead of having separate members here?

controllers/state_manager.go
```diff
@@ -592,9 +593,11 @@ func getRuntimeString(node corev1.Node) (gpuv1.Runtime, error) {
 	case strings.HasPrefix(runtimeVer, "cri-o"):
 		runtime = gpuv1.CRIO
 	default:
-		return "", fmt.Errorf("runtime not recognized: %s", runtimeVer)
+		return "", "", fmt.Errorf("runtime not recognized: %s", runtimeVer)
 	}
-	return runtime, nil
+	version := strings.SplitAfter(runtimeVer, "//")[1]
+	vVersion := strings.Join([]string{"v", strings.TrimPrefix(version, "v")}, "")
```
elezar: I think the following is easier to read?
```suggestion
	vVersion := "v" + strings.TrimPrefix(version, "v")
```

controllers/object_controls.go
cdesiniotis: I can update this so that we set this envvar conditionally.

On a first pass, I didn't think it was required to remove this setting when using native CDI support for containers -- since our device-plugin would be using a different annotation prefix, `cdi.k8s.io`.

controllers/object_controls.go
cdesiniotis: Good question. 

I think we should drop envvar if native CDI is supported. 

As for `cdi-cri`, I think it depends on what versions of k8s we will support moving forward. The `CDIDevices` field in the CRI was alpha in 1.28 and beta since 1.29. So one could argue that we switch to `cdi-cri` altogether in this PR and never configure `cdi-annotations` as the device-list-strategy.

cdesiniotis: What would the behavior be if we configure both `cdi-annotations` and `cdi-cri`? What is the behavior if we attempt to set the CRI field in the allocate response but we are running on k8s < 1.28?

controllers/object_controls.go
cdesiniotis: My intent was to retain the existing behavior when CDI is disabled. That is, use the hook for `cri-o`.

controllers/state_manager.go
cdesiniotis: Assuming this information is not available via the CRI, is there any other way we could potentially get this information instead of checking version strings here in the controller?

tests/scripts/enable-cdi.sh
cdesiniotis: Thanks for catching this. Updated.

api/nvidia/v1/clusterpolicy_types.go
cdesiniotis: Updated.

controllers/object_controls.go
cdesiniotis: I have updated this now to conditionally set this envvar only when native CDI is not supported.

controllers/object_controls.go
cdesiniotis: I have updated this now to drop `envvar` and use `cdi-cri` when native CDI is supported. We fallback to `cdi-annotations` only when runtimes do not support CDI.

controllers/state_manager.go
cdesiniotis: Agree. Updated as you suggested.

controllers/state_manager.go
elezar: For containerd, one could do a config dump and check whether `enable_cdi` and / or the `cdi_spec_dirs` are visible? This is technically part of the runtime config, so it should actually be reported as:
```
    "enableCDI": false,
    "cdiSpecDirs": [
      "/etc/cdi",
      "/var/run/cdi"
    ],
```
via the CRI api. Not sure if this is visible in the context of the GPU Operator though.

controllers/object_controls.go
elezar: One slightly unrelated comment is that this in incompatible whith where we want to get with the NVIDIA Container Toolkit and we should definitely look at transitioning to a config-based mechansim for CRI-O too.

controllers/state_manager.go
cdesiniotis: Ah, I see. What are your thoughts on adding this logic to the toolkit container? That is, the toolkit container would check if CDI is supported in containerd (by doing a config dump), and if supported it would ensure native-CDI is used for workloads by NOT configuring `nvidia` as the default runtime. If CDI is not supported, the toolkit container would fallback to configuring `nvidia` as the default runtime.

elezar: I think that sounds reasonable. Do we just assume native CDI support for CRI-O?

In this mode of operation we would:

1. Read the existing containerd config
2. If `enableCDI` is present
    1. we know that the container engine supports CDI and it can be enabled.
    2. we add the `nvidia` container runtime as a non-default runtime so that a runtime class can be created for management containers.
    3. The device plugin is configured to use the `cdi.k8s.io/` prefix for annotations and the CRI field.
4. If `enableCDI` is not present
    5. we assume an older Containerd version that does not support CDI and we don't enable it
    6. We add the `nvidia` container runtime (in CDI mode) as a `default` runtime.
    7. The device plugin is configured to use the `nvidia.cdi.k8s.io` prefix for annotations.

For `cri-o` we would basically always follow the "If `enableCDI` is present" path above.

Some questions:
* Can we remove `docker` from the list of "supported" runtimes in the toolkit container?
* How do we plan to trigger this behaviour?
* How do we known whether to use annotations or CRI?
* Is there something we're not thinking about for CRI-O?

controllers/object_controls.go
```diff
@@ -1195,6 +1197,28 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
...
+		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCtrRuntimeCDIPrefixesEnvName, "nvidia.cdi.k8s.io/")
+	}
+
+	// When the container runtime supports CDI, we do not configure 'nvidia' as the default runtime.
+	// Instead, we leverage native CDI support in containerd / cri-o to inject GPUs into workloads.
+	// The 'nvidia' runtime will be set as the runtime class for our management containers so that they
+	// get access to all GPUs.
+	if n.runtimeSupportsCDI {
+		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaRuntimeSetAsDefaultEnvName, "false")
+	}
```
elezar, line 1219: So we're saying that we could pull this logic into the toolkit container instead?

cdesiniotis: Yes, that is what I was suggesting. We would need to incorporate this logic in both the toolkit container and device-plugin actually since the configuration of both components is dependent on whether native-CDI is supported.

controllers/state_manager.go
cdesiniotis: > Can we remove docker from the list of "supported" runtimes in the toolkit container?

From the perspective of GPU Operator, I am comfortable saying yes since our next operator release will not support a K8s version that supports docker.

> Do we just assume native CDI support for CRI-O?

This is what I was envisioning as CRI-O has supported CDI since 1.23.2. Since we are shortening our K8s support matrix to n-3 at the time of release, and CRI-O follows the K8s release cycle with respect to minor versions, I think it is relatively safe to always assume native-CDI support for CRI-O. But maybe I am overlooking something.

> How do we plan to trigger this behavior?

I propose triggering this behavior whenever `CDI_ENABLED=true` is set in the toolkit container.

> How do we known whether to use annotations or CRI?

I am assuming this is with respect to the device list strategy we configure in the plugin. I believe if we push the "native-CDI" detection logic to the toolkit container, we would have to do something similar in the plugin. That is, the behavior of the toolkit container and device-plugin when `cdi.enabled=true` in the operator is dependent on whether native-CDI is supported or not. Either we 1) run the same logic in both the toolkit container and the device-plugin to detect if native-CDI is supported, or 2) leverage the `toolkit-ready` status file to communicate information to the device-plugin that was collected in the toolkit container.

I'll try to write-up this proposal in more detail before the weekend.

cdesiniotis:
Closing in favor of https://github.com/NVIDIA/gpu-operator/pull/1578

PR #1302, opened 2025-02-28
Remove all CRDs when `operator.cleanupCRD` is `true`

### LINKED ISSUE #1301: operator.cleanupCRD is not deleting all the CRDs
Installed the GPU operator with --set operator.cleanupCRD=true, but when its uninstalled only `clusterpolicy.nvidia.com` CRD got deleted, but 4 other CRDs,`  nodefeaturegroups`,  ` nodefeaturerules`, ` nodefeatures`, ` nvidiadrivers ` are still present

chipzoller:
Signed-off-by: chipzoller <chipzoller@gmail.com>

Closes #1301

# Changes

Prior to this PR, when a user defined the value `operator.cleanupCRD: true`, only the ClusterPolicy CRD was removed. After these changes, all other CRDs deployable by this chart, in any configuration (except NodeResourceTopology), will also be removed on best effort basis (ignored if not found). These additional CRDs are the following:

* NodeFeature
* NodeFeatureGroup
* NodeFeatureRule
* NVIDIADriver

### DIFF
--- a/deployments/gpu-operator/templates/cleanup_crd.yaml
+++ b/deployments/gpu-operator/templates/cleanup_crd.yaml
@@ -40,6 +40,10 @@ spec:
           - >
               kubectl delete clusterpolicy cluster-policy;
               kubectl delete crd clusterpolicies.nvidia.com;
-
+              kubectl delete crd nvidiadrivers.nvidia.com --ignore-not-found=true;
+              kubectl delete crd nodefeatures.nfd.k8s-sigs.io --ignore-not-found=true;
+              kubectl delete crd nodefeaturegroups.nfd.k8s-sigs.io --ignore-not-found=true;
+              kubectl delete crd nodefeaturerules.nfd.k8s-sigs.io --ignore-not-found=true;
+              kubectl delete crd noderesourcetopologies.topology.node.k8s.io --ignore-not-found=true;
       restartPolicy: OnFailure
 {{- end }}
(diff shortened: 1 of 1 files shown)


deployments/gpu-operator/templates/cleanup_crd.yaml
```diff
@@ -40,6 +40,10 @@ spec:
           - >
               kubectl delete clusterpolicy cluster-policy;
               kubectl delete crd clusterpolicies.nvidia.com;
-
+              kubectl delete crd nvidiadrivers.nvidia.com --ignore-not-found=true;
```
shivamerla, line 43: we can remove these only when NFD is enabled with the gpu-operator.

shivamerla:
@ArangoGutierrez do you see any issues with removing NFD CRs/CRDs cluster wide? Though NFD is un-installed, what is the behavior of services using them.

deployments/gpu-operator/templates/cleanup_crd.yaml
```diff
@@ -40,6 +40,11 @@ spec:
...
```
tariq1890, line 43: We initially chose not to delete this CRD because it could be disruptive for users who may have deployed `NVIDIADriver` CRs out-of-band. Those CRs would get deleted when the CRD is deleted.  

However, I don't feel strongly about it anymore. @cdesiniotis @shivamerla to confirm

cdesiniotis (approved):

tariq1890 (approved):

PR #1273, opened 2025-02-14
Fix migManager data sample config

fprzewozny:
Hey,
Small but needed change, using the current template generation of custom ConfigMap simply fails due to incorrect data type provided to template. Also official documentation shows the correct syntax: [https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-operator-mig.html#example-custom-mig-configuration](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-operator-mig.html#example-custom-mig-configuration)
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-mig-config
data:
  config.yaml: |
    version: v1
    mig-configs:
      all-disabled:
        - devices: all
          mig-enabled: false
      
      five-1g-one-2g:
        - devices: all 
          mig-enabled: true
          mig-devices:
            "1g.10gb": 5
            "2g.20gb": 1
```
Thanks!

PR #1315, opened 2025-03-06
Fixed example in values.yaml

robin-thoni:
The `data` key must be an object, as described below, but the example shows a string.


tariq1890:
Hi @robin-thoni, thank you for your contribution! I believe this is a duplicate of #1273

robin-thoni:
You're right, closing :)

PR #1317, opened 2025-03-06
[GDS][GDRCopy] remove unnecessary liveness probe

tariq1890:
This PR removes the liveness probes from the GDS and GDRCopy containers of the driver-daemonset. We've noticed cases where the livenessProbes timeout due to long response times of the `lsmod` commands which have led to undesirable restarts of the container leaving the driver daemonset in a bad state

### DIFF
--- a/assets/state-driver/0500_daemonset.yaml
+++ b/assets/state-driver/0500_daemonset.yaml
@@ -212,15 +212,6 @@ spec:
           successThreshold: 1
           periodSeconds: 10
           timeoutSeconds: 10
-        livenessProbe:
-          exec:
-            command:
-              [sh, -c, 'lsmod | grep nvidia_fs']
-          periodSeconds: 30
-          initialDelaySeconds: 30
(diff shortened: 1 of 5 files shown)


shivamerla (approved):
LGTM! `lsmod` is a formatted representation of `/proc/modules`, so if it was timing out then there are other serious problems on the system :). Nevertheless we don't expect users to manually unload the modules, so this can be safely removed. At some point MOFED drivers were unloading `nvidia_fs`, just double check with NO team.

cdesiniotis (approved):

tariq1890:
> lsmod is a formatted representation of /proc/modules, so if it was timing out then there are other serious problems on the system :)

Agreed. We've also observed that the lsmod delays are temporary, so restarting the container through a liveness probe is not necessary. The long term fix for this would be to ensure idempotent deploys of the driver containers i.e. if a desired driver module is already loaded, the container will not go through a module unload and load cycle.

PR #1320, opened 2025-03-10
fixed cluster policy reconcile err type

libranwmq:
err = fmt.Errorf("Failed to get ClusterPolicy object: %v", err)
if apierrors.IsNotFound(err) {} this branch never true;


rahulait:
Thanks @libranwmq . In latest main of gpu-operator, this has been updated to:
https://github.com/NVIDIA/gpu-operator/blob/main/controllers/clusterpolicy_controller.go#L98C1-L107C4
```
	if err := r.Get(ctx, req.NamespacedName, instance); err != nil {
		err = fmt.Errorf("failed to get ClusterPolicy object: %w", err)
		r.Log.Error(err, "unable to fetch ClusterPolicy")
		clusterPolicyCtrl.operatorMetrics.reconciliationStatus.Set(reconciliationStatusClusterPolicyUnavailable)
		if apierrors.IsNotFound(err) {
			// Request object not found, could have been deleted after reconcile request.
			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
			// Return and don't requeue
		}
```

Since we now use `%w`, apierrors.IsNotFound(err) detects the nested wrapped error. We no longer need this change. Closing this PR for now as its no longer required. Please feel free to re-open if you have any queries.

PR #1321, opened 2025-03-10
[vgpu-manager] Remove create/update/delete perms for pods in ClusterRole

cdesiniotis:
This was missed in https://github.com/NVIDIA/gpu-operator/commit/6330c78808cedd96ee16e3fc7de46123b8a1457f

### DIFF
--- a/assets/state-vgpu-manager/0210_clusterrole.yaml
+++ b/assets/state-vgpu-manager/0210_clusterrole.yaml
@@ -23,14 +23,15 @@ rules:
   - ""
   resources:
   - pods
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ""
+  resources:
   - pods/eviction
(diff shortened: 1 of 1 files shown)


assets/state-vgpu-manager/0210_clusterrole.yaml
```diff
@@ -23,14 +23,15 @@ rules:
...
-  - create
-  - delete
   - get
   - list
-  - patch
-  - update
+- apiGroups:
+    - ""
+  resources:
+    - pods/eviction
```
tariq1890: let's indent the yaml list items in the same level as the list.

cdesiniotis: Done.

tariq1890 (approved):

PR #1332, opened 2025-03-12
bump k8s-operator-libs, openshift/client-go and openshift/api deps

tariq1890:
Keeping the OpenShift dependencies updated since we need to support OCP 4.18


cdesiniotis:
Is there a reason dependabot did not raise a PR for the openshift deps?

tariq1890:
dependabot doesn't seem to be raising MRs for non-versioned imports. I am not sure if it supports it

cdesiniotis (approved):

PR #1336, opened 2025-03-13
[nvidia-driver] mount /run/nvidia-fabricmanager

tariq1890:
This PR adds support for Blackwell GPUs like B200. 

For GPUs with NVLink5+ switches, the fabricmanager needs to be able to access the `/run/nvidia-fabricmanager/fm_sm_ipc.socket` created on the host.

### DIFF
--- a/assets/state-driver/0500_daemonset.yaml
+++ b/assets/state-driver/0500_daemonset.yaml
@@ -111,6 +111,8 @@ spec:
           - name: run-nvidia
             mountPath: /run/nvidia
             mountPropagation: Bidirectional
+          - name: run-nvidia-fabricmanager
+            mountPath: /run/nvidia-fabricmanager
           - name: run-nvidia-topologyd
             mountPath: /run/nvidia-topologyd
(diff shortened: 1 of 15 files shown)


assets/state-driver/0500_daemonset.yaml
```diff
@@ -111,6 +111,8 @@ spec:
           - name: run-nvidia
             mountPath: /run/nvidia
             mountPropagation: Bidirectional
+          - name: run-nvidia-fabricmanager
```
cdesiniotis, line 114: Question -- we may not know the answer at this point, but do we also need to mount this directory in the `openshift-driver-toolkit-ctr`? I think the answer is likely "no" since we start FM in the main ctr.

tariq1890: Yes, this mount is required by `nv-fabricmanager` daemon. We only mount this in the container which runs that process

cdesiniotis (approved):

PR #1401, opened 2025-04-16
Updated .release:staging to stage images in nvstaging

JunAr7112:
Updated .release:staging to stage images in nvstaging. These changes are needed to ensure compliance with new NGC guidelines.

### DIFF
--- a/.common-ci.yml
+++ b/.common-ci.yml
@@ -135,7 +135,7 @@ trigger-pipeline:
     # Since OUT_IMAGE_NAME and OUT_IMAGE_VERSION are set, this will push the CI image to the
     # Target
     - make push-${DIST}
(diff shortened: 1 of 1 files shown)


.common-ci.yml
```diff
@@ -118,6 +118,7 @@ trigger-pipeline:
     - !reference [.regctl-setup, before_script]
     # We ensure that the OUT_IMAGE_VERSION is set
     - 'echo Version: ${OUT_IMAGE_VERSION} ; [[ -n "${OUT_IMAGE_VERSION}" ]] || exit 1'
+    - 'echo Version: ${OUT_IMAGE_NAME} ; [[ -n "${OUT_IMAGE_NAME}" ]] || exit 1'
```
tariq1890: instead of adding an extra echo statement, you can modify the previous one to print both the image name and the image tag

JunAr7112: Updated this to have one echo check statement.

PR #1417, opened 2025-04-24
Switch to distroless base image

cdesiniotis:


validator/Dockerfile
```diff
@@ -51,22 +51,16 @@ FROM ${CUDA_SAMPLE_IMAGE} AS sample-builder
...
 RUN cp /cuda-samples/vectorAdd /artifacts/vectorAdd
 
-FROM nvcr.io/nvidia/cuda:12.8.1-base-ubi9
+# The C/C++ distroless image is used as a base since the CUDA vectorAdd \
+# sample application depends on C/C++ libraries.
+FROM nvcr.io/nvidia/distroless/cc:v3.1.7-dev
 
-# Remove CUDA libs(compat etc) in favor of libs installed by the NVIDIA driver
-RUN dnf remove -y cuda-*
+USER 0:0
```
tariq1890: Does validator need to run as root?

validator/main.go
```diff
@@ -649,7 +650,7 @@ func setEnvVar(envvars []string, key, value string) []string {
 // For driver container installs, check existence of .driver-ctr-ready to confirm running driver
 // container has completed and is in Ready state.
 func assertDriverContainerReady(silent bool) error {
-	command := "bash"
+	command := "sh"
```
tariq1890: can we move this to a constant? Let's call it `shell`

validator/Dockerfile
cdesiniotis: Hmm it doesn't appear I have anything written in my notes concerning this (I made these changes some time ago). My best guess is that this was needed for the `driver-validation` container where we create the `/dev/char` symlinks. But it is definitely worth validating and removing this if not required.

tariq1890: Sure, once you find out why it's needed and we are certain that there is no way around this, please update the PR description with the details

docker/Dockerfile
```diff
@@ -55,10 +55,14 @@ ARG VERSION="unknown"
 ARG GIT_COMMIT="unknown"
 RUN make gpu-operator
 
-FROM nvcr.io/nvidia/cuda:12.8.1-base-ubi9
+# Download must-gather dependency: `kubectl`
+ARG TARGETARCH
+RUN OS_ARCH=${TARGETARCH/x86_64/amd64} && OS_ARCH=${OS_ARCH/aarch64/arm64} && curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${OS_ARCH}/kubectl
```
tariq1890: Let's merge these two `RUN` statements

validator/main.go
cdesiniotis: Updated.

docker/Dockerfile
cdesiniotis: Updated.

docker/Dockerfile
```diff
@@ -55,10 +55,15 @@ ARG VERSION="unknown"
 ARG GIT_COMMIT="unknown"
 RUN make gpu-operator
 
-FROM nvcr.io/nvidia/cuda:12.8.1-base-ubi9
+# Download must-gather dependency: `kubectl`
```
shivamerla: I think we can avoid packaging `kubectl` binary too. This often have CVEs. For openshift, we can publish a minimal image into RH registry like [these](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/support/gathering-cluster-data#gathering-data-specific-features_gathering-cluster-data). Which is used only during log collection using `oc adm must-gather --image <debug-image>`. For K8s users, they will use the script directly from the GH repo anyway.

validator/Dockerfile
shivamerla: we run `chroot` to check for host driver installation, which needs to be run with root user.

docker/Dockerfile
cdesiniotis: We currently still need it for the `upgrade-crd` and `cleanup-crd` helm hooks.

shivamerla: yes, those can be replaced with the `crd-apply` tool that the NO is using.

tariq1890: For sure, Shiva. We do have that plan. So, It will be done in a follow-up PR.

cmd/nvidia-validator/main.go
```diff
@@ -649,7 +652,7 @@ func setEnvVar(envvars []string, key, value string) []string {
 // For driver container installs, check existence of .driver-ctr-ready to confirm running driver
 // container has completed and is in Ready state.
 func assertDriverContainerReady(silent bool) error {
-	command := "bash"
+	command := shell
```
elezar, line 655: Out of scope for this PR, but it may be cleaner and easy to reason about if we implement a "Runner" here (see something like https://github.com/NVIDIA/nvidia-container-toolkit/blob/81fb7bb9c117d653a94d41e6a8bad7019537db9e/tests/e2e/runner.go#L82-L102) where we could better encapsulate the shell specifics.

docker/Dockerfile
```diff
@@ -84,40 +84,21 @@ LABEL vsc-ref=${GIT_COMMIT}
...
 COPY deployments/gpu-operator/crds/nvidia.com_nvidiadrivers.yaml /opt/gpu-operator/nvidia.com_nvidiadrivers.yaml
 COPY deployments/gpu-operator/charts/node-feature-discovery/crds/nfd-api-crds.yaml /opt/gpu-operator/nfd-api-crds.yaml
 
-# Install / upgrade packages here that are required to resolve CVEs
-ARG CVE_UPDATES
-RUN if [ -n "${CVE_UPDATES}" ]; then \
-        dnf update -y ${CVE_UPDATES} && \
-        rm -rf /var/cache/yum/*; \
-    fi
-
```
elezar, line 120: Let's remove the references to `CVE_UPDATES` in the rest of the codebase as well.

docker/Dockerfile
```diff
@@ -56,17 +56,17 @@ ARG VERSION="unknown"
...
-# Remove CUDA libs(compat etc) in favor of libs installed by the NVIDIA driver
-RUN dnf remove -y cuda-*
+FROM ${CUDA_SAMPLE_IMAGE} AS sample-builder
 
-RUN dnf install -y \
-        kmod \
-        pciutils && \
-    rm -rf /var/cache/yum/*
+# The C/C++ distroless image is used as a base since the CUDA vectorAdd
+# sample application depends on C/C++ libraries.
```
elezar, line 67: Not in scope for this PR, but does this mean that we also use this container to validate functionality? Why do we reference the CUDA samples images in other places then?

docker/Dockerfile
```diff
@@ -84,40 +84,21 @@ LABEL vsc-ref=${GIT_COMMIT}
...
 COPY --from=builder /workspace/gpu-operator /usr/bin/
+COPY --from=builder /workspace/kubectl /usr/bin/
 COPY --from=builder /workspace/nvidia-validator /usr/bin/
 COPY --from=sample-builder /cuda-samples/vectorAdd /usr/bin/vectorAdd
 
-# gpu-operator manifests
-RUN mkdir -p /opt/gpu-operator/manifests
 COPY assets /opt/gpu-operator/
 COPY manifests /opt/gpu-operator/manifests
+COPY validator/manifests /opt/validator/manifests
```
elezar, line 96: Out of scope / nit: Does puting these in `/opt/gpu-operator/validator/manifests` makes things clearer when one is in the container?

docker/Dockerfile
```diff
@@ -56,17 +56,17 @@ ARG VERSION="unknown"
...
-FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi9
+# Download must-gather dependency: `kubectl`
+ARG TARGETARCH
+RUN OS_ARCH=${TARGETARCH/x86_64/amd64} && OS_ARCH=${OS_ARCH/aarch64/arm64} && \
+    curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${OS_ARCH}/kubectl && \
+    chmod +x ./kubectl
 
-# Remove CUDA libs(compat etc) in favor of libs installed by the NVIDIA driver
-RUN dnf remove -y cuda-*
+FROM ${CUDA_SAMPLE_IMAGE} AS sample-builder
```
elezar: One note about this is that using a build arg for this doesn't allow automatic updating of the docker images.

docker/Dockerfile
```diff
@@ -84,40 +84,21 @@ LABEL vsc-ref=${GIT_COMMIT}
...
 COPY assets /opt/gpu-operator/
 COPY manifests /opt/gpu-operator/manifests
+COPY validator/manifests /opt/validator/manifests
 
-# validator manifests
-RUN mkdir -p /opt/validator/manifests
-COPY validator/manifests/plugin-workload-validation.yaml /opt/validator/manifests
-COPY validator/manifests/cuda-workload-validation.yaml /opt/validator/manifests
-
-RUN mkdir /licenses && mv /NGC-DL-CONTAINER-LICENSE /licenses/NGC-DL-CONTAINER-LICENSE
```
elezar, line 100: Do we no longer need the license?

cdesiniotis: I removed this line since this particular license file is not present in the distroless image. 

@shivamerla do you know if we actually need this license? Or is it sufficient to just include the source code license (Apache 2.0) as you do in the NIM Operator: https://github.com/NVIDIA/k8s-nim-operator/blob/fc92202505b72fb0c9a4f62febc87c45df825dc8/deployments/container/Dockerfile#L66

shivamerla: we have the EULA added here: https://catalog.ngc.nvidia.com/orgs/nvidia/teams/cloud-native/containers/k8s-nim-operator, so remove this from the container itself. For gpu-operator better to double check if this is required as driver deployment etc are involved.

cdesiniotis: On NGC, we include a link to the EULA for the GPU Operator as well: https://catalog.ngc.nvidia.com/orgs/nvidia/containers/gpu-operator

I have gone ahead and added the Apache 2.0 license to the `/licenses` folder in the image.

docker/Dockerfile
cdesiniotis: Good point. I have updated this so that we directly refer to the image here to allow for automatic updates.

docker/Dockerfile
cdesiniotis: I have removed all references to `CVE_UPDATES` now.

docker/Dockerfile
cdesiniotis: This is a fair question that definitely warrants discussion. I would need to sync up with @shivamerla to better understand the motivation behind the current CUDA validation workflow. This is out of scope for this PR.

As I currently understand it, we do not actually deploy the CUDA samples image at runtime -- we only use it at build time to extract the `vectorAdd` binary and include it in the validator image. At runtime, we perform "CUDA validation" by executing the `vectorAdd` sample that is included in the validator image.

cdesiniotis:
I have updated the Makefile and CI/CD definitions to no longer reference `ubi9` (this was missed in my first iteration).

.gitlab-ci.yml
elezar, line 1: Out of scope: We can also remove this file from this repo.

docker/Dockerfile
elezar: OK. That is good to know. I suppose that's why the sample images have not really been kept up to date.

One thing to be aware of here is whether CUDA Forward Compatibility might then become an issue for this image -- especially since we are removing the CUDA Forward Compatibility libraries from this image by switching to distroless.

cdesiniotis: Thanks for mentioning CUDA Forward Compatibility. Without the compat libraries, the samples would fail to run when the major version of the CUDA driver installed is **_less than_** the major version that the CUDA sample is built against, correct?

What about including the CUDA compat libs in the final image like below?
```
FROM nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda12.5.0 AS sample-builder

FROM nvcr.io/nvidia/distroless/cc:v3.1.7-dev

COPY --from=sample-builder /usr/local/cuda/compat /usr/local/cuda/compat
```

tariq1890: "Copying the cuda compat libraries" is a reasonable approach to me. As I understand it, we need it so our current validation workflows don't break. Given that the validator image runs the `vectorAdd` binary to perform validation, it makes sense to add it here as we are no longer using a separate "operator-validator" image.

docker/Dockerfile
```diff
@@ -56,17 +55,17 @@ ARG VERSION="unknown"
 ARG GIT_COMMIT="unknown"
 RUN make cmds
 
-FROM ${CUDA_SAMPLE_IMAGE} AS sample-builder
-
-FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi9
+# Download must-gather dependency: `kubectl`
+ARG TARGETARCH
+RUN OS_ARCH=${TARGETARCH/x86_64/amd64} && OS_ARCH=${OS_ARCH/aarch64/arm64} && \
```
tariq1890, line 60: Once you rebase this PR, please ensure that you use the newer kubectl download URL as seen in this PR - https://github.com/NVIDIA/gpu-operator/pull/1493

docker/Dockerfile
cdesiniotis: I went ahead and made this change. However, the [image builds for ARM](https://github.com/NVIDIA/gpu-operator/actions/runs/15935880646/job/44955747604?pr=1417#step:6:276) are failing with:
```
>>> COPY --from=sample-builder /usr/local/cuda/compat /usr/local/cuda/compat
  .
  .
  .
"/usr/local/cuda/compat": not found
```
I inspected the CUDA base images for ARM, and it looks like the compat libs are only included starting with 12.5.1. The base images I inspected that are <= 12.5.0 don't have the compat libs.

@elezar @tariq1890 what are your thoughts on if we published new CUDA samples images for the latest version of CUDA (or a more recent version)? That would unblock this PR and allow us to include the compat libs in the operator image.

docker/Dockerfile
```diff
@@ -84,40 +83,24 @@ LABEL vsc-ref=${GIT_COMMIT}
 
 WORKDIR /
 COPY --from=builder /workspace/gpu-operator /usr/bin/
+COPY --from=builder /workspace/kubectl /usr/bin/
 COPY --from=builder /workspace/nvidia-validator /usr/bin/
 COPY --from=sample-builder /cuda-samples/vectorAdd /usr/bin/vectorAdd
+COPY --from=sample-builder /usr/local/cuda/compat /usr/local/cuda/compat
```
tariq1890:
```suggestion
COPY --from=builder /usr/local/cuda/compat /usr/local/cuda/compat
```

docker/Dockerfile
cdesiniotis: I have rebased from `main`. the new URL is being used now.

docker/Dockerfile
cdesiniotis: Applied this change.

docker/Dockerfile
```diff
@@ -56,17 +55,16 @@ ARG VERSION="unknown"
 ARG GIT_COMMIT="unknown"
 RUN make cmds
 
-FROM ${CUDA_SAMPLE_IMAGE} AS sample-builder
-
-FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi9
+# Install must-gather dependency: `kubectl`
+ARG TARGETARCH
+RUN OS_ARCH=${TARGETARCH/x86_64/amd64} && OS_ARCH=${OS_ARCH/aarch64/arm64} && curl -LO https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/${OS_ARCH}/kubectl
```
tariq1890: Let's merge these two `RUNs` into one

cdesiniotis: Updated

docker/Dockerfile
elezar: That should be fine. One concern would be whether we'd be introducing an effective minimum driver version of 525. (https://docs.nvidia.com/deploy/cuda-compatibility/#use-the-right-cuda-forward-compatibility-package).

I'm not sure what the expectation is with regards to driver support in the GPU Operator, but if this is not a problem, then moving to newer images is fine.

cdesiniotis: In general, our stance has been that we support active (currently supported) datacenter driver branches. So R525 is out of scope today. (https://docs.nvidia.com/datacenter/tesla/drivers/#supported-drivers-and-cuda-toolkit-versions)

docker/Dockerfile
jgehrcke: Good discussion. This feels like the right solution.

Issue #1421, opened 2025-04-25
Prometheus unable to scrape stats as the scrape port annotation is not set on the dcgm-exporter service


DominicWatson:
**Describe the bug**

No metrics collected by prometheus that is collecting on a default port of 9100.

**Expected behavior**

Metrics collected by prometheus from the dcgm-exporter pods.

**To Reproduce**

Tough one, as I guess there are a lot of ways in which telemetry/prometheus can be configured. Our prometheus is external to the cluster and collects metrics from multiple clusters. However, the default port that our Prometheus setup is scraping metrics on is 9100. We cannot change that just for the nvidia-gpu-operator.

The dcgm-exporter service is running on port 9400 and this does not appear to be configurable. It also uses the annotation `prometheus.io/scrape: true` but does not also include `prometheus.io/port: 9400` - presumably as that is some default that other environments are using? If I manually edit the service and specify this port annotation, our prometheus setup starts to correctly collect stats.

**Environment (please provide the following information):**
 - GPU Operator Version: v25.3.0
 - Kubernetes Distro and Version: EKS 1.31.5

**Summary**

Of course, this may not be a bug and likely me missing a heap of context. BUT, is there a good reason not to include `prometheus.io/port: 9400` annotation in the service manifest and could it be added if not?

Manually editing the service does not appear to be a permanent workaround as it appears to get rewritten after a short time and metrics stop being collected again.

cdesiniotis:
Closing as this is a duplicate of https://github.com/NVIDIA/gpu-operator/issues/1716. Feel free to continue the discussion there.

PR #1422, opened 2025-04-25
#1421 Add port annotation for prometheus

DominicWatson:
Relates to issue #1421

Add port annotation for Prometheus to use when scraping for metrics. This should allow support for more hosting scenarios where Prometheus may be trying a default port, e.g. 9100, rather than trying the ports defined on the service automatically.


rahulait:
Adding custom annotations to dcgm-exporter daemonset has been done in https://github.com/NVIDIA/gpu-operator/issues/2271. Closing this PR. Please feel free to open an issue with your requirements if that doesn't help you.

PR #1424, opened 2025-05-02
Add unit test for TransformNodeStatusExporter

shivakunv:


controllers/transforms_test.go
```diff
@@ -1163,3 +1163,45 @@ func TestTransformSandboxValidator(t *testing.T) {
...
+
+func TestTransformNodeStatusExporter(t *testing.T) {
+	testCases := []struct {
+		description   string
+		ds            Daemonset
+		cpSpec        *gpuv1.ClusterPolicySpec
+		expectedDs    Daemonset
+		errorExpected bool
+	}{
+		{
```
cdesiniotis, line 1175: Can we add one test case before this where the `NodeStatusExporterSpec` is empty in ClusterPolicy? An error will be expected.

controllers/transforms_test.go
```diff
@@ -1163,3 +1163,45 @@ func TestTransformSandboxValidator(t *testing.T) {
...
+			description: "node status exporter",
```
cdesiniotis:
```suggestion
			description: "valid node status exporter spec",
```

shivakunv:
PTAL @cdesiniotis

controllers/transforms_test.go
shivakunv: not added the empty node status spec ( below code)- 
```
{
			description: "empty node status exporter spec",
			ds: NewDaemonset().
				WithContainer(corev1.Container{Name: "dummy"}),
			cpSpec: &gpuv1.ClusterPolicySpec{
				NodeStatusExporter: gpuv1.NodeStatusExporterSpec{},
			},
			expectedDs: NewDaemonset().
				WithContainer(corev1.Container{Name: "dummy"}),
},
```

due to error:    

Failed for function imagePath which expects non-empty image Path 

> empty image path provided through both ClusterPolicy CR and ENV VALIDATOR_IMAGE

cdesiniotis: Yes, an error is expected. You can set `expectedError=true` to signify that an error is expected when running this test case.

shivakunv: Thanks . Added

cdesiniotis (approved):

cdesiniotis:
@shivakunv please squash your commits.

PR #1443, opened 2025-05-15
Bump Device Plugin 0.17.2

JunAr7112:
Updated the bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml, config/manifests/bases/gpu-operator.clusterserviceversion.yaml, and deployments/gpu-operator/values.yaml files so that the k8s-device-plugin version is 0.17.2.

### DIFF
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -245,7 +245,7 @@ devicePlugin:
   enabled: true
   repository: nvcr.io/nvidia
   image: k8s-device-plugin
-  version: v0.17.1
+  version: v0.17.2
   imagePullPolicy: IfNotPresent
   imagePullSecrets: []
   args: []
@@ -359,7 +359,7 @@ gfd:
   enabled: true
   repository: nvcr.io/nvidia
   image: k8s-device-plugin
-  version: v0.17.1
+  version: v0.17.2
   imagePullPolicy: IfNotPresent
   imagePullSecrets: []
   env:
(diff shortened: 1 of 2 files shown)


config/manifests/bases/gpu-operator.clusterserviceversion.yaml
```diff
@@ -48,7 +48,7 @@ spec:
     - name: driver-image
       image: nvcr.io/nvidia/driver@sha256:017d2e2a1bc410e9d2ac116683fc49a6f28e935cadee86e712cc147ea0239469
     - name: device-plugin-image
-      image: nvcr.io/nvidia/k8s-device-plugin@sha256:ea353fe57628f4d3c3e1cda126455d02f2277e3ee92fa9da441d409d76b2ac92
+      image: nvcr.io/nvidia/k8s-device-plugin@sha256:037160a36de0f060fc21cc0cb2f795d980282ff1471b55530433ca4350b24c4f
```
cdesiniotis: While this change is technically accurate, let's revert this change since this file has not been updated in 4 years. We only maintain the CSV files under `bundle/`.

deployments/gpu-operator/values.yaml
```diff
@@ -245,7 +245,7 @@ devicePlugin:
   enabled: true
   repository: nvcr.io/nvidia
   image: k8s-device-plugin
-  version: v0.17.1
+  version: v0.17.2
```
cdesiniotis, line 248: We also need to update the image used for `gfd`. See line 362.

cdesiniotis:
@JunAr7112 also please sign your commit: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

deployments/gpu-operator/values.yaml
JunAr7112: Alright, I updated this as well

config/manifests/bases/gpu-operator.clusterserviceversion.yaml
JunAr7112: Ok I reverted this and also signed with GPG.

tariq1890 (approved):

cdesiniotis (approved):

PR #1450, opened 2025-05-19
bump cuda sample image to use CUDA toolkit 12

tariq1890:


validator/Makefile
```diff
@@ -75,7 +75,7 @@ $(BUILD_TARGETS): build-%:
 		--build-arg VERSION="$(VERSION)" \
 		--build-arg GIT_COMMIT="$(GIT_COMMIT)" \
 		--build-arg BUILDER_IMAGE="$(BUILDER_IMAGE)" \
-		--build-arg CUDA_SAMPLE_IMAGE=nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda$(CUDA_SAMPLES_VERSION)-ubi8 \
+		--build-arg CUDA_SAMPLE_IMAGE=nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda$(CUDA_SAMPLES_VERSION) \
```
cdesiniotis, line 78: I am not quite following this change. Is the dist suffix no longer required?

tariq1890: I am using the ubuntu22.04 dist variant. The tag without the distro suffix is the same image as the one with `ubuntu22.04` suffix. See [here](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/k8s/containers/cuda-sample/tags)

tariq1890: I made the switch to the `ubuntu22.04` distro variant as it's more recent than `ubi8`

PR #1458, opened 2025-05-28
Updated to use distroless kata image

JunAr7112:
Updated the daemonset and configmap to use the distroless kata image.


assets/state-kata-manager/0600_daemonset.yaml
```diff
@@ -66,23 +66,23 @@ spec:
...
+                  [ -f "${pid_file}" ] && [ -d "/proc/$(cat $pid_file)" ] && exit 0 || exit 1
             initialDelaySeconds: 5
             periodSeconds: 5
       terminationGracePeriodSeconds: 30
       volumes:
         - name: nvidia-kata-manager-entrypoint
           configMap:
             name: nvidia-kata-manager-entrypoint
-            defaultMode: 448
+            defaultMode: 493
```
cdesiniotis: I don't believe this change should be needed if the `k8s-kata-manager` is running with uid 0 and gid 0. Can you please verify?

assets/state-kata-manager/0600_daemonset.yaml
```diff
@@ -66,23 +66,23 @@ spec:
...
         - name: nvidia-kata-manager-config
           configMap:
             name: nvidia-kata-manager-config
-            defaultMode: 448
+            defaultMode: 493
```
cdesiniotis: I don't believe this change should be needed if the `k8s-kata-manager` is running with uid 0 and gid 0. Can you please verify?

assets/state-kata-manager/0600_daemonset.yaml
JunAr7112: I just verified that 493 is no longer needed for uid 0. Reverting this.

assets/state-kata-manager/0600_daemonset.yaml
JunAr7112: I just verified that 493 is no longer needed for uid 0. Reverting this.

cdesiniotis:
Closing in favor of https://github.com/NVIDIA/gpu-operator/pull/1496

PR #1462, opened 2025-05-29
chore(deploy/chart): remove enableNodeFeatureApi option from nfd

hhk7734:
Since NodeFeatureAPI is enabled by default starting from v0.17, the enableNodeFeatureApi has been removed.

https://kubernetes-sigs.github.io/node-feature-discovery/v0.17/reference/feature-gates.html

### DIFF
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -559,7 +559,6 @@ ccManager:
   resources: {}
 
 node-feature-discovery:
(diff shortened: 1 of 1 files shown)


tariq1890:
Verified that this field `enableNodeFeatureApi` is no longer used in the NFD helm charts of v0.17.x

Thank you @hhk7734 !

tariq1890 (approved):

PR #1474, opened 2025-06-05
Add driver rootfs and /host to vgpu-device-manager

mresvanis:
This PR mounts the NVIDIA driver rootfs, the host rootfs and the `gpu-clients` ConfigMap to the `vgpu-device-manager` container to enable MIG configuration before the vGPU configuration when MIG-backed vGPU types are included in the selected vGPU configuration.

`nvidia-mig-parted` requires NVML to configure MIG and the host rootfs to start/stop systemd services.

Related vGPU device manager PR: https://github.com/NVIDIA/vgpu-device-manager/pull/93

### DIFF
--- a/api/nvidia/v1/clusterpolicy_types.go
+++ b/api/nvidia/v1/clusterpolicy_types.go
@@ -26,6 +26,7 @@ import (
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
@@ -2093,4 +2094,30 @@ func (c *CCManagerSpec) IsEnabled() bool {
 		return false
 	}
 	return *c.Enabled
+}
+
+// +kubebuilder:object:generate=false
+type ConfigWithName interface {
+	GetName() string
+}
+
+// GetConfigMapName returns the config's name if it's non-empty and differs from defaultName,
+// otherwise returns defaultName. The boolean indicates whether a custom name was used.
+func GetConfigMapName[T ConfigWithName](config T, defaultName string) (string, bool) {
+	if name := config.GetName(); name != "" {
+		return name, name != defaultName
+	}
+	return defaultName, false
+}
+
+func (c *MIGGPUClientsConfigSpec) GetName() string {
+	return ptr.Deref(c, MIGGPUClientsConfigSpec{}).Name
+}
+
+func (c *MIGPartedConfigSpec) GetName() string {
+	return ptr.Deref(c, MIGPartedConfigSpec{}).Name
+}
+
+func (c *VGPUDevicesConfigSpec) GetName() string {
+	return ptr.Deref(c, VGPUDevicesConfigSpec{}).Name
 }
--- a/assets/state-vgpu-device-manager/0600_daemonset.yaml
+++ b/assets/state-vgpu-device-manager/0600_daemonset.yaml
@@ -54,6 +54,12 @@ spec:
           name: vgpu-config
         - mountPath: /sys
           name: host-sys
+        - mountPath: /host
+          name: host-root
+          mountPropagation: HostToContainer
+        - name: driver-install-dir
+          mountPath: /driver-root
+          mountPropagation: HostToContainer
       volumes:
       - name: vgpu-config
         configMap:
@@ -62,6 +68,13 @@ spec:
         hostPath:
           path: /sys
           type: Directory
+      - name: host-root
+        hostPath:
+          path: "/"
+      - name: driver-install-dir
+        hostPath:
+          path: "/run/nvidia/driver"
+          type: DirectoryOrCreate
       - name: run-nvidia-validations
         hostPath:
           path: /run/nvidia/validations
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -535,24 +535,24 @@ func createConfigMap(n ClusterPolicyController, configMapIdx int) (gpuv1.State, 
 
 	// avoid creating default 'mig-parted-config' ConfigMap if custom one is provided
 	if obj.Name == MigPartedDefaultConfigMapName {
-		if config.MIGManager.Config != nil && config.MIGManager.Config.Name != "" && config.MIGManager.Config.Name != MigPartedDefaultConfigMapName {
-			logger.Info(fmt.Sprintf("Not creating resource, custom ConfigMap provided: %s", config.MIGManager.Config.Name))
+		if name, isCustom := gpuv1.GetConfigMapName(config.MIGManager.Config, MigPartedDefaultConfigMapName); isCustom {
+			logger.Info("Not creating resource, custom ConfigMap provided", "Name", name)
 			return gpuv1.Ready, nil
 		}
 	}
 
 	// avoid creating default 'gpu-clients' ConfigMap if custom one is provided
 	if obj.Name == MigDefaultGPUClientsConfigMapName {
-		if config.MIGManager.GPUClientsConfig != nil && config.MIGManager.GPUClientsConfig.Name != "" {
-			logger.Info(fmt.Sprintf("Not creating resource, custom ConfigMap provided: %s", config.MIGManager.GPUClientsConfig.Name))
+		if name, isCustom := gpuv1.GetConfigMapName(config.MIGManager.GPUClientsConfig, MigDefaultGPUClientsConfigMapName); isCustom {
+			logger.Info("Not creating resource, custom ConfigMap provided", "Name", name)
 			return gpuv1.Ready, nil
 		}
 	}
 
 	// avoid creating default vGPU device manager ConfigMap if custom one provided
 	if obj.Name == VgpuDMDefaultConfigMapName {
-		if config.VGPUDeviceManager.Config != nil && config.VGPUDeviceManager.Config.Name != "" {
-			logger.Info(fmt.Sprintf("Not creating resource, custom ConfigMap provided: %s", config.VGPUDeviceManager.Config.Name))
+		if name, isCustom := gpuv1.GetConfigMapName(config.VGPUDeviceManager.Config, VgpuDMDefaultConfigMapName); isCustom {
+			logger.Info("Not creating resource, custom ConfigMap provided", "Name", name)
 			return gpuv1.Ready, nil
 		}
 	}
@@ -1808,10 +1808,7 @@ func TransformMIGManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 			continue
 		}
 
-		name := MigPartedDefaultConfigMapName
-		if config.MIGManager.Config != nil && config.MIGManager.Config.Name != "" && config.MIGManager.Config.Name != MigPartedDefaultConfigMapName {
-			name = config.MIGManager.Config.Name
-		}
+		name, _ := gpuv1.GetConfigMapName(config.MIGManager.Config, MigPartedDefaultConfigMapName)
 		obj.Spec.Template.Spec.Volumes[i].ConfigMap.Name = name
 		break
 	}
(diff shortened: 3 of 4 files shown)


assets/state-vgpu-device-manager/0600_daemonset.yaml
```diff
@@ -62,7 +76,17 @@ spec:
...
           type: Directory
+      - name: driver-install-dir
+        hostPath:
+          path: "/run/nvidia/driver"
+          type: DirectoryOrCreate
       - name: run-nvidia-validations
         hostPath:
           path: /run/nvidia/validations
           type: DirectoryOrCreate
+      - name: mig-parted-config
```
tariq1890: Instead of adding these two configMap references here, we could just update the unmarshaled ConfigMap object with these configMap references when we detect that the clusterpolicy custom resource includes additional configMaps

mresvanis: Thank you for pointing this out, it slipped my TODOs when drafting this PR. What do you think about implementing the same logic found in the respective [MIG Manager transformation](https://github.com/NVIDIA/gpu-operator/blob/dfbb6c5c4464723f6f1e4c736d3e9c3e2660deed/controllers/object_controls.go#L1807-L1833)? 

I also tried to reduce the duplicate checks that came up, but let me know if you think that's less readable and/or not maintainable and I'll go back to simple checks.

assets/state-vgpu-device-manager/0600_daemonset.yaml
```diff
@@ -62,7 +76,17 @@ spec:
         hostPath:
           path: /sys
           type: Directory
+      - name: driver-install-dir
```
cdesiniotis, line 74: Note: the `reconfigure_mig.sh` script assumes that the host root fs is mounted at `/host` in the container (it chroot's into the host rootfs to stop / start systemd services when drivers are installed directly on the host). So we need to add a volume for the host root.

assets/state-vgpu-device-manager/0600_daemonset.yaml
```diff
@@ -45,15 +45,29 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: LD_PRELOAD
+          value: "/driver-root/usr/lib64/libnvidia-ml.so.1"
```
cdesiniotis: Ideally we wouldn't set this in the spec. Instead, we would discover the path to `libnvidia-ml.so.1` in the vgpu-device-manager code and set the `LD_PRELOAD` envvar when invoking the mig-parted binaries.  We have precedent for this in other components. For example, here is our code in the `gpu-operator-validator` that validates a driver container installation: https://github.com/NVIDIA/gpu-operator/blob/6324d2aca562edf46d93cbf9d2a0837ab5c12e59/cmd/nvidia-validator/main.go#L731-L749

assets/state-vgpu-device-manager/0600_daemonset.yaml
mresvanis: Great catch and thank you for the explanation! I added the `/host` mount.

assets/state-vgpu-device-manager/0600_daemonset.yaml
mresvanis: This makes sense, thank you for pointing out the use of such code in other components. I refactored the vgpu-device-manager [PR to search for the library](https://github.com/NVIDIA/vgpu-device-manager/pull/93/files#diff-88b1ccb8fb9b4bc000927197c45c3e32d10905fddef38dce4496d5b7d4245a71R604-R609) and setup `LD_PRELOAD` for the mig-parted commands.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -2094,3 +2099,36 @@ func (c *CCManagerSpec) IsEnabled() bool {
...
+// otherwise returns defaultName. The boolean indicates whether a custom name was used.
+func GetConfigMapName[T ConfigWithName](config T, defaultName string) (string, bool) {
+	if config.GetName() != "" && config.GetName() != defaultName {
+		return config.GetName(), true
+	}
+	return defaultName, false
+}
+
+// safeGetName safely gets the name from a pointer, handling nil cases.
+func safeGetName[T any](ptr *T, getName func(T) string) string {
```
tariq1890: Can we import this method from `k8s.io/utils` instead? 

https://github.com/kubernetes/utils/tree/master/pointer

controllers/object_controls.go
```diff
@@ -532,24 +532,28 @@ func createConfigMap(n ClusterPolicyController, configMapIdx int) (gpuv1.State,
 
 	// avoid creating default 'mig-parted-config' ConfigMap if custom one is provided
 	if obj.Name == MigPartedDefaultConfigMapName {
-		if config.MIGManager.Config != nil && config.MIGManager.Config.Name != "" && config.MIGManager.Config.Name != MigPartedDefaultConfigMapName {
-			logger.Info(fmt.Sprintf("Not creating resource, custom ConfigMap provided: %s", config.MIGManager.Config.Name))
+		if name, isCustom := gpuv1.GetConfigMapName(config.MIGManager.Config, MigPartedDefaultConfigMapName); isCustom {
+			logger.Info(fmt.Sprintf("Not creating resource, custom ConfigMap provided: %s", name))
```
tariq1890: We don't need `fmt.Sprintf` as `logger.Info` already supports formatted messages

See the method signature below
```
func (l Logger) Info(msg string, keysAndValues ...any) {
	if l.sink == nil {
		return
	}
	if l.sink.Enabled(l.level) { // see comment in Enabled
		if withHelper, ok := l.sink.(CallStackHelperLogSink); ok {
			withHelper.GetCallStackHelper()()
		}
		l.sink.Info(l.level, msg, keysAndValues...)
	}
}
```

tariq1890: This is an example of a logger.Info() call

```
		w.log.Info("Operands are disabled for node", "NodeName", w.node, "Label", commonOperandsLabelKey, "Value", "false")
```

controllers/object_controls.go
```diff
@@ -2060,10 +2058,18 @@ func TransformVGPUDeviceManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
-		if config.VGPUDeviceManager.Config != nil && config.VGPUDeviceManager.Config.Name != "" {
-			name = config.VGPUDeviceManager.Config.Name
+		name, _ := gpuv1.GetConfigMapName(config.VGPUDeviceManager.Config, VgpuDMDefaultConfigMapName)
+		obj.Spec.Template.Spec.Volumes[i].ConfigMap.Name = name
+		break
+	}
+
+	// set ConfigMap name for "gpu-clients" Volume
+	for i, vol := range obj.Spec.Template.Spec.Volumes {
+		if !strings.Contains(vol.Name, "gpu-clients") {
```
tariq1890: I'd prefer an exact string match instead of a `Contains`

mresvanis: I can definitely update this, I just wasn't sure about why `Contains` is used in other such cases, e.g. [here](https://github.com/NVIDIA/gpu-operator/blob/b28606d7d1acc30c2c9df4a027d3ff85ee211946/controllers/object_controls.go#L2062) and thought I'd adhere to what is already used.

controllers/object_controls.go
mresvanis: Thanks for pointing that out and the detailed explanation! I blindly followed the existing log message. I have updated these changes.

api/nvidia/v1/clusterpolicy_types.go
mresvanis: Nice! I saw that `k8s/utils` was an indirect import and wasn't sure if you also think it makes sense to use it here (and thus make it a direct import).

Did you have something like this in mind?

```go
import (
	"k8s.io/utils/ptr"
)
...
type ConfigWithName interface {
	GetName() string
}

// GetConfigMapName returns the config's name if it's non-empty and differs from defaultName,
// otherwise returns defaultName. The boolean indicates whether a custom name was used.
func GetConfigMapName[T ConfigWithName](config T, defaultName string) (string, bool) {
	if name := config.GetName(); name != "" {
		return name, name != defaultName
	}
	return defaultName, false
}

func (c *MIGGPUClientsConfigSpec) GetName() string {
	return ptr.Deref(c, MIGGPUClientsConfigSpec{}).Name
}

func (c *MIGPartedConfigSpec) GetName() string {
	return ptr.Deref(c, MIGPartedConfigSpec{}).Name
}

func (c *VGPUDevicesConfigSpec) GetName() string {
	return ptr.Deref(c, VGPUDevicesConfigSpec{}).Name
}
```

Happy to go back to the simplified solution if you think this abstraction isn't worth it here.

controllers/object_controls.go
```diff
@@ -2060,10 +2058,18 @@ func TransformVGPUDeviceManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
 
-		name := VgpuDMDefaultConfigMapName
-		if config.VGPUDeviceManager.Config != nil && config.VGPUDeviceManager.Config.Name != "" {
-			name = config.VGPUDeviceManager.Config.Name
+		name, _ := gpuv1.GetConfigMapName(config.VGPUDeviceManager.Config, VgpuDMDefaultConfigMapName)
+		obj.Spec.Template.Spec.Volumes[i].ConfigMap.Name = name
+		break
+	}
+
+	// set ConfigMap name for "gpu-clients" Volume
```
cdesiniotis: A general question / comment - what happens if the MIG Manager component is explicitly disabled in ClusterPolicy? In this case, the `default-gpu-clients` ConfigMap would not have been created by the GPU Operator, and thus, the vGPU Device Manager would fail to run (due to the missing volume).

We don't have any precedent for this (so I am not sure if there are any adverse effects), but what about creating a symlink to the `default-gpu-clients` ConfigMap contents?
```
assets/state-vgpu-device-manager/0510_configmap.yaml --> assets/state-mig-manager/0410_configmap.yaml
```
This way, when reconciling the vGPU Device Manager component, the GPU Operator will always attempt to create the `default-gpu-clients` ConfigMap when the user has not specified a custom ConfigMap -- regardless if the MIG Manager is enabled or not.

controllers/object_controls.go
cdesiniotis: No good reason AFAIK. Someone probably made the decision to use `Contains` a while ago and it just stuck throughout. We can update the conditional here to check for an exact string match.

controllers/object_controls.go
cdesiniotis: Thinking about this more, a symlink will not work. If the MIG Manager OR vGPU Device Manager is disabled in ClusterPolicy, the ConfigMap will get deleted here: https://github.com/NVIDIA/gpu-operator/blob/1d56b20fa5367ffaa9b9cc83349585561dde17a2/controllers/object_controls.go#L527-L534

We might need to add a copy of this ConfigMap in `assets/state-vgpu-device-manager` (same contents, different name) unless one of us thinks of a more clever solution.

cdesiniotis: One thing to note, the `gpu-clients` configuration file is only relevant when drivers are pre-installed on the host (not installed via the driver container) -- the `gpu-clients` configuration file contains a list of systemd services that use the GPU and need to be restarted across a MIG re-configuration. 

We technically don't support pre-installed vGPU manager (aka the host driver) today. This becomes apparent when reading this code snippet / TODO item: https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-vgpu-device-manager/0600_daemonset.yaml#L25-L28

Because we don't support pre-installed vGPU Manager, I would be open to removing the gpu-clients configuration handling from this PR entirely. It may be more appropriate to add the gpu-clients configuration handling in the same PR that adds preinstalled vGPU Manager support (whenever this becomes a requirement and prioritized).

assets/state-vgpu-device-manager/0600_daemonset.yaml
cdesiniotis: Is this comment still relevant?

mresvanis: We have removed both the `mig-parted-config` and the `gpu-clients` ConfigMaps, as neither is currently relevant to these changes. 

`mig-parted-config` is not needed since the vgpu-device-manager handles the conversion of of the vGPU config to the respective MIG config. 

`gpu-clients` is not needed since there is currently [no support for pre-installed vGPU host drivers](https://github.com/NVIDIA/gpu-operator/pull/1474#discussion_r2191161929), which is the purpose of this ConfigMap in the first place. 

Thank you both for the detailed explanations and suggestions!

controllers/object_controls.go
mresvanis: Thank you very much for the detailed explanation! I agree with you that we should not include the `gpu-clients` ConfigMap as it's not currently relevant. I have updated the PR accordingly.

api/nvidia/v1/clusterpolicy_types.go
tariq1890: Yes, this looks good to me. Thanks!

cdesiniotis (approved):

cdesiniotis:
@tariq1890 can you take another look at this PR?

tariq1890 (approved):

tariq1890:
Can you rebase this PR @mresvanis ?

PR #1480, opened 2025-06-10
Ensured user-supplied envvars take Precedence

JunAr7112:


controllers/object_controls.go
```diff
@@ -3255,6 +3250,12 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 			setContainerEnv(driverContainer, UseHostMOFEDEnvName, "true")
 		}
 	}
+	// set/append environment variables for driver container
+	if len(config.Driver.Env) > 0 {
+		for _, env := range config.Driver.Env {
+			setContainerEnv(driverContainer, env.Name, env.Value)
+		}
+	}
```
cdesiniotis: Note, on line 3401 (later in this function) an environment variable is being set.

controllers/object_controls.go
```diff
@@ -1890,6 +1883,13 @@ func TransformKataManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec
 	// set env used by readinessProbe to determine path to kata-manager pid file.
 	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), "KATA_ARTIFACTS_DIR", artifactsDir)
 
+	// set/append environment variables for mig-manager container
+	if len(config.KataManager.Env) > 0 {
+		for _, env := range config.KataManager.Env {
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
+		}
+	}
+
```
cdesiniotis: Note, on line 1903 (later in this function) we call the `transformForRuntime()` method which sets envvars for the main container -- so there are still certain envvars that a user could never customize for this container.

This brings up a general question -- what is the "optimal" way to *always* ensure that user-supplied envvars take precedence over the variables that we set in these transform functions? How do we future-proof this so we don't accidentally add code to one of these transform functions that invalidates the order of precedence?

Without putting much thought into this, one option might be to add a defer statement in each `Transform*` function where user-supplied envvars are processed. This would ensure user-supplied envvars are always added last after all other transformations are applied. For example:
```
func TransformToolkit(. . .) error {
  defer func() {
	if len(config.Toolkit.Env) > 0 {
		for _, env := range config.Toolkit.Env {
			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
		}
	}
  }
  . . .
}
```
@JunAr7112 @tariq1890 thoughts on this?

JunAr7112: I think a defer function might work well here, that way it will always execute after other environment variables are set in the transform functions.

tariq1890: I would advise against adding any business logic to `defer` blocks. That's the not the purpose of defer statements.

controllers/object_controls.go
JunAr7112: Fixed this

controllers/object_controls.go
```diff
@@ -3405,6 +3400,13 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 	if err != nil {
 		return err
 	}
+
+	if len(config.Driver.Env) > 0 {
+		for _, env := range config.Driver.Env {
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
+		}
+	}
```
cdesiniotis: I know I requested to move this logic to the end of this function. However, `TransformDriver` may be an exception to the rule. There are a few places where we potentially return early from this function. On line 3323 we return early when precompiled is enabled. On line 3391 we return early if not running on OpenShift. In either case, we return early and never apply user-supplied envvars. So let's move this code block to before line 3323.

JunAr7112: Alright I'll move it back

controllers/object_controls.go
```diff
@@ -3405,6 +3406,7 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 	if err != nil {
 		return err
 	}
+
```
cdesiniotis: nit: let's remove this unnecessary change.

tariq1890 (dismissed):

tariq1890:
@JunAr7112  Please rebase this PR

JunAr7112:
> @JunAr7112 Please rebase this PR

I rebased to main. Will update the test cases right now

tariq1890:
@JunAr7112 Can you address the failing unit tests?

controllers/transforms_test.go
```diff
@@ -563,6 +565,301 @@ func TestApplyCommonDaemonsetMetadata(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			err := TransformDCGMExporter(tc.ds.DaemonSet, tc.cpSpec, ClusterPolicyController{runtime: gpuv1.Containerd, logger: ctrl.Log.WithName("test")})
+			if err != nil {
+				t.Fatalf("TransformDCGMExporter failed: %v", err)
+			}
+
+			if !reflect.DeepEqual(tc.ds.DaemonSet, tc.expectedDs.DaemonSet) {
```
tariq1890: Let's not use `reflect.DeepEqual` please. I prefer `require.EqualValues`

JunAr7112: Alright

tariq1890 (approved):

cdesiniotis:
+1.0% code coverage :sunglasses:

PR #1483, opened 2025-06-14
Consolidate operator and validator into one image

cdesiniotis:


docker/Dockerfile
```diff
@@ -53,13 +54,23 @@ COPY *.mk .
...
 
 FROM nvcr.io/nvidia/cuda:12.9.0-base-ubi9
 
 # Remove CUDA libs(compat etc) in favor of libs installed by the NVIDIA driver
 RUN dnf remove -y cuda-*
 
+RUN dnf install -y \
+        kmod \
+        pciutils && \
+    rm -rf /var/cache/yum/*
```
cdesiniotis, line 69: We should be able to remove this once we move to the distroless base image in https://github.com/NVIDIA/gpu-operator/pull/1417.

The distroless `dev` tags comes with busybox, which include the utilities needed from these packages, namely `lsmod` and `lspci`. 

For some historical context, `kmod` was added to the validator image in https://github.com/NVIDIA/gpu-operator/commit/24c7bd0d7b32d81dd6507877f49e45b4c04b5fdb specifically to run `lsmod`. `pciutils` was added in https://github.com/NVIDIA/gpu-operator/commit/c657465f44aa332da16da5ed05bd7f55722abdf8 specifically to run `lspci` [here](https://github.com/NVIDIA/gpu-operator/blob/dfbb6c5c4464723f6f1e4c736d3e9c3e2660deed/validator/metrics.go#L248).

deployments/gpu-operator/templates/operator.yaml
```diff
@@ -58,6 +58,9 @@ spec:
               fieldPath: metadata.namespace
         - name: "DRIVER_MANAGER_IMAGE"
           value: "{{ include "driver-manager.fullimage" . }}"
+        securityContext:
+          runAsUser: 65532
+          runAsGroup: 65532
```
elezar: Was this required in the context of this PR?

docker/Dockerfile
```diff
@@ -53,13 +54,23 @@ COPY *.mk .
 # Build
 ARG VERSION="unknown"
 ARG GIT_COMMIT="unknown"
-RUN make gpu-operator
+RUN make cmd-gpu-operator
+RUN make cmd-validator
```
elezar: In the device plugin we do the following:
```
RUN mkdir /artifacts
ARG VERSION="N/A"
ARG GIT_COMMIT="unknown"
RUN make PREFIX=/artifacts cmds
```
(Not sure if the prefix is supported in the GPU Operator makefile).

My concrete suggestion here would be:
```suggestion
RUN make cmds
```

docker/Dockerfile
```diff
@@ -53,13 +54,23 @@ COPY *.mk .
...
+
+FROM ${CUDA_SAMPLE_IMAGE} AS sample-builder
+RUN mkdir /artifacts
+RUN cp /cuda-samples/vectorAdd /artifacts/vectorAdd
```
elezar: Not critical, but is there a reason that we don't just copy `/cuda-samples/vectorAdd` in the final image?

docker/Dockerfile
```diff
@@ -76,10 +87,19 @@ LABEL vsc-ref=${GIT_COMMIT}
 
 WORKDIR /
 COPY --from=builder /workspace/gpu-operator /usr/bin/
+COPY --from=builder /workspace/validator /usr/bin/nvidia-validator
```
elezar: nit: let's create a follow-up to also rename the actual executable.

docker/Dockerfile
```diff
@@ -76,10 +87,19 @@ LABEL vsc-ref=${GIT_COMMIT}
...
 
+# gpu-operator manifests
 RUN mkdir -p /opt/gpu-operator/manifests
 COPY assets /opt/gpu-operator/
 COPY manifests /opt/gpu-operator/manifests
+
+# validator manifests
+RUN mkdir -p /opt/validator/manifests
+COPY cmd/validator/plugin-workload-validation.yaml /opt/validator/manifests
+COPY cmd/validator/cuda-workload-validation.yaml /opt/validator/manifests
```
elezar: nit: These manifests should not be in `cmd/validator` in the source.

docker/Dockerfile
```diff
@@ -101,7 +121,8 @@ RUN if [ -n "${CVE_UPDATES}" ]; then \
         rm -rf /var/cache/yum/*; \
     fi
 
-RUN useradd gpu-operator
-USER gpu-operator
+# The validator needs to run as uid 0 but the operator does not.
+# When deploying the operator, run as a non-root user.
+USER 0:0
```
elezar: This is a bit strange since we're setting this here by the default command / entrypoint is `gpu-operator` and not `nvidia-validator`.

elezar (commented):
Thanks for this cleanup. My primary question would be whether we also want to go ahead and rename the validator in the source directly. Not a blocker though.

docker/Dockerfile
cdesiniotis: I guess I could update the entrypoint to `nvidia-validator` for consistency. The reason I chose to default to `USER 0:0` is because it reduces the amount of changes required in our manifests -- we have many instances of the "validator" container, but only one "operator" container.

deployments/gpu-operator/templates/operator.yaml
cdesiniotis: Before this PR, the gpu-operator container was running as a non-root user by default. With this PR, I am setting `USER 0:0` in the Dockerfile, so yes, these changes are needed.

elezar: As mentioned where the Dockerfile change was made, does it make sense to default to `65532:65532` there and set the user explicitly for the validator?

docker/Dockerfile
elezar: Can these not be applied by the operator itself when construcing the validator?

cdesiniotis: I don't see why not. That would allow us to default to `65532:65532` in the Dockerfile (which is better from a security perspective) and only overwrite this once in the code that renders the validator containers. Thanks for the suggestion. Let me try this in my next iteration.

deployments/gpu-operator/templates/operator.yaml
cdesiniotis: Updated to `65532:65532`.

docker/Dockerfile
cdesiniotis: Updated to 
```
RUN make cmds
```

docker/Dockerfile
cdesiniotis: No particular reason. I naively copied this from the validator's Dockerfile. I have updated this now - we simply copy `/cuda-samples/vectorAdd` in the final image.

docker/Dockerfile
cdesiniotis: Updated to `65532:65532`. Also updated the code so that `runAsUser: 0` gets set in the securityContext for all validation containers we render.

controllers/object_controls.go
```diff
@@ -2079,6 +2079,19 @@ func TransformVGPUDeviceManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+// container so that it runs as uid 0. Some of the validations run commands
+// that require root privileges (e.g. chroot). In addition, all validations
+// create / delete status files in the '/run/nvidia/validations' host path
+// volume. This directory is initially created by the kubelet and thus has
+// the same group and ownership as the kubelet.
+func transformValidatorSecurityContext(ctr *corev1.Container) {
+	if ctr.SecurityContext == nil {
+		ctr.SecurityContext = &corev1.SecurityContext{}
+	}
+	ctr.SecurityContext.RunAsUser = new(int64)
```
tariq1890: Let's create a constant int64 for the root UID and use that everywhere
```suggestion
	ctr.SecurityContext.RunAsUser = new(int64)
```

controllers/object_controls.go
```diff
@@ -192,6 +192,9 @@ const (
 	Readiness ContainerProbe = "readiness"
 )
 
+// rootUID represents user 0
+var rootUID = new(int64)
```
tariq1890: Can you make this a constant and move it to the top of the file?
```suggestion
	const rootUID int64 = 0
```

controllers/transforms_test.go
```diff
@@ -596,11 +596,17 @@ func TestTransformValidationInitContainer(t *testing.T) {
 				Image:           "nvcr.io/nvidia/cloud-native/gpu-operator-validator:v1.0.0",
 				ImagePullPolicy: corev1.PullIfNotPresent,
 				Env:             []corev1.EnvVar{{Name: "foo", Value: "bar"}},
+				SecurityContext: &corev1.SecurityContext{
+					RunAsUser: new(int64),
```
tariq1890: let's reference the constant here as well

docker/Dockerfile
tariq1890: Agreed. Can we move the source assets to `validator/manifests` ?

docker/Dockerfile
tariq1890: Let's name it either gpu-operator-validator or operator-validator
```suggestion
COPY --from=builder /workspace/validator /usr/bin/operator-validator
```

tariq1890 (commented):
Looks great , just a few comments.

Thanks for working on this @cdesiniotis!

controllers/object_controls.go
cdesiniotis: This won't work since you cannot get the address of a constant in Go.

tariq1890: ahh got it. Can we create a util method to convert the int64 to a pointer ? We already do that for booleans. See [here](https://github.com/NVIDIA/gpu-operator/blob/7068fec931e78af2994f82a6ecfea55cbf737b48/internal/utils/utils.go#L61)

docker/Dockerfile
cdesiniotis: I have updated the binary name to `nvidia-validator` following a discussion with @tariq1890

docker/Dockerfile
cdesiniotis: Moved the yaml files to `validator/manifests`

controllers/object_controls.go
cdesiniotis: Sure, updated.

controllers/transforms_test.go
cdesiniotis: Done

tariq1890 (approved):

PR #1494, opened 2025-06-18
auto nvcr image version update

shivakunv:


shivakunv:
closing . as we consider renovate tools for this purpose

PR #1496, opened 2025-06-18
Use POSIX sh instead of bash in entrypoint scripts and container probes

cdesiniotis:
This change is needed now that we are moving all of our operands to distroless. The distroless `-dev` tags contain `sh` and not `bash`.

### DIFF
--- a/assets/state-mig-manager/0420_configmap.yaml
+++ b/assets/state-mig-manager/0420_configmap.yaml
@@ -7,9 +7,9 @@ metadata:
     app: nvidia-mig-manager
 data:
   entrypoint.sh: |-
-    #!/bin/bash
+    #!/bin/sh
 
-    until [[ -f /run/nvidia/validations/driver-ready ]]
+    until [ -f /run/nvidia/validations/driver-ready ]
     do
       echo "waiting for the driver validations to be ready..."
       sleep 5
--- a/assets/state-vfio-manager/0400_configmap.yaml
+++ b/assets/state-vfio-manager/0400_configmap.yaml
@@ -7,7 +7,7 @@ metadata:
     app: nvidia-vfio-manager
 data:
   vfio-manage.sh: |-
-    #!/bin/bash
+    #!/bin/sh
 
     set -eu
 
@@ -25,9 +25,7 @@ data:
     }
 
     unbind_from_driver() {
-        local gpu=$1
-        local existing_driver_name
-        local existing_driver
+        gpu=$1
 
         [ -e "/sys/bus/pci/devices/$gpu/driver" ] || return 0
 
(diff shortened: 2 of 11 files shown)


assets/state-vfio-manager/0400_configmap.yaml
cdesiniotis, line 1: I have tested the changes to this script:

```
$ sudo ./vfio-manage.sh bind --all
binding device 0000:3b:00.0
binding device 0000:86:00.0

$ sudo ./vfio-manage.sh unbind --all
unbinding device 0000:3b:00.0
unbinding device 0000:3b:00.0 from driver vfio-pci
unbinding device 0000:86:00.0
unbinding device 0000:86:00.0 from driver vfio-pci

$ sudo ./vfio-manage.sh bind -d 0000:3b:00.0
binding device 0000:3b:00.0

$ sudo ./vfio-manage.sh unbind -d 0000:3b:00.0
unbinding device 0000:3b:00.0
unbinding device 0000:3b:00.0 from driver vfio-pci
```

cdesiniotis: @shivakunv the changes I have made here are nearly identical to the ones you proposed in https://github.com/NVIDIA/k8s-driver-manager/pull/74. I noted two differences and left review comments on your PR accordingly.

assets/state-mig-manager/0420_configmap.yaml
```diff
@@ -7,9 +7,9 @@ metadata:
     app: nvidia-mig-manager
 data:
   entrypoint.sh: |-
-    #!/bin/bash
+    #!/bin/sh
```
elezar, line 10: We're not yet switching to distroless for the mig-manager, or has work started to also port its bash script to shell?

assets/state-vfio-manager/0400_configmap.yaml
shivakunv: Yes , These are only warnings. I tried to fix them for a clean shellcheck output.
reverted

assets/state-mig-manager/0420_configmap.yaml
cdesiniotis: We haven't switched mig-manager to distroless yet, but we were planning to start that effort soon -- either by porting the bash script to shell or rewriting it in Go.

Regardless, updating the entrypoint script to use `sh` now (before the switch to distroless) shouldn't be an issue, right?

cdesiniotis:
> Thank you! This can be quite a challenging change. Good luck to us!
> 
> Did you use any tooling, or did you try to reason about this manually? Sometimes (often) one finds the breakage later. Which may be fine, if we allow ourselves to test this properly.
> 
> I sometimes forget that even `echo` behaves differently in bash compared to POSIX (where it has less options).

I used `shellcheck` to identfy bashisms that needed to be converted to POSIX sh. Otherwise, I am reasoning about this manually :) We will definitely need to test this extensively to catch any regressions before release.

tariq1890 (approved):

PR #1509, opened 2025-06-30
[Openshift] add 25.3.1 OLM bundle and remove old ones

tariq1890:


cdesiniotis:
Question -- What should our retention policy be?

tariq1890:
Great question, What if we align the retention policy with the versions maintained in our official docs page?

https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/getting-started.html

cdesiniotis:
Do we even need to store the released bundles in this repo? They are already stored at https://github.com/redhat-openshift-ecosystem/certified-operators. Maybe a compromise would be to commit released bundles to release branches? E.g. the 25.3.1 and 25.3.0 bundles would be committed to the `release-25.3` branch (but not main). This way, we don't clutter main.

cdesiniotis:
I don't have a strong opinion though, we could follow the same retention policy as we follow on our docs page. Everytime we add a new released OLM bundle, we also remove one.

tariq1890:
Closing this in favour of #1526

PR #1516, opened 2025-07-02
[dcgm-exporter] Add configure externalTrafficPolicy and nodePort for dcgmExporter service

Alja9:
Because the dcgmExporter service supports configuring types, in my case I use a different cluster as a monitoring center so I need NodePort type and I can't use IP pod or original source IP for scraping, but `externalTrafficPolicy: Cluster` would really help in that case.

But this will also help users if they need `externalTrafficPolicy: Local` such as cases where only specific users need node local routing or to preserve the original IP. Reference : https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip


Alja9:
I have raised PR to dcgm exporter too : https://github.com/NVIDIA/dcgm-exporter/pull/518

PR #1529, opened 2025-07-09
Per-node container runtime support

### LINKED ISSUE #1528: Support containerd and crio nodes in the same k8s cluster
Hi,

It seems that currently gpu-operator doesn't support clusters with a mix of containerd and crio nodes. There is one global check here: https://github.com/NVIDIA/gpu-operator/blob/1c71f9883e2b1d85d4b8afdcef51882c71f2f9fc/controllers/state_manager.go#L714

This makes it hard to move containerd clusters to crio because all nodes have to be rebuilt at once. Would you accept a PR to fix this?

This change would be much easier if https://github.com/NVIDIA/gpu-operator/commit/efaef36f9425fce59415b5eb8889f87969d76d04 could first be reverted, as then most daemonsets would be identical for crio and containerd. The commit message for that commit is "Disable RuntimeClass with CRI-O until toolkit changes are ready to update /etc/crio/crio.conf", implying it wouldn't be needed for ever.

Thanks,

Rob

suprjinx:
Key Features Implemented:

  1. Per-Runtime DaemonSets: Creates separate DaemonSets for driver, toolkit, and device-plugin components based on detected node runtimes
  2. Automatic Runtime Detection: Detects container runtime from node status and handles OpenShift special cases
  3. Node Runtime Labeling: Automatically labels nodes with nvidia.com/gpu.runtime.<runtime> labels
  4. Dynamic Runtime Map: Maintains real-time mapping of nodes to their container runtimes
  5. Mixed Runtime Support: Supports clusters with nodes using different container runtimes (containerd, cri-o, docker)
  6. Full Functionality Preservation: All original driver, toolkit, and device plugin functionality maintained through enhanced per-runtime transformations

  Architecture Changes:

  - From: Single cluster-wide DaemonSets with runtime class selection
  - To: Multiple runtime-specific DaemonSets with node selectors
  - Benefits: Eliminates runtime compatibility issues in mixed environments and ensures proper GPU software deployment across all node types

  Test Coverage:

  - 511 lines of new comprehensive tests covering all runtime scenarios
  - Tests for node lifecycle management (add/remove)
  - Tests for runtime detection in various cluster configurations
  - Tests for runtime map management and node labeling

  This implementation ensures that GPU workloads can run correctly on clusters with mixed container runtimes while maintaining all existing functionality and performance characteristics.

Fixes #1528

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -4672,7 +4672,9 @@ func RuntimeClasses(n ClusterPolicyController) (gpuv1.State, error) {
 	}
 
 	createRuntimeClassFunc := transformRuntimeClass
-	if semver.Compare(n.k8sVersion, nodev1MinimumAPIVersion) <= 0 {
+	// For Kubernetes versions <= v1.20.0, use legacy v1beta1 API
+	// For newer versions, use v1 API (v1beta1 was deprecated in 1.22 and removed in 1.25+)
+	if n.k8sVersion != "" && semver.IsValid(n.k8sVersion) && semver.Compare(n.k8sVersion, nodev1MinimumAPIVersion) <= 0 {
 		createRuntimeClassFunc = transformRuntimeClassLegacy
 	}
 


controllers/state_manager.go
```diff
@@ -868,11 +931,22 @@ func (n *ClusterPolicyController) init(ctx context.Context, reconciler *ClusterP
 	}
 
 	// detect the container runtime on worker nodes
-	err = n.getRuntime()
+	n.nodeRuntimeMap, err = n.getNodeRuntimeMap()
 	if err != nil {
 		return err
 	}
-	n.logger.Info(fmt.Sprintf("Using container runtime: %s", n.runtime.String()))
+	n.logger.Info(fmt.Sprintf("Detected container runtimes per node: %v", n.nodeRuntimeMap))
```
suprjinx: we may not want to log this map, could be large!

rahulait:
This PR should not be auto-closed. We need to evaluate if we need this and then decide.

PR #1535, opened 2025-07-10
bump k8s-operator-libs and openshift/[client-go,api]

tariq1890:


tariq1890:
This change has been validated on an OpenShift 4.14 cluster

PR #1541, opened 2025-07-17
Integrate NVIDIA DRA Driver for GPUs as an operand

cdesiniotis:
# Tests

## K8s Cluster that does not support DRA

Cluster setup:
```
$ kubectl get nodes
NAME           STATUS   ROLES           AGE   VERSION
cnt-server-2   Ready    control-plane   61d   v1.31.0
```

Install GPU Operator with default options. Then, try enabling the DRA driver and verify the operator emits an error message indicating this is an invalid configuration:
```
$ kubectl patch clusterpolicies.nvidia.com/cluster-policy     --type='json'     -p='[{"op": "replace", "path": "/spec/draDriver/computeDomains/enabled", "value":true}]'
clusterpolicy.nvidia.com/cluster-policy patched

$ kubectl get clusterpolicies.nvidia.com/cluster-policy   
NAME             STATUS     AGE
cluster-policy   notReady   2025-07-18T22:24:47Z


$ kubectl get clusterpolicy -ojsonpath='{.items[0].status}{"\n"}' | jq .
{
  "conditions": [
    {
      "lastTransitionTime": "2025-07-18T22:24:47Z",
      "message": "",
      "reason": "Error",
      "status": "False",
      "type": "Ready"
    },
    {
      "lastTransitionTime": "2025-07-18T22:24:47Z",
      "message": "failed to initialize ClusterPolicy controller: ClusterPolicy validation failed: failed to validate DRA: the NVIDIA DRA driver is enabled in ClusterPolicy but Dynamic Resource Allocation is not enabled in the kubernetes cluster",
      "reason": "ReconcileFailed",
      "status": "True",
      "type": "Error"
    }
  ],
  "namespace": "gpu-operator",
  "state": "notReady"
}
```

If a user tries to enable the DRA driver during a helm install or upgrade, we emit an error message:
```
$ helm upgrade gpu-operator gpu-operator-v25.8.0-devel.tgz -n gpu-operator --set draDriver.computeDomains.enabled=true
Error: UPGRADE FAILED: execution error at (gpu-operator/templates/validation.yaml:13:4): 
Cannot enable NVIDIA DRA Driver for GPUs on a Kubernetes cluster that does not support DRA
```

## K8s Cluster that supports DRA

Cluster setup:
```
$ kubectl get nodes        
NAME                   STATUS   ROLES                  AGE   VERSION
gb-nvl-043-compute03   Ready    control-plane,worker   8d    v1.32.4
gb-nvl-043-compute09   Ready    worker                 8d    v1.32.4
```
Install GPU Operator with default options. The DRA driver is disabled by default.
```
$ helm install gpu-operator gpu-operator-v25.8.0-devel.tgz  -n gpu-operator \
  --set operator.repository=ghcr.io/nvidia --set operator.version=1b983f26 \
  --set validator.repository=ghcr.io/nvidia --set validator.version=1b983f26
  
$ kubectl get clusterpolicies.nvidia.com/cluster-policy   
NAME             STATUS   AGE
cluster-policy   ready    2025-07-19T18:54:37Z

$ kubectl get pods -n gpu-operator --field-selector spec.nodeName=gb-nvl-043-compute09
NAME                                               READY   STATUS      RESTARTS   AGE
gpu-feature-discovery-2fwbj                        1/1     Running     0          2m12s
gpu-operator-node-feature-discovery-worker-4gmr4   1/1     Running     0          2m45s
nvidia-container-toolkit-daemonset-jhq5w           1/1     Running     0          2m12s
nvidia-cuda-validator-pdm7r                        0/1     Completed   0          33s
nvidia-dcgm-exporter-lxxxp                         1/1     Running     0          2m12s
nvidia-device-plugin-daemonset-8l8gh               1/1     Running     0          2m12s
nvidia-driver-daemonset-tcv7q                      1/1     Running     0          2m20s
nvidia-mig-manager-9hjr9                           1/1     Running     0          2m12s
nvidia-operator-validator-vgtjv                    1/1     Running     0          2m12s
```
Deploy the DRA Driver, only enable ComputeDomains:
```
$ kubectl patch clusterpolicies.nvidia.com/cluster-policy --type='json' -p='[{"op": "replace", "path": "/spec/draDriver/computeDomains/enabled", "value":true}]' 
clusterpolicy.nvidia.com/cluster-policy patched

$ kubectl get pods -n gpu-operator -lapp=nvidia-dra-driver-controller 
NAME                                           READY   STATUS    RESTARTS   AGE
nvidia-dra-driver-controller-b454444fc-pg4bw   1/1     Running   0          85s

$ kubectl get pods -n gpu-operator -lapp=nvidia-dra-driver-kubelet-plugin
NAME                                     READY   STATUS    RESTARTS   AGE
nvidia-dra-driver-kubelet-plugin-24755   1/1     Running   0          91s
nvidia-dra-driver-kubelet-plugin-x7m9p   1/1     Running   0          91s

$ kubectl get resourceslice
NAME                                                   NODE                   DRIVER                      POOL                   AGE
gb-nvl-043-compute03-compute-domain.nvidia.com-22tm2   gb-nvl-043-compute03   compute-domain.nvidia.com   gb-nvl-043-compute03   111s
gb-nvl-043-compute09-compute-domain.nvidia.com-87gdh   gb-nvl-043-compute09   compute-domain.nvidia.com   gb-nvl-043-compute09   111s

$ (echo -e "NODE\tLABEL\tCLIQUE"; kubectl get nodes -o json | \
    jq -r '.items[] | [.metadata.name, "nvidia.com/gpu.clique", .metadata.labels["nvidia.com/gpu.clique"]] | @tsv') | \
    column -t
NODE                  LABEL                  CLIQUE
gb-nvl-043-compute03  nvidia.com/gpu.clique  6a130f54-faaa-4b8f-847f-be44ab70f917.32766
gb-nvl-043-compute09  nvidia.com/gpu.clique  6a130f54-faaa-4b8f-847f-be44ab70f917.32766
```
At this point, I am able to run the simple IMEX channel injection sample and the Multi-Node MPI test from https://github.com/NVIDIA/k8s-dra-driver-gpu/discussions/249.

Now, let's try enabling the GPU portion of the DRA driver and verify that the operator emits an error message indicating this is an invalid configuration:
```
$ kubectl patch clusterpolicies.nvidia.com/cluster-policy     --type='json'     -p='[{"op": "replace", "path": "/spec/draDriver/gpus/enabled", "value":true}]'
clusterpolicy.nvidia.com/cluster-policy patched

$ kubectl get clusterpolicies.nvidia.com/cluster-policy     
NAME             STATUS     AGE
cluster-policy   notReady   2025-07-19T18:54:37Z

$ kubectl get clusterpolicy -ojsonpath='{.items[0].status}{"\n"}' | jq .
{
  "conditions": [
    {
      "lastTransitionTime": "2025-07-19T19:01:28Z",
      "message": "",
      "reason": "Error",
      "status": "False",
      "type": "Ready"
    },
    {
      "lastTransitionTime": "2025-07-19T19:01:28Z",
      "message": "failed to initialize ClusterPolicy controller: ClusterPolicy validation failed: failed to validate DRA: the device-plugin and DRA driver for GPUs cannot both be enabled in ClusterPolicy",
      "reason": "ReconcileFailed",
      "status": "True",
      "type": "Error"
    }
  ],
  "namespace": "gpu-operator",
  "state": "notReady"
}
```

If a user tries to enable the GPU portion of the DRA driver during a helm install or upgrade and forgets to disable the device-plugin, we emit an error message:
```
$ helm upgrade gpu-operator gpu-operator-v25.8.0-devel.tgz -n gpu-operator --set draDriver.gpus.enabled=true
Error: UPGRADE FAILED: execution error at (gpu-operator/templates/validation.yaml:7:4): 
NVIDIA device plugin and NVIDIA DRA Driver for GPUs cannot both be enabled
```

Let's disable the device plugin and then enable the GPU portion of the DRA driver:
```
$ kubectl patch clusterpolicies.nvidia.com/cluster-policy     --type='json'     -p='[{"op": "replace", "path": "/spec/devicePlugin/enabled", "value":false}]'           
clusterpolicy.nvidia.com/cluster-policy patched

$ kubectl patch clusterpolicies.nvidia.com/cluster-policy     --type='json'     -p='[{"op": "replace", "path": "/spec/draDriver/gpus/enabled", "value":true}]' 
clusterpolicy.nvidia.com/cluster-policy patched

$ kubectl get pods -n gpu-operator -lapp=nvidia-dra-driver-kubelet-plugin
NAME                                     READY   STATUS    RESTARTS   AGE
nvidia-dra-driver-kubelet-plugin-9zv7q   2/2     Running   0          23s
nvidia-dra-driver-kubelet-plugin-mtpvd   2/2     Running   0          20s

$  kubectl get resourceslice
NAME                                                   NODE                   DRIVER                      POOL                   AGE
gb-nvl-043-compute03-compute-domain.nvidia.com-kwz75   gb-nvl-043-compute03   compute-domain.nvidia.com   gb-nvl-043-compute03   31s
gb-nvl-043-compute03-gpu.nvidia.com-6vgqv              gb-nvl-043-compute03   gpu.nvidia.com              gb-nvl-043-compute03   30s
gb-nvl-043-compute09-compute-domain.nvidia.com-chjk5   gb-nvl-043-compute09   compute-domain.nvidia.com   gb-nvl-043-compute09   27s
gb-nvl-043-compute09-gpu.nvidia.com-th4gj              gb-nvl-043-compute09   gpu.nvidia.com              gb-nvl-043-compute09   27s

$ kubectl get clusterpolicies.nvidia.com/cluster-policy     
NAME             STATUS   AGE
cluster-policy   ready    2025-07-19T18:54:37Z
```
Run a sample GPU workload using DRA. The sample runs one pod with two containers. Each container shares access to the same GPU:
```
$ kubectl apply -f https://raw.githubusercontent.com/NVIDIA/k8s-dra-driver-gpu/refs/heads/main/demo/specs/quickstart/gpu-test2.yaml
namespace/gpu-test2 created
resourceclaimtemplate.resource.k8s.io/single-gpu created
pod/pod created

$ kubectl get pods -n gpu-test2 
NAME   READY   STATUS    RESTARTS   AGE
pod    2/2     Running   0          9s

$ kubectl logs -n gpu-test2 pod -c ctr0
GPU 0: NVIDIA GB200 (UUID: GPU-6fd2fa89-7bca-5235-b73a-0278535e1dd3)

$ kubectl logs -n gpu-test2 pod -c ctr1
GPU 0: NVIDIA GB200 (UUID: GPU-6fd2fa89-7bca-5235-b73a-0278535e1dd3)
```


assets/state-dra-driver/0200_clusterrole.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nvidia-dra-driver
+  namespace: "FILLED BY THE OPERATOR"
+rules:
+  - apiGroups: ["resource.nvidia.com"]
```
tariq1890: nit: Let's stick to the same YAML format when specifying arrays as seen in other clusterrole manifests like [this one](https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-driver/0210_clusterrole.yaml) for example

assets/state-dra-driver/0600_configmap.yaml
```diff
@@ -0,0 +1,40 @@
...
+    #!/bin/sh
+
+    if [ "$#" -ne 1 ]; then
+      echo "Usage: $0 COMMAND"
+      exit 1
+    fi
+    
+    entrypoint=$1
+    
+    until [[ -f /run/nvidia/validations/driver-ready ]]
```
tariq1890: Is this POSIX-compliant?

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -756,6 +778,84 @@ spec:
...
+          verbs:
+          - get
+          - list
+          - watch
+          - create
+          - update
+          - patch
+          - delete
+        - apiGroups:
+          - resource.k8s.io
```
tariq1890, line 825: For my reference, what is the k8s version that introduces these APIs?

assets/state-dra-driver/0500_deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+      app: nvidia-dra-driver-controller
+  template:
+    metadata:
+      labels:
+        app: nvidia-dra-driver-controller
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: nvidia-dra-driver
+      containers:
+        - name: compute-domain
```
tariq1890: ~We currently follow the convention of placing the array elements in the same indent level as the parent array field. Let's follow that here~

tariq1890: oh never mind

assets/state-dra-driver/0700_daemonset.yaml
```diff
@@ -0,0 +1,171 @@
...
+              mountPath: /run/nvidia/validations
+              mountPropagation: Bidirectional
+            - name: host-root
+              mountPath: /host
+              readOnly: true
+              mountPropagation: HostToContainer
+            - name: host-dev-char
+              mountPath: /host-dev-char
+      containers:
+      - name: compute-domains
```
tariq1890: needs indent.

assets/state-dra-driver/0500_deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+                  fieldPath: metadata.namespace
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: "node-role.kubernetes.io/control-plane"
+                    operator: "Exists"
+      tolerations:
+      - key: node-role.kubernetes.io/control-plane
```
tariq1890: needs indent

deployments/gpu-operator/templates/upgrade_crd.yaml
```diff
@@ -88,6 +88,7 @@ spec:
           - >
             kubectl apply -f /opt/gpu-operator/nvidia.com_clusterpolicies.yaml;
             kubectl apply -f /opt/gpu-operator/nvidia.com_nvidiadrivers.yaml;
+            kubectl apply -f /opt/gpu-operator/resource.nvidia.com_computedomains.yaml;
```
tariq1890: We should probably run this only if both draDriver and computeDomains are enabled, yes?

assets/state-dra-driver/0200_clusterrole.yaml
```diff
@@ -0,0 +1,33 @@
...
+    resources: ["computedomains"]
```
guptaNswati: Why are these permissions listed individually?

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -849,6 +851,104 @@ type SandboxDevicePluginSpec struct {
...
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Resource Requirements"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:resourceRequirements"
+	Resources *ResourceRequirements `json:"resources,omitempty"`
+
+	// Optional: Set tolerations
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Tolerations"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:io.kubernetes:Tolerations"
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
```
guptaNswati, line 1009: why Tolerations?

tariq1890: We add tolerations to every component managed by the gpu-operator. It's common for gpu-operator operands to run on nodes with specific taints, so we need to support setting the corresponding tolerations to the operands

assets/state-dra-driver/0200_clusterrole.yaml
tariq1890: Since they all have the same verbs, I would lean towards condensing this clusterrole yaml

controllers/object_controls.go
```diff
@@ -3748,17 +3820,63 @@ func getDaemonsetControllerRevisionHash(ctx context.Context, daemonset *appsv1.D
...
 }
 
+// TransformDRADriverController transforms nvidia-dra-driver-controller deployment with required config as per ClusterPolicy
+func TransformDRADriverController(obj *appsv1.Deployment, spec *gpuv1.ClusterPolicySpec) error {
+	config := spec.DRADriver
+	image, err := gpuv1.ImagePath(&config)
+	if err != nil {
+		return err
+	}
+	obj.Spec.Template.Spec.Containers[0].Image = image
```
tariq1890: Two comments

- Let's extract the image index by name, instead of a hardcoded reference to the zero-index
- move `obj.Spec.Template.Spec.Containers[i]` to a variable to avoid redundancy. I know we've done this in the past, but we should break away from this practice

assets/state-dra-driver/0200_clusterrole.yaml
cdesiniotis: Updated.

assets/state-dra-driver/0200_clusterrole.yaml
cdesiniotis: I have condensed the ClusterRole object as suggested. I originally copied the ClusterRole unmodified from the standalone helm chart.

assets/state-dra-driver/0500_deployment.yaml
cdesiniotis: Updated.

assets/state-dra-driver/0600_configmap.yaml
cdesiniotis: Good catch. Updated.

assets/state-dra-driver/0700_daemonset.yaml
cdesiniotis: Updated.

controllers/object_controls.go
cdesiniotis: Updated to address your feedback.

deployments/gpu-operator/templates/upgrade_crd.yaml
cdesiniotis: I am not sure actually and I wanted to discuss this. As currently implemented, the computedomains CRD always gets installed, regardless if computeDomains are enabled. What should the desired behavior be?

controllers/state_manager.go
```diff
@@ -220,6 +222,38 @@ func KubernetesVersion() (string, error) {
...
+		for _, resource := range resourceList.APIResources {
+			if resource.Kind == kind {
+				matches = append(matches, resourceList.GroupVersion)
+			}
+		}
+	}
+
+	draSupported := len(matches) > 0
+	if draSupported {
+		logger.Info(fmt.Sprintf("Kind '%s' exists in the following group/versions: %s", kind, strings.Join(matches, ", ")))
```
tariq1890: Does this work?
```suggestion
		logger.Info(fmt.Sprintf("Kind %q exists in the following group/versions: %s", kind, strings.Join(matches, ", ")))
```

cdesiniotis: Updated.

deployments/gpu-operator/templates/upgrade_crd.yaml
tariq1890: I am not sure what's the benefit of installing this CRD when we are not going to use it

cdesiniotis: We always install the NVIDIADriver CRD even if users don't enable that feature :) 

I guess the alternative to my current approach is to render this CRD in the controller. That way, we can create / delete the CRD depending on if compute domains are enabled / disabled in ClusterPolicy. We don't have precedent for rendering CRDs in the controller, but I don't see why we couldn't do it.

tariq1890: Oh I am not saying we create CRDs via the controller.  I am talking about adding an if condition in the helm chart

tariq1890: Actually,after thinking about this more. I am okay with leaving this as-is.

Resolving my comment

assets/state-dra-driver/0110_compute_domain_daemon-service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: compute-domain-daemon-service-account
+  namespace: "FILLED BY THE OPERATOR"
```
elezar: nit: newline

assets/state-dra-driver/0320_compute_domain_daemon-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: compute-domain-daemon-clusterrole-binding
+  namespace: "FILLED BY THE OPERATOR"
+subjects:
+  - kind: ServiceAccount
+    name: compute-domain-daemon-service-account
+    namespace: "FILLED BY THE OPERATOR"
+roleRef:
+  kind: ClusterRole
+  name: compute-domain-daemon-clusterrole
+  apiGroup: rbac.authorization.k8s.io
```
elezar: nit: newline

assets/state-dra-driver/0210_role.yaml
cdesiniotis, line 1: Note, the NVIDIA DRA driver today assumes it has RBAC permissions for `daemonsets` at the cluster-scope. Restricting the permissions to be namespace-scoped (as I have done here) will only work if the following change is merged in the upstream NVIDIA DRA driver: https://github.com/NVIDIA/k8s-dra-driver-gpu/pull/434.

controllers/clusterpolicy_validator.go
```diff
@@ -0,0 +1,43 @@
...
+	err := validateDRA(n.singleton, n.draSupported)
+	if err != nil {
+		return fmt.Errorf("failed to validate DRA: %w", err)
+	}
+	return nil
+}
+
+func validateDRA(clusterpolicy *gpuv1.ClusterPolicy, draSupported bool) error {
+	if !draSupported && clusterpolicy.Spec.DRADriver.IsEnabled() {
+		return fmt.Errorf("the NVIDIA DRA driver is enabled in ClusterPolicy but Dynamic Resource Allocation is not enabled in the kubernetes cluster")
```
jgehrcke: Two writing style aspects:

- "the NVIDIA DRA Driver for GPUs" or "NVIDIA's DRA Driver for GPUs
- "Kubernetes" (capitalized)

jgehrcke: (consistency is important here because this seems to show up towards users, as shown in the current PR description)

controllers/clusterpolicy_validator.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if clusterpolicy.Spec.DevicePlugin.IsEnabled() && clusterpolicy.Spec.DRADriver.IsGPUsEnabled() {
+		return fmt.Errorf("the device-plugin and DRA driver for GPUs cannot both be enabled in ClusterPolicy")
```
jgehrcke: If this message may be exposed to users we should write "NVIDIA's DRA driver for GPUs".

jgehrcke: And now I've browsed the Internet a bit to see how we commonly describe the device plugin in prose -- we should probably write "NVIDIA device plugin" here.

Genitive is an OK reason to deviate, so I think we should decide between:

1) "NVIDIA's device plugin and NVIDIA's DRA Driver for GPUs ..." or
2) "The NVIDIA device plugin and the NVIDIA DRA Driver for GPUs ..."

api/nvidia/v1/clusterpolicy_types.go
jgehrcke: To connect some dots, some vaguely related locations in code and discussion:

- https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/9951be59ac1bf827d02a1f0f3dfc0cea65c7aca3/deployments/helm/nvidia-dra-driver-gpu/templates/controller.yaml#L81
- https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/9951be59ac1bf827d02a1f0f3dfc0cea65c7aca3/deployments/helm/nvidia-dra-driver-gpu/values.yaml#L63
- https://github.com/NVIDIA/k8s-dra-driver-gpu/issues/308

deployments/gpu-operator/templates/upgrade_crd.yaml
jgehrcke: Interesting. I have no opinion really.

>  the NVIDIADriver CRD

What is that CRD for? Is that for the GPU driver?

Maybe one mildly valuable thought is: when people install the GPU Operator they are (by a large factor) more likely to use the GPU driver than they are likely to use ComputeDomains.

I don't have an intuition for how much of an "environment pollution" it is considered for operators to install CRDs that are potentially not used. Maybe it's really not considered a pollution at all, and it's completely fine and expected for something like GPU Operator to always install all CRDs regardless of potentially disabled sub-features.

jgehrcke (commented):
Thanks, Chris! That already looks great (as far as I can judge, and I can't judge a lot of things in here).

I am not really familiar with the cluster policy approach you've shown in the PR description, so I focus more on the Helm cmdline way of installing and configuring. Here, I have two questions about just-in-time actionable error feedback we (hopefully can) give to users when they submit an install command that we know cannot lead to success:

1) Can you show how a `helm install ...` command would fail right away when a user picks a bad config parameter combination? (example: when the device plugin is implicitly or explicitly enabled, and when `--set draDriver.gpus.enabled=true`). If we provide a great error message in that case right away and let `helm install ...` exit non-zero then I can think users may really appreciate that.

2) Can we make `helm install ...` fail with an actionable error message when 1) the DRA feature gate isn't enabled or the DRA API group isn't enabled and 2) either `draDriver.gpus.enabled=true` or `draDriver.computeDomains.enabled=true`?


About (2), I can see that it might require a bit of work, but I can also imagine that it may pay off in terms of reducing support request noise... :) For completeness, I asked in the planning doc:

> Let's think about (install-time?) validation with actionable feedback: is DRA / the API group `resource.k8s.io/...`. Or if that's not a good idea: maybe we can have a validator in the GPU operator emit a nice error message when it's apparent that the k8s cluster isn't ready for DRA.

cdesiniotis:
@jgehrcke thanks for the feedback! I agree we should aim to validate as much as we can at helm install / upgrade time and fail-fast if user's misconfigure our chart. I have attempted to address both of your questions by adding validations to our helm chart. @klueska pointed me to https://github.com/NVIDIA/k8s-device-plugin/blob/012c1a652fbc001beeb4836c2763f63c899af5f2/deployments/helm/nvidia-device-plugin/templates/validation.yml which gave me inspiration.

For (1), below is what users will see if both `devicePlugin.enabled` and `draDriver.gpus.enabled` are set to true.
```
$ helm install gpu-operator-foo gpu-operator-v25.8.0-devel.tgz -n gpu-operator --set draDriver.gpus.enabled=true
Error: INSTALLATION FAILED: execution error at (gpu-operator/templates/validation.yaml:7:4): 
The NVIDIA device plugin and the NVIDIA DRA Driver for GPUs cannot both be enabled
```
For (2), below is what users will see if they attempt to enable the DRA driver but their Kubernetes cluster does not support DRA. To check for DRA support, I am currently checking for the presence of the DeviceClass resource, which is what the official [docs](https://kubernetes.io/docs/tasks/configure-pod-container/assign-resources/set-up-dra-cluster/#verify) recommend. Let me know if there is a "better" or more robust check we can implement for this.
```
$ kubectl version
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.31.9

$ helm install gpu-operator-foo gpu-operator-v25.8.0-devel.tgz -n gpu-operator --set draDriver.computeDomains.enabled=true
Error: INSTALLATION FAILED: execution error at (gpu-operator/templates/validation.yaml:13:4): 
Cannot enable the NVIDIA DRA Driver for GPUs on a Kubernetes cluster that does not support DRA
```

controllers/clusterpolicy_validator.go
cdesiniotis: Thanks for being thorough here! I have updated the error message. I have gone with "the NVIDIA DRA Driver for GPUs"

controllers/clusterpolicy_validator.go
cdesiniotis: I have updated the error message. I have gone with "the NVIDIA device plugin and the NVIDIA DRA Driver for GPUs . . ."

assets/state-dra-driver/0320_compute_domain_daemon-clusterrolebinding.yaml
cdesiniotis: Updated.

assets/state-dra-driver/0110_compute_domain_daemon-service_account.yaml
cdesiniotis: Updated.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -850,6 +852,104 @@ type SandboxDevicePluginSpec struct {
...
+
+	// Image pull secrets
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Image pull secrets"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:io.kubernetes:Secret"
+	ImagePullSecrets []string `json:"imagePullSecrets,omitempty"`
+
+	// GPUs defines configuration for GPUs in the NVIDIA DRA Driver
+	GPUs DRADriverGPUs `json:"gpus,omitempty"`
```
guptaNswati, line 958: this sounds more intuitive : DRAGPUDriver and DRAComputeDomainDriver

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -850,6 +852,104 @@ type SandboxDevicePluginSpec struct {
...
+
+	// Controller defines configuration for the NVIDIA DRA Driver controller
+	Controller DRADriverController `json:"controller,omitempty"`
+
+	// KubeletPlugin defines configuration for the NVIDIA DRA Driver kubelet plugin
+	KubeletPlugin DRADriverKubeletPlugin `json:"kubeletPlugin,omitempty"`
+}
+
+// DRADriverController defines configuration for the NVIDIA DRA Driver controller
+type DRADriverController struct {
```
guptaNswati, line 992: Nitpick: DRAComputeDomainController

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -850,6 +852,104 @@ type SandboxDevicePluginSpec struct {
...
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Environment Variables"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:text"
+	Env []EnvVar `json:"env,omitempty"`
+
+	// Optional: Define resources requests and limits
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Resource Requirements"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:resourceRequirements"
+	Resources *ResourceRequirements `json:"resources,omitempty"`
```
guptaNswati, line 1003: are there any resource limitations for the controller? can it be resource intensive? do we set any limits in our upstream dra-driver

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1793,6 +1893,9 @@ func ImagePath(spec interface{}) (string, error) {
 	case *SandboxDevicePluginSpec:
 		config := spec.(*SandboxDevicePluginSpec)
 		return imagePath(config.Repository, config.Image, config.Version, "SANDBOX_DEVICE_PLUGIN_IMAGE")
+	case *DRADriverSpec:
```
guptaNswati, line 2150: what does this do for dcgm-exporter. we will also need a flag to enable DRA metrics when dra driver is installed and let dcgm-exporter emit those metrics. Its not turned on by default.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1897,6 +2000,21 @@ func (p *DevicePluginSpec) IsEnabled() bool {
 	return *p.Enabled
 }
 
+// IsEnabled returns true if the DRA Driver is enabled through gpu-operator
+func (d *DRADriverSpec) IsEnabled() bool {
+	return d.IsGPUsEnabled() || d.IsComputeDomainsEnabled()
+}
+
+// IsGPUsEnabled returns true if the GPUs resource is enabled in the DRA Driver
+func (d *DRADriverSpec) IsGPUsEnabled() bool {
```
guptaNswati, line 2274: nitpick: isGPUDriverEnabled and is ComputeDomainDriverEnabled

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1897,6 +2000,21 @@ func (p *DevicePluginSpec) IsEnabled() bool {
...
+	return d.GPUs.Enabled != nil && *d.GPUs.Enabled
+}
+
+// IsComputeDomainsEnabled returns true if the ComputeDomains resource is enabled in the DRA Driver
+func (d *DRADriverSpec) IsComputeDomainsEnabled() bool {
+	return d.ComputeDomains.Enabled != nil && *d.ComputeDomains.Enabled
+}
+
 // IsEnabled returns true if dcgm-exporter is enabled(default) through gpu-operator
 func (e *DCGMExporterSpec) IsEnabled() bool {
```
guptaNswati, line 2284: whats the role of this?

controllers/clusterpolicy_validator.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (n *ClusterPolicyController) validateClusterPolicy() error {
+	err := validateDRA(n.singleton, n.draSupported)
+	if err != nil {
+		return fmt.Errorf("failed to validate DRA: %w", err)
+	}
+	return nil
+}
+
+func validateDRA(clusterpolicy *gpuv1.ClusterPolicy, draSupported bool) error {
```
guptaNswati: can we also add a simple imex injection and gpu test run when DRA is enabled.

assets/state-dra-driver/0200_clusterrole.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nvidia-dra-driver
+  namespace: "FILLED BY THE OPERATOR"
+rules:
+  - apiGroups:
+      - resource.nvidia.com
+    resources:
+      - computedomains
```
guptaNswati, line 10: why is there a separate clusterrole for computedomain daemon?

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
kannon92: According to https://github.com/kubernetes/enhancements/issues/4381, 1.30 introduced this for structured parameters.

I think it may have been earlier when classical DRA was introduced.

kannon92: Yes, https://github.com/kubernetes/enhancements/issues/3063.

Looks like 1.27 introduced the resource api.

controllers/clusterpolicy_validator.go
```diff
@@ -0,0 +1,43 @@
...
+	if !draSupported && clusterpolicy.Spec.DRADriver.IsEnabled() {
+		return fmt.Errorf("the NVIDIA DRA driver for GPUs is enabled in ClusterPolicy but Dynamic Resource Allocation is not enabled in the Kubernetes cluster")
+	}
+
+	if clusterpolicy.Spec.DevicePlugin.IsEnabled() && clusterpolicy.Spec.DRADriver.IsGPUsEnabled() {
+		return fmt.Errorf("the NVIDIA device plugin and the NVIDIA DRA driver for GPUs cannot both be enabled in ClusterPolicy")
+	}
```
bg-furiosa: pin point for device mgnt wg slack channel
- https://kubernetes.slack.com/archives/C0409NGC1TK/p1756152632076249?thread_ts=1755901961.772459&cid=C0409NGC1TK

dims:
Is this accurate summary (so far!) @cdesiniotis ?

Integrates the [NVIDIA DRA Driver for GPUs](https://github.com/NVIDIA/k8s-dra-driver-gpu) as a first-class managed operand within the GPU Operator. DRA (Dynamic Resource Allocation) is a Kubernetes-native mechanism (GA in k8s 1.32) for fine-grained, structured device sharing — an alternative to the classic device plugin model. The GPU Operator can now deploy and lifecycle-manage the DRA driver alongside existing operands.

### 1. New API Surface (`ClusterPolicy`)
Adds a `draDriver` section to `ClusterPolicySpec` with two sub-features:
- `draDriver.computeDomains` — enables the compute-domain DRA driver (IMEX/NVLink fabric topology, multi-node GPU cliques)
- `draDriver.gpus` — enables the GPU DRA driver (structured GPU resource advertising via `ResourceSlice`); **mutually exclusive** with the device plugin
### 2. New State: `state-dra-driver`
New manifests under `assets/state-dra-driver/`:
- ServiceAccounts for driver and compute-domain daemon
- ClusterRoles/Roles/Bindings for controller, kubelet-plugin, compute-domain daemon
- 4 DeviceClass objects: `compute-domain-daemon`, `compute-domain-default-channel`, `gpu`, `mig`
- Controller Deployment
- ConfigMap for driver config
- DaemonSet for DRA kubelet-plugin (privileged on OpenShift via dedicated RoleBinding)
### 3. Controller Integration
- `state_manager.go` — new `StateDRADriver` handler registered in the state machine
- `object_controls.go` — ~290 lines of DRA reconciliation (controller Deployment, kubelet-plugin DaemonSet, ConfigMap); applies standard GPU Operator transforms (image overrides, tolerations, resource limits)
- `resource_manager.go` — updated for DRA cluster-level permissions
### 4. Validation (Two Layers)
- **Helm-time** (`validation.yaml`): blocks install/upgrade if DRA enabled on non-DRA cluster, or device-plugin + DRA GPUs both enabled
- **Operator-time** (`clusterpolicy_validator.go`): checks `resource.k8s.io` API group at startup; sets `ClusterPolicy` notReady with human-readable error if DRA requested but not supported
### 5. Helm Chart Updates
- `values.yaml`: new `draDriver.computeDomains.*` and `draDriver.gpus.*` values (disabled by default)
- `clusterrole.yaml` / `role.yaml` expanded with DRA permissions
- `clusterpolicy.yaml` template updated
- CRDs regenerated; `upgrade_crd.yaml` / `cleanup_crd.yaml` updated for ComputeDomain CRD
### 6. Tests
- Unit tests in `controllers/transforms_test.go` (+330 lines) for DRA object transforms
- Manual E2E documented in PR: non-DRA cluster (error path), DRA cluster with compute-domains only, DRA cluster with GPU DRA (device-plugin disabled)

| Decision | Rationale |
|---|---|
| DRA GPUs mutually exclusive with device-plugin | Prevents double-allocation of the same GPU resources |
| ComputeDomains can coexist with device-plugin | Compute-domain (IMEX) resources are distinct from raw GPU resources |
| Disabled by default | DRA requires k8s ≥1.32; safe opt-in for existing users |
| DeviceClass apiVersion conditional | Handles both `v1alpha3` (k8s 1.31) and `v1beta1` (k8s 1.32+) |
| OpenShift privileged via separate RoleBinding | Follows existing GPU Operator pattern for OpenShift SCCs |

cdesiniotis:
Superceded by https://github.com/NVIDIA/gpu-operator/pull/2571

PR #1543, opened 2025-07-18
Add 'nodefeatures/finalizers' for nfs worker Role

etharendil:
I've run into following issue:
```
worker E0718 10:20:12.156046       1 main.go:91] "error while running" err="failed to advertise features (via CRD API): failed to update NodeFeature object \"skynet-dev-4dxsm\": nodefeatures.nfd.k8s-sigs.io \"skynet-dev-4dxsm\" is forbidden: cannot set an ownerRef on a resource you can't delete: , <nil>"
```
NFD role does not have permissions for the finalizers and failing 

helm values:
```
node-feature-discovery:
  worker:
    serviceAccount:
      name: node-feature-discovery-worker
      create: true
```

### DIFF
--- a/deployments/gpu-operator/charts/node-feature-discovery/templates/role.yaml
+++ b/deployments/gpu-operator/charts/node-feature-discovery/templates/role.yaml
@@ -11,6 +11,7 @@ rules:
   - nfd.k8s-sigs.io
   resources:
   - nodefeatures
+  - nodefeatures/finalizers
   verbs:
   - create
   - get


deployments/gpu-operator/charts/node-feature-discovery/templates/role.yaml
```diff
@@ -11,6 +11,7 @@ rules:
   - nfd.k8s-sigs.io
   resources:
   - nodefeatures
+  - nodefeatures/finalizers
```
tariq1890, line 14: You'll need to open a PR against the upstream node-feature-discovery repository. This is the file you want to change: https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/deployment/helm/node-feature-discovery/templates/role.yaml

etharendil: thank you @tariq1890 
I'll do so

Issue #1340, opened 2025-03-13
Driver should mount license from a Secret not a ConfigMap


lindhe:
Today, the GPU license token may be mounted via a ConfigMap to the driver pod:

https://github.com/NVIDIA/gpu-operator/blob/55508f8384b114078d366a6d5ca8dd119f9c4626/deployments/gpu-operator/templates/nvidiadriver.yaml#L51-L55

I assume the license is sensitive information, so it should really be mounted from a Secret not a ConfigMap. This would let users handle the license contents securely, e.g. via SealedSecrets or ExternalSecrets operators, and just in general help distinguish between sensitive information and non-sensitive information.

patchon:
This implementation appears to be problematic.

Sensitive information should be stored using Kubernetes Secrets, as intended. Using a ConfigMap for this purpose not only raises security concerns, but also undermines the principles of a secure and consistent GitOps workflow.

At a minimum, users should be given the option to choose between using a Secret or a ConfigMap—although it's unclear why a ConfigMap would be preferred for sensitive data in most cases.

bartekle:
Why can't the license/gridd be mounted from the host? Driver already needs to access host so it could just look in predefined places eg. /etc/nvidia/ClientConfigToken/? This would be ideal for dynamic vendor data mechanism

tariq1890:
This is fixed and GPU Operator 25.10.0 supports mounting licenses from Secrets

PR #1544, opened 2025-07-19
Use a Secret instead of a ConfigMap for NLS licensing token

shivakunv:
refer: [https://github.com/NVIDIA/gpu-operator/issues/1340](https://github.com/NVIDIA/gpu-operator/issues/1340)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,15 @@ linters:
       - third_party$
       - builtin$
       - examples$
+    rules:
+      - linters:
+          - staticcheck
+        path: (.+)\.go$
+        text: 'SA1019: config.Driver.LicensingConfig.ConfigMapName is deprecated(.+)'
+      - linters:
+          - staticcheck
+        path: (.+)\.go$
+        text: 'SA1019: cr.Spec.LicensingConfig.Name is deprecated(.+)'
 formatters:
   enable:
     - gofmt
--- a/api/nvidia/v1/clusterpolicy_types.go
+++ b/api/nvidia/v1/clusterpolicy_types.go
@@ -1120,11 +1120,19 @@ type DriverCertConfigSpec struct {
 
 // DriverLicensingConfigSpec defines licensing server configuration for NVIDIA Driver container
 type DriverLicensingConfigSpec struct {
+	// Deprecated: ConfigMapName has been deprecated in favour of SecretName. Please use secrets to handle the licensing server configuration more securely
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="ConfigMap Name"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	ConfigMapName string `json:"configMapName,omitempty"`
+
+	// SecretName indicates the name of the secret containing the licensing token
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Secret Name"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
+	SecretName string `json:"secretName,omitempty"`
 
 	// NLSEnabled indicates if NVIDIA Licensing System is used for licensing.
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
@@ -1898,6 +1906,14 @@ func (d *DriverSpec) UsePrecompiledDrivers() bool {
 // OpenKernelModulesEnabled returns true if driver install is enabled using open GPU kernel modules
 func (d *DriverSpec) OpenKernelModulesEnabled() bool {
 	return d.KernelModuleType == "open"
+}
+
+// IsVGPULicensingEnabled returns true if the vgpu driver license config is provided
+func (d *DriverSpec) IsVGPULicensingEnabled() bool {
+	if d.LicensingConfig == nil {
+		return false
+	}
+	return d.LicensingConfig.ConfigMapName != "" || d.LicensingConfig.SecretName != ""
 }
 
 // IsEnabled returns true if device-plugin is enabled(default) through gpu-operator
--- a/api/nvidia/v1alpha1/nvidiadriver_types.go
+++ b/api/nvidia/v1alpha1/nvidiadriver_types.go
@@ -430,6 +430,13 @@ type DriverRepoConfigSpec struct {
 type DriverLicensingConfigSpec struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Secret Name"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
+	SecretName string `json:"secretName,omitempty"`
+
+	// Deprecated: ConfigMapName has been deprecated in favour of SecretName. Please use secrets to handle the licensing server configuration more securely
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="ConfigMap Name"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	Name string `json:"name,omitempty"`
@@ -688,7 +695,7 @@ func (d *NVIDIADriverSpec) IsVGPULicensingEnabled() bool {
 	if d.LicensingConfig == nil {
 		return false
 	}
-	return d.LicensingConfig.Name != ""
+	return d.LicensingConfig.Name != "" || d.LicensingConfig.SecretName != ""
 }
 
 // IsKernelModuleConfigEnabled returns true if kernel module config is provided
--- a/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+++ b/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
@@ -72,7 +72,7 @@ metadata:
               },
               "licensingConfig": {
                 "nlsEnabled": true,
-                "configMapName": ""
+                "secretName": ""
               },
               "virtualTopology": {
                 "config": ""
@@ -170,7 +170,7 @@ metadata:
             },
             "licensingConfig": {
               "nlsEnabled": true,
-              "name": ""
+              "secretName": ""
             },
             "virtualTopologyConfig": {
               "name": ""
--- a/bundle/manifests/nvidia.com_clusterpolicies.yaml
+++ b/bundle/manifests/nvidia.com_clusterpolicies.yaml
@@ -685,11 +685,18 @@ spec:
                       licensing'
                     properties:
                       configMapName:
+                        description: 'Deprecated: ConfigMapName has been deprecated
+                          in favour of SecretName. Please use secrets to handle the
+                          LicensingConfig more securely'
                         type: string
                       nlsEnabled:
                         description: NLSEnabled indicates if NVIDIA Licensing System
                           is used for licensing.
                         type: boolean
+                      secretName:
+                        description: SecretName indicates the name of the secret containing
+                          the licensing token
+                        type: string
                     type: object
                   livenessProbe:
                     description: NVIDIA Driver container liveness probe settings
--- a/bundle/manifests/nvidia.com_nvidiadrivers.yaml
+++ b/bundle/manifests/nvidia.com_nvidiadrivers.yaml
@@ -234,11 +234,16 @@ spec:
                 description: 'Optional: Licensing configuration for NVIDIA vGPU licensing'
                 properties:
                   name:
+                    description: 'Deprecated: ConfigMapName has been deprecated in
+                      favour of SecretName. Please use secrets to handle the LicensingConfig
+                      more securely'
                     type: string
                   nlsEnabled:
                     description: NLSEnabled indicates if NVIDIA Licensing System is
                       used for licensing.
                     type: boolean
+                  secretName:
+                    type: string
                 type: object
               livenessProbe:
                 description: NVIDIA Driver container liveness probe settings
--- a/config/crd/bases/nvidia.com_clusterpolicies.yaml
+++ b/config/crd/bases/nvidia.com_clusterpolicies.yaml
@@ -685,11 +685,18 @@ spec:
                       licensing'
                     properties:
                       configMapName:
+                        description: 'Deprecated: ConfigMapName has been deprecated
+                          in favour of SecretName. Please use secrets to handle the
+                          licensing server configuration more securely'
                         type: string
                       nlsEnabled:
                         description: NLSEnabled indicates if NVIDIA Licensing System
                           is used for licensing.
                         type: boolean
+                      secretName:
+                        description: SecretName indicates the name of the secret containing
+                          the licensing token
+                        type: string
                     type: object
                   livenessProbe:
                     description: NVIDIA Driver container liveness probe settings
--- a/config/samples/nvidia_v1alpha1_nvidiadriver.yaml
+++ b/config/samples/nvidia_v1alpha1_nvidiadriver.yaml
@@ -26,7 +26,7 @@ spec:
     name: ""
   # vGPU licensing configuration
   licensingConfig:
-    name: ""
+    secretName: ""
     nlsEnabled: true
   # vGPU topology daemon configuration
   virtualTopologyConfig:
--- a/config/samples/v1_clusterpolicy.yaml
+++ b/config/samples/v1_clusterpolicy.yaml
@@ -90,7 +90,7 @@ spec:
       name: ""
     # vGPU licensing configuration
     licensingConfig:
-      configMapName: ""
+      secretName: ""
       nlsEnabled: true
     # vGPU topology daemon configuration
     virtualTopology:
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -3266,6 +3266,52 @@ func createEmptyDirVolume(volumeName string) corev1.Volume {
 	}
 }
 
+func applyLicensingConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, driverContainer *corev1.Container) {
+	podSpec := &obj.Spec.Template.Spec
+
+	// add new volume mount
+	licensingConfigVolMount := corev1.VolumeMount{Name: "licensing-config", ReadOnly: true, MountPath: VGPULicensingConfigMountPath, SubPath: VGPULicensingFileName}
+	driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, licensingConfigVolMount)
+
+	// gridd.conf always mounted
+	licenseItemsToInclude := []corev1.KeyToPath{
+		{
+			Key:  VGPULicensingFileName,
+			Path: VGPULicensingFileName,
+		},
+	}
+	// client config token only mounted when NLS is enabled
+	if config.Driver.LicensingConfig.IsNLSEnabled() {
+		licenseItemsToInclude = append(licenseItemsToInclude, corev1.KeyToPath{
+			Key:  NLSClientTokenFileName,
+			Path: NLSClientTokenFileName,
+		})
+		nlsTokenVolMount := corev1.VolumeMount{Name: "licensing-config", ReadOnly: true, MountPath: NLSClientTokenMountPath, SubPath: NLSClientTokenFileName}
+		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, nlsTokenVolMount)
+	}
+
+	var licensingConfigVolumeSource corev1.VolumeSource
+	if config.Driver.LicensingConfig.SecretName != "" {
+		licensingConfigVolumeSource = corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: config.Driver.LicensingConfig.SecretName,
+				Items:      licenseItemsToInclude,
+			},
+		}
+	} else if config.Driver.LicensingConfig.ConfigMapName != "" {
+		licensingConfigVolumeSource = corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: config.Driver.LicensingConfig.ConfigMapName,
+				},
+				Items: licenseItemsToInclude,
+			},
+		}
+	}
+	licensingConfigVol := corev1.Volume{Name: "licensing-config", VolumeSource: licensingConfigVolumeSource}
+	podSpec.Volumes = append(podSpec.Volumes, licensingConfigVol)
+}
+
 func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) error {
 	podSpec := &obj.Spec.Template.Spec
 	driverContainer := findContainerByName(podSpec.Containers, "nvidia-driver-ctr")
@@ -3335,37 +3381,8 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 	}
 
 	// set any licensing configuration required
-	if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.ConfigMapName != "" {
-		licensingConfigVolMount := corev1.VolumeMount{Name: "licensing-config", ReadOnly: true, MountPath: VGPULicensingConfigMountPath, SubPath: VGPULicensingFileName}
-		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, licensingConfigVolMount)
-
-		// gridd.conf always mounted
-		licenseItemsToInclude := []corev1.KeyToPath{
-			{
-				Key:  VGPULicensingFileName,
-				Path: VGPULicensingFileName,
-			},
-		}
-		// client config token only mounted when NLS is enabled
-		if config.Driver.LicensingConfig.IsNLSEnabled() {
-			licenseItemsToInclude = append(licenseItemsToInclude, corev1.KeyToPath{
-				Key:  NLSClientTokenFileName,
-				Path: NLSClientTokenFileName,
-			})
-			nlsTokenVolMount := corev1.VolumeMount{Name: "licensing-config", ReadOnly: true, MountPath: NLSClientTokenMountPath, SubPath: NLSClientTokenFileName}
-			driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, nlsTokenVolMount)
-		}
-
-		licensingConfigVolumeSource := corev1.VolumeSource{
-			ConfigMap: &corev1.ConfigMapVolumeSource{
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: config.Driver.LicensingConfig.ConfigMapName,
-				},
-				Items: licenseItemsToInclude,
-			},
-		}
-		licensingConfigVol := corev1.Volume{Name: "licensing-config", VolumeSource: licensingConfigVolumeSource}
-		podSpec.Volumes = append(podSpec.Volumes, licensingConfigVol)
+	if config.Driver.IsVGPULicensingEnabled() {
+		applyLicensingConfig(obj, config, driverContainer)
 	}
 
 	// set virtual topology daemon configuration if specified for vGPU driver
@@ -3434,7 +3451,7 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 		}
 		volumeMounts, itemsToInclude, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
 		if err != nil {
-			return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %v", err)
+			return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
 		}
 		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
 		podSpec.Volumes = append(podSpec.Volumes, createConfigMapVolume(config.Driver.CertConfig.Name, itemsToInclude))
--- a/controllers/transforms_test.go
+++ b/controllers/transforms_test.go
@@ -156,6 +156,11 @@ func (d Daemonset) WithHostPID(enabled bool) Daemonset {
 	return d
 }
 
+func (d Daemonset) WithVolume(volume corev1.Volume) Daemonset {
+	d.Spec.Template.Spec.Volumes = append(d.Spec.Template.Spec.Volumes, volume)
+	return d
+}
+
 // Pod is a Pod wrapper used for testing
 type Pod struct {
 	*corev1.Pod
@@ -2634,4 +2639,151 @@ func TestGetRuntimeConfigFiles(t *testing.T) {
 		})
 	}
 
+}
+
+func TestTransformDriverWithLicensingConfig(t *testing.T) {
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-node",
+			Labels: map[string]string{
+				nfdKernelLabelKey: "6.8.0-60-generic",
+				commonGPULabelKey: "true",
+			},
+		},
+	}
+	mockClient := fake.NewFakeClient(node)
+
+	testCases := []struct {
+		description   string
+		ds            Daemonset
+		cpSpec        *gpuv1.ClusterPolicySpec
+		client        client.Client
+		expectedDs    Daemonset
+		errorExpected bool
+	}{
+		{
+			description: "transform driver dependent containers with secretName",
+			ds: NewDaemonset().WithContainer(corev1.Container{Name: "nvidia-driver-ctr"}).
+				WithInitContainer(corev1.Container{Name: "k8s-driver-manager"}),
+			cpSpec: &gpuv1.ClusterPolicySpec{
+				Driver: gpuv1.DriverSpec{
+					Repository:      "nvcr.io/nvidia",
+					Image:           "driver",
+					ImagePullPolicy: "IfNotPresent",
+					Version:         "570.172.08",
+					Manager: gpuv1.DriverManagerSpec{
+						Repository:      "nvcr.io/nvidia/cloud-native",
+						Image:           "k8s-driver-manager",
+						ImagePullPolicy: "IfNotPresent",
+						Version:         "v0.8.0",
+					},
+					LicensingConfig: &gpuv1.DriverLicensingConfigSpec{
+						SecretName: "test-secret",
+					},
+				},
+			},
+			client: mockClient,
+			expectedDs: NewDaemonset().WithContainer(corev1.Container{
+				Name:            "nvidia-driver-ctr",
+				Image:           "nvcr.io/nvidia/driver:570.172.08-",
+				ImagePullPolicy: corev1.PullIfNotPresent,
+				VolumeMounts: []corev1.VolumeMount{
+					{
+						Name:      "licensing-config",
+						ReadOnly:  true,
+						MountPath: VGPULicensingConfigMountPath,
+						SubPath:   VGPULicensingFileName,
+					},
+				},
+			}).WithInitContainer(corev1.Container{
+				Name:            "k8s-driver-manager",
+				Image:           "nvcr.io/nvidia/cloud-native/k8s-driver-manager:v0.8.0",
+				ImagePullPolicy: corev1.PullIfNotPresent,
+			}).WithVolume(corev1.Volume{
+				Name: "licensing-config",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: "test-secret",
+						Items: []corev1.KeyToPath{
+							{
+								Key:  VGPULicensingFileName,
+								Path: VGPULicensingFileName,
+							},
+						},
+					},
+				},
+			}),
+			errorExpected: false,
+		},
+		{
+			description: "transform driver dependent containers with configMapName",
+			ds: NewDaemonset().WithContainer(corev1.Container{Name: "nvidia-driver-ctr"}).
+				WithInitContainer(corev1.Container{Name: "k8s-driver-manager"}),
+			cpSpec: &gpuv1.ClusterPolicySpec{
+				Driver: gpuv1.DriverSpec{
+					Repository:      "nvcr.io/nvidia",
+					Image:           "driver",
+					ImagePullPolicy: "IfNotPresent",
+					Version:         "570.172.08",
+					Manager: gpuv1.DriverManagerSpec{
+						Repository:      "nvcr.io/nvidia/cloud-native",
+						Image:           "k8s-driver-manager",
+						ImagePullPolicy: "IfNotPresent",
+						Version:         "v0.8.0",
+					},
+					LicensingConfig: &gpuv1.DriverLicensingConfigSpec{
+						ConfigMapName: "test-configmap",
+					},
+				},
+			},
+			client: mockClient,
+			expectedDs: NewDaemonset().WithContainer(corev1.Container{
+				Name:            "nvidia-driver-ctr",
+				Image:           "nvcr.io/nvidia/driver:570.172.08-",
+				ImagePullPolicy: corev1.PullIfNotPresent,
+				VolumeMounts: []corev1.VolumeMount{
+					{
+						Name:      "licensing-config",
+						ReadOnly:  true,
+						MountPath: VGPULicensingConfigMountPath,
+						SubPath:   VGPULicensingFileName,
+					},
+				},
+			}).WithInitContainer(corev1.Container{
+				Name:            "k8s-driver-manager",
+				Image:           "nvcr.io/nvidia/cloud-native/k8s-driver-manager:v0.8.0",
+				ImagePullPolicy: corev1.PullIfNotPresent,
+			}).WithVolume(corev1.Volume{
+				Name: "licensing-config",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: "test-configmap",
+						},
+						Items: []corev1.KeyToPath{
+							{
+								Key:  VGPULicensingFileName,
+								Path: VGPULicensingFileName,
+							},
+						},
+					},
+				},
+			}),
+			errorExpected: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			err := TransformDriver(tc.ds.DaemonSet, tc.cpSpec,
+				ClusterPolicyController{client: tc.client, runtime: gpuv1.Containerd,
+					operatorNamespace: "test-ns", logger: ctrl.Log.WithName("test")})
+			if tc.errorExpected {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.EqualValues(t, tc.expectedDs, tc.ds)
+		})
+	}
 }
--- a/deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
+++ b/deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
@@ -685,11 +685,18 @@ spec:
                       licensing'
                     properties:
                       configMapName:
+                        description: 'Deprecated: ConfigMapName has been deprecated
+                          in favour of SecretName. Please use secrets to handle the
+                          LicensingConfig more securely'
                         type: string
                       nlsEnabled:
                         description: NLSEnabled indicates if NVIDIA Licensing System
                           is used for licensing.
                         type: boolean
+                      secretName:
+                        description: SecretName indicates the name of the secret containing
+                          the licensing token
+                        type: string
                     type: object
                   livenessProbe:
                     description: NVIDIA Driver container liveness probe settings
--- a/deployments/gpu-operator/crds/nvidia.com_nvidiadrivers.yaml
+++ b/deployments/gpu-operator/crds/nvidia.com_nvidiadrivers.yaml
@@ -234,11 +234,16 @@ spec:
                 description: 'Optional: Licensing configuration for NVIDIA vGPU licensing'
                 properties:
                   name:
+                    description: 'Deprecated: ConfigMapName has been deprecated in
+                      favour of SecretName. Please use secrets to handle the LicensingConfig
+                      more securely'
                     type: string
                   nlsEnabled:
                     description: NLSEnabled indicates if NVIDIA Licensing System is
                       used for licensing.
                     type: boolean
+                  secretName:
+                    type: string
                 type: object
               livenessProbe:
                 description: NVIDIA Driver container liveness probe settings
--- a/deployments/gpu-operator/templates/nvidiadriver.yaml
+++ b/deployments/gpu-operator/templates/nvidiadriver.yaml
@@ -48,7 +48,11 @@ spec:
   certConfig:
     name: {{ .Values.driver.certConfig.name }}
   {{- end }}
-  {{- if .Values.driver.licensingConfig.configMapName }}
+  {{- if .Values.driver.licensingConfig.secretName }}
+  licensingConfig:
+    secretName: {{ .Values.driver.licensingConfig.secretName }}
+    nlsEnabled: {{ .Values.driver.licensingConfig.nlsEnabled | default true }}
+  {{- else if .Values.driver.licensingConfig.configMapName }}
   licensingConfig:
     name: {{ .Values.driver.licensingConfig.configMapName }}
     nlsEnabled: {{ .Values.driver.licensingConfig.nlsEnabled | default true }}
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -206,7 +206,7 @@ driver:
     name: ""
   # vGPU licensing configuration
   licensingConfig:
-    configMapName: ""
+    secretName: ""
     nlsEnabled: true
   # vGPU topology daemon configuration
   virtualTopology:
--- a/internal/state/driver_volumes.go
+++ b/internal/state/driver_volumes.go
@@ -245,13 +245,23 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
 			additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, nlsTokenVolMount)
 		}
 
-		licensingConfigVolumeSource := corev1.VolumeSource{
-			ConfigMap: &corev1.ConfigMapVolumeSource{
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: cr.Spec.LicensingConfig.Name,
+		var licensingConfigVolumeSource corev1.VolumeSource
+		if cr.Spec.LicensingConfig.SecretName != "" {
+			licensingConfigVolumeSource = corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: cr.Spec.LicensingConfig.SecretName,
+					Items:      licenseItemsToInclude,
 				},
-				Items: licenseItemsToInclude,
-			},
+			}
+		} else {
+			licensingConfigVolumeSource = corev1.VolumeSource{
+				ConfigMap: &corev1.ConfigMapVolumeSource{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: cr.Spec.LicensingConfig.Name,
+					},
+					Items: licenseItemsToInclude,
+				},
+			}
 		}
 		licensingConfigVol := corev1.Volume{Name: "licensing-config", VolumeSource: licensingConfigVolumeSource}
 		additionalCfgs.Volumes = append(additionalCfgs.Volumes, licensingConfigVol)
--- a/scripts/gpu-operator-nvaie.sh
+++ b/scripts/gpu-operator-nvaie.sh
@@ -67,11 +67,11 @@ create_nls_config() {
         exit 1
     fi
 
-    # Return if the configmap is already created
-    kubectl get configmap licensing-config -n ${NAMESPACE} > /dev/null 2>&1 && echo "licensing-config already exists" && return 0
+    # Return if the secret is already created
+    kubectl get secret licensing-config -n ${NAMESPACE} > /dev/null 2>&1 && echo "licensing-config already exists" && return 0
 
-    # Create a configmap for vGPU licensing
-    touch gridd.conf && kubectl create configmap licensing-config \
+    # Create a secret for vGPU licensing
+    touch gridd.conf && kubectl create secret generic licensing-config \
         -n ${NAMESPACE} --from-file=gridd.conf --from-file=client_configuration_token.tok
 }
 
@@ -136,7 +136,7 @@ install_operator() {
           --set driver.image=${VGPU_DRIVER_NAME} \
           --set driver.version="${VGPU_DRIVER_VERSION}" \
           --set driver.imagePullSecrets={${REGISTRY_SECRET_NAME}} \
-          --set driver.licensingConfig.configMapName="licensing-config" \
+          --set driver.licensingConfig.secretName="licensing-config" \
           ${HELM_INSTALL_OPTS}
 
     # List all deployed pods in the namespace
@@ -146,7 +146,7 @@ install_operator() {
 cleanup_operator() {
     helm delete gpu-operator -n ${NAMESPACE}
     kubectl delete secret ${REGISTRY_SECRET_NAME} -n ${NAMESPACE}
-    kubectl delete configmap licensing-config -n ${NAMESPACE}
+    kubectl delete secret licensing-config -n ${NAMESPACE}
 }
 
 usage() {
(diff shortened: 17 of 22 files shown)


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1111,7 +1111,7 @@ type DriverLicensingConfigSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="ConfigMap Name"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
-	ConfigMapName string `json:"configMapName,omitempty"`
+	SecretName string `json:"secretName,omitempty"`
```
tariq1890, line 1135: We can't replace the fields just yet. This would be a breaking change as multiple users will be depending on this change. 

Let's introduce `SecretName` as an additional field and mark the `ConfigMapName` as Deprecated.  And moving forward, if users specify a secret, that should be prioritised over the  ConfigMap by the operator

cdesiniotis: +1

shivakunv: for testing purpose only, trying to resolve all the references first

shivakunv: I have included both the confimap and the secret to ensure the existing application continues to work. 
Next TODO: I'll add logic based on the chart version e.g. if version > 25.8.0 : it should use the secret instead of configmap . 

still testing the code ( currently containerd plugin-test is failing )

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -764,6 +764,11 @@ type DevicePluginSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Configuration for the NVIDIA Device Plugin via the ConfigMap"
 	Config *DevicePluginConfig `json:"config,omitempty"`
 
+	// Optional: Configuration for the NVIDIA Device Plugin via the Secret
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Configuration for the NVIDIA Device Plugin via the Secret"
+	Secret *DevicePluginSecret `json:"secret,omitempty"`
```
cdesiniotis: We do not need to inject any secrets into the device-plugin container. Please remove.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1113,6 +1134,11 @@ type DriverLicensingConfigSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="ConfigMap Name"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	ConfigMapName string `json:"configMapName,omitempty"`
+	// +kubebuilder:validation:Optional
```
cdesiniotis, line 1131:
```suggestion
	// SecretName indicates the name of the secret containing the licensing token
	// +kubebuilder:validation:Optional
```

api/nvidia/v1/clusterpolicy_types.go
cdesiniotis: Can you add the following comment on top of the ConfigMapName field (or some variation of this):
```
Deprecated: This field is deprecated in favor of the SecretName field. Use a secret for storing the licensing token instead of a ConfigMap.
```

controllers/object_controls.go
```diff
@@ -69,6 +69,8 @@ const (
 	DefaultCRIOConfigFile = "/etc/crio/crio.conf.d/99-nvidia.conf"
 	// TrustedCAConfigMapName indicates configmap with custom user CA injected
 	TrustedCAConfigMapName = "gpu-operator-trusted-ca"
+	// TrustedCAsecretName indicates secret with custom user CA injected
+	TrustedCAsecretName = "gpu-operator-trusted-ca"
```
cdesiniotis: This change appears unrelated to this PR.

controllers/object_controls.go
```diff
@@ -1142,6 +1144,72 @@ func applyOCPProxySpec(n ClusterPolicyController, podSpec *corev1.PodSpec) error
 	return nil
 }
 
+// applyOCPProxySpecSecret applies proxy settings to podSpec
+func applyOCPProxySpecSecret(n ClusterPolicyController, podSpec *corev1.PodSpec) error {
```
cdesiniotis: This change appears unrelated to this PR.

controllers/object_controls.go
```diff
@@ -1186,6 +1254,50 @@ func getOrCreateTrustedCAConfigMap(n ClusterPolicyController, name string) (*cor
 	return found, nil
 }
 
+// getOrCreateTrustedCA creates or returns an existing Trusted CA Bundle Secret.
+func getOrCreateTrustedCASecret(n ClusterPolicyController, name string) (*corev1.Secret, error) {
```
cdesiniotis: This change appears unrelated to this PR.

controllers/object_controls.go
```diff
@@ -2551,6 +2663,8 @@ func handleDevicePluginConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 	}
 	// setup volumes from configmap and shared emptyDir
 	obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, createConfigMapVolume(config.DevicePlugin.Config.Name, nil))
+	// setup volumes from secret, and shared emptyDir
+	obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, createSecretVolume(config.DevicePlugin.Secret.Name, nil))
```
cdesiniotis: This should be removed as we should not be adding anything to the device plugin.

controllers/object_controls.go
```diff
@@ -2706,6 +2820,12 @@ func transformPeerMemoryContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
 				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for kernel module configuration: %v", err)
 			}
 			obj.Spec.Template.Spec.Containers[i].VolumeMounts = append(obj.Spec.Template.Spec.Containers[i].VolumeMounts, volumeMounts...)
+
+			volumeMounts, _, err = createSecretVolumeMounts(n, config.Driver.KernelModuleConfig.Name, destinationDir)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for kernel module configuration: %v", err)
+			}
+			obj.Spec.Template.Spec.Containers[i].VolumeMounts = append(obj.Spec.Template.Spec.Containers[i].VolumeMounts, volumeMounts...)
```
cdesiniotis: This should be removed as no modifications should be made to the `transformPeerMemoryContainer` function.

controllers/object_controls.go
```diff
@@ -2779,6 +2899,12 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
 			}
 			gdsContainer.VolumeMounts = append(gdsContainer.VolumeMounts, volumeMounts...)
+
+			volumeMounts, _, err = createSecretVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
+			}
+			gdsContainer.VolumeMounts = append(gdsContainer.VolumeMounts, volumeMounts...)
```
cdesiniotis: This should be removed as no modifications should be made to the GDS container.

controllers/object_controls.go
```diff
@@ -2858,6 +2984,12 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
 			}
 			gdrcopyContainer.VolumeMounts = append(gdrcopyContainer.VolumeMounts, volumeMounts...)
+
+			volumeMounts, _, err = createSecretVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
+			}
+			gdrcopyContainer.VolumeMounts = append(gdrcopyContainer.VolumeMounts, volumeMounts...)
```
cdesiniotis: This should be removed as no modifications should be made to the GDRCopy container.

controllers/object_controls.go
```diff
@@ -3331,6 +3536,13 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 		}
 		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
 		podSpec.Volumes = append(podSpec.Volumes, createConfigMapVolume(config.Driver.KernelModuleConfig.Name, itemsToInclude))
+
+		volumeMounts, itemsToInclude, err = createSecretVolumeMounts(n, config.Driver.KernelModuleConfig.Name, destinationDir)
+		if err != nil {
+			return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for kernel module configuration: %v", err)
+		}
+		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
+		podSpec.Volumes = append(podSpec.Volumes, createSecretVolume(config.Driver.KernelModuleConfig.Name, itemsToInclude))
```
cdesiniotis: I am not following this. The kernel module configuration file is a ConfigMap, so this code should be removed.

controllers/object_controls.go
```diff
@@ -3370,6 +3582,12 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 		}
 		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
 		podSpec.Volumes = append(podSpec.Volumes, createConfigMapVolume(config.Driver.CertConfig.Name, itemsToInclude))
+		volumeMounts, itemsToInclude, err = createSecretVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
+		if err != nil {
+			return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %v", err)
+		}
+		driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
+		podSpec.Volumes = append(podSpec.Volumes, createSecretVolume(config.Driver.CertConfig.Name, itemsToInclude))
```
cdesiniotis: I am not following why this change was made in this PR. It is out of scope.

deployments/gpu-operator/templates/nvidiadriver.yaml
```diff
@@ -53,6 +53,11 @@ spec:
     name: {{ .Values.driver.licensingConfig.configMapName }}
     nlsEnabled: {{ .Values.driver.licensingConfig.nlsEnabled | default true }}
   {{- end }}
+  {{- if .Values.driver.licensingConfig.secretName }}
+  licensingConfig:
+    secretName: {{ .Values.driver.licensingConfig.secretName }}
+    nlsEnabled: {{ .Values.driver.licensingConfig.nlsEnabled | default true }}
+  {{- end }}
```
cdesiniotis, line 59: There is already a licensingConfig field defined right above this. We need to add a `secretName` field to that struct. Additionally, we need to make the same change to the clusterpolicy template.

Also note, this is the NVIDIAdriver template your are updating. You have not updated the NVIDIADriver CRD in this PR as of yet. You will also need to add the `secretName` field to that CRD as well.

scripts/gpu-operator-nvaie.sh
```diff
@@ -137,6 +137,7 @@ install_operator() {
           --set driver.version="${VGPU_DRIVER_VERSION}" \
           --set driver.imagePullSecrets={${REGISTRY_SECRET_NAME}} \
           --set driver.licensingConfig.configMapName="licensing-config" \
+          --set driver.licensingConfig.secretName="licensing-config" \
```
cdesiniotis, line 139: Let's remove the previous line which sets the `configMapName` field. The intent of this PR is to deprecate that field so we should not be using it in this script.

controllers/object_controls.go
```diff
@@ -3267,6 +3442,36 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 	}
 
 	// set any licensing configuration required
+	if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.SecretName != "" {
```
cdesiniotis: Note, if both SecretName and ConfigMapName are specified, we should only use the secret. We should update the below check on line 3475 to be:
```
config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.ConfigMapName != "" && config.Driver.LicensingConfig.SecretName == "" {
```

controllers/object_controls.go
tariq1890: yes, let's remove unrelated changes

controllers/object_controls.go
shivakunv: `if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.SecretName != ""{
}
else if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.ConfigMapName != "" {
}
`

if secretName !=null : use secretName

deployments/gpu-operator/templates/nvidiadriver.yaml
```diff
@@ -48,7 +48,11 @@ spec:
   certConfig:
     name: {{ .Values.driver.certConfig.name }}
   {{- end }}
-  {{- if .Values.driver.licensingConfig.configMapName }}
+  {{- if .Values.driver.licensingConfig.secretName }}
+  licensingConfig:
+    name: {{ .Values.driver.licensingConfig.secretName }}
+    nlsEnabled: {{ .Values.driver.licensingConfig.nlsEnabled | default true }}
+  {{- else if .Values.driver.licensingConfig.configMapName }}
```
shivakunv, line 55: same 
`name:`

is used in crd

controllers/object_controls.go
```diff
@@ -3267,7 +3267,37 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
...
+
+		licensingConfigVolumeSource := corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: config.Driver.LicensingConfig.SecretName,
+				Items:      licenseItemsToInclude,
+			},
+		}
+		licensingConfigVol := corev1.Volume{Name: "licensing-config", VolumeSource: licensingConfigVolumeSource}
+		podSpec.Volumes = append(podSpec.Volumes, licensingConfigVol)
+	} else if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.ConfigMapName != "" {
```
tariq1890: @shivakunv I saw the lint failures, and I think it's fine to use `// nolint` here

tariq1890: Please add `//nolint` to resolve the lint error in L3324 as well

controllers/object_controls.go
```diff
@@ -3267,7 +3267,36 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
...
```
tariq1890: We also need to think of enabling a migration path from configMap to secretMap. 

Consider this scenario:

i) User has both configMapName and secretName set
ii) The driver container already has mounted a configMap

Expected Behaviour: The ClusterPolicy and NVIDIADriver reconcilers have to delete the configMap reference in favour of the new secret reference. @cdesiniotis Can you let me know if I am missing anything here?

Please make sure you cover this scenario when you test your PR

tariq1890:
@shivakunv  Thanks for working on this. Please add unit tests so that coveralls check passes as well

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1108,11 +1108,19 @@ type DriverCertConfigSpec struct {
...
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="ConfigMap Name"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
 	ConfigMapName string `json:"configMapName,omitempty"`
+	// SecretName indicates the name of the secret containing the licensing token
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Secret Name"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:text"
+	SecretName string `json:"secretName,omitempty"`
```
tariq1890, line 1135: We have to make a similar change in the `DriverLicensingConfigSpec` struct of  "nvidiadriver_types.go"

controllers/object_controls.go
shivakunv: @cdesiniotis Please check @tariq1890 comments

tariq1890: Can you try this out for the time being? Implement the suggested change without committing them for now

shivakunv: skipped already mounted volume in append and mounted again new volume

controllers/object_controls.go
```diff
@@ -3283,7 +3283,47 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 	}
 
 	// set any licensing configuration required
-	if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.ConfigMapName != "" {
+	if config.Driver.LicensingConfig != nil && config.Driver.LicensingConfig.SecretName != "" {
```
cdesiniotis: This transformation code is quite long. Let's move this entire code block to its's own function. It will be easier to write unit tests for it that way. E.g. `applyLicensingConfig()`. We can add unit tests for this new function in `transforms_test.go`

controllers/object_controls.go
cdesiniotis: Are you saying your code already does this? Have you tested it? I believe having unit tests for the various scenarios would help exhibit how your code behaves and whether it properly address this scenario.

shivakunv: need to test this scenario

scripts/gpu-operator-nvaie.sh
```diff
@@ -136,7 +136,7 @@ install_operator() {
           --set driver.image=${VGPU_DRIVER_NAME} \
           --set driver.version="${VGPU_DRIVER_VERSION}" \
           --set driver.imagePullSecrets={${REGISTRY_SECRET_NAME}} \
-          --set driver.licensingConfig.configMapName="licensing-config" \
```
shivakunv, line 139: Please check need of both configMapName and secret in this file @tariq1890 @cdesiniotis

controllers/object_controls.go
```diff
@@ -3180,6 +3197,39 @@ func createConfigMapVolumeMounts(n ClusterPolicyController, configMapName string
...
+// in the ConfigMap. Use subPath to ensure original contents
+// at destinationDir are not overwritten.
+func createSecretVolumeMounts(n ClusterPolicyController, secretName string, destinationDir string) ([]corev1.VolumeMount, []corev1.KeyToPath, error) {
+	ctx := n.ctx
+	// get the Secret
+	secret := &corev1.Secret{}
+	opts := client.ObjectKey{Namespace: n.operatorNamespace, Name: secretName}
+	err := n.client.Get(ctx, opts, secret)
+	if err != nil {
+		return nil, nil, fmt.Errorf("ERROR: could not get Secret %s from client: %v", secretName, err)
```
```suggestion
		return nil, nil, fmt.Errorf("ERROR: could not get Secret %s from client: %w", secretName, err)
```

controllers/object_controls.go
```diff
@@ -3380,12 +3477,23 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
...
 			return fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %v", err)
 		}
-		volumeMounts, itemsToInclude, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
-		if err != nil {
-			return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %v", err)
+
+		if config.Driver.LicensingConfig.SecretName != "" {
+			volumeMounts, itemsToInclude, err := createSecretVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %v", err)
```
```suggestion
				return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
```

controllers/object_controls.go
```diff
@@ -3380,12 +3477,23 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
...
+			}
+			driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
+			podSpec.Volumes = append(podSpec.Volumes, createSecretVolume(config.Driver.CertConfig.Name, itemsToInclude))
+		} else {
+			volumeMounts, itemsToInclude, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %v", err)
```
```suggestion
				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
```

controllers/transforms_test.go
```diff
@@ -145,6 +145,11 @@ func (d Daemonset) WithRuntimeClassName(name string) Daemonset {
 	return d
 }
 
+func (d Daemonset) WithVolume(volume corev1.Volume) Daemonset {
```
tariq1890, line 159: This method is unnecessary. Please remove

internal/state/driver_volumes.go
```diff
@@ -245,13 +256,23 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
 			additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, nlsTokenVolMount)
 		}
 
-		licensingConfigVolumeSource := corev1.VolumeSource{
-			ConfigMap: &corev1.ConfigMapVolumeSource{
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: cr.Spec.LicensingConfig.Name,
+		var licensingConfigVolumeSource corev1.VolumeSource
+		if cr.Spec.IsVGPULicensingEnabledWithSecret() {
```
tariq1890:
```suggestion
		if len(cr.Spec.LicensingConfig.SecretName) > 0 {
```

internal/state/driver_volumes.go
```diff
@@ -222,7 +233,7 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
 	}
 
 	// set any licensing configuration required
-	if cr.Spec.IsVGPULicensingEnabled() {
+	if cr.Spec.IsVGPULicensingEnabled() || cr.Spec.IsVGPULicensingEnabledWithSecret() {
```
tariq1890:
```suggestion
	if cr.Spec.IsVGPULicensingEnabled() {
```

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -691,6 +699,14 @@ func (d *NVIDIADriverSpec) IsVGPULicensingEnabled() bool {
 	return d.LicensingConfig.Name != ""
 }
 
+// IsVGPULicensingEnabledWithSecret returns true if the vgpu driver license config is provided with secret
+func (d *NVIDIADriverSpec) IsVGPULicensingEnabledWithSecret() bool {
```
tariq1890: Instead of creating this new method, change L699 to 

```
return d.LicensingConfig.Name != "" || d.LicensingConfig.SecretName != ""
```

controllers/transforms_test.go
shivakunv: using in TestTransformDriverWithLicensingConfig

shivakunv:
incorporated all the review comments @tariq1890 PTAL

controllers/object_controls.go
```diff
@@ -2774,11 +2774,20 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 			if err != nil {
 				return fmt.Errorf("ERROR: failed to get destination directory for ssl key/cert config: %w", err)
 			}
-			volumeMounts, _, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
-			if err != nil {
-				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
+
+			if config.Driver.LicensingConfig.SecretName != "" {
```
tariq1890: This change does not seem correct. You are adding this code in an area that does not use the NLS license. This is in the `transformGDS` method

controllers/object_controls.go
```diff
@@ -2861,11 +2870,19 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 			if err != nil {
 				return fmt.Errorf("ERROR: failed to get destination directory for ssl key/cert config: %w", err)
 			}
-			volumeMounts, _, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
-			if err != nil {
-				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
+			if config.Driver.LicensingConfig.SecretName != "" {
```
tariq1890: same comment as the previous one. This does not seem right to me.

controllers/object_controls.go
```diff
@@ -3180,6 +3197,39 @@ func createConfigMapVolumeMounts(n ClusterPolicyController, configMapName string
 	return volumeMounts, itemsToInclude, nil
 }
 
+// createSecretVolumeMounts creates a VolumeMount for each key
+// in the ConfigMap. Use subPath to ensure original contents
```
```suggestion
// in the Secret. Use subPath to ensure original contents
```

controllers/object_controls.go
```diff
@@ -3201,6 +3261,74 @@ func createEmptyDirVolume(volumeName string) corev1.Volume {
 	}
 }
 
+func applyLicensingConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, driverIndex int) {
+	podSpec := &obj.Spec.Template.Spec
```
tariq1890, line 3270: there is a lot of duplication in this method. Can we address that?

controllers/object_controls.go
```diff
@@ -3380,12 +3477,23 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 		if err != nil {
 			return fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %v", err)
 		}
-		volumeMounts, itemsToInclude, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
-		if err != nil {
-			return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %v", err)
+
+		if config.Driver.LicensingConfig.SecretName != "" {
+			volumeMounts, itemsToInclude, err := createSecretVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
```
tariq1890: Why does the creation of volume mounts from `Driver.CertConfig` depend on this ?

```
config.Driver.LicensingConfig.SecretName != ""
```

controllers/object_controls.go
shivakunv: Thanks . removed

controllers/object_controls.go
shivakunv: thanks . removed

controllers/object_controls.go
shivakunv: Thanks . removed duplicatation

shivakunv: Thanks . modified

controllers/object_controls.go
shivakunv: tested . the pod will restart, the old configmap volume will be cleared and a new secrets volume will be created .

controllers/object_controls.go
shivakunv: No . it does not depend on this . 
I am not able to find condition( for secret) for custom ssl certificate configuration , that is why I added this if condition . 

Quest: 
Shall we delete the configmap and use only secrets for the volume mounts ?

replace below code

```
volumeMounts, itemsToInclude, err := createConfigMapVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
			if err != nil {
				return fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
			}
			driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
			podSpec.Volumes = append(podSpec.Volumes, createConfigMapVolume(config.Driver.CertConfig.Name, itemsToInclude))
```
			
with: 

```
volumeMounts, itemsToInclude, err := createSecretVolumeMounts(n, config.Driver.CertConfig.Name, destinationDir)
			if err != nil {
				return fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
			}
			driverContainer.VolumeMounts = append(driverContainer.VolumeMounts, volumeMounts...)
			podSpec.Volumes = append(podSpec.Volumes, createSecretVolume(config.Driver.CertConfig.Name, itemsToInclude))
```

without if condition similar to configmap

controllers/object_controls.go
```diff
@@ -3372,38 +3465,7 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
...
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: config.Driver.LicensingConfig.ConfigMapName,
-				},
-				Items: licenseItemsToInclude,
-			},
-		}
-		licensingConfigVol := corev1.Volume{Name: "licensing-config", VolumeSource: licensingConfigVolumeSource}
-		podSpec.Volumes = append(podSpec.Volumes, licensingConfigVol)
-	}
+	applyLicensingConfig(obj, config, driverIndex)
```
```suggestion
    if config.Driver.LicensingConfig.IsVGPULicensingEnabled() {
	     applyLicensingConfig(obj, config, driverIndex)
	}
```

controllers/object_controls.go
```diff
@@ -3290,6 +3333,56 @@ func createEmptyDirVolume(volumeName string) corev1.Volume {
 	}
 }
 
+func applyLicensingConfig(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, driverIndex int) {
+	if config.Driver.LicensingConfig == nil || (config.Driver.LicensingConfig.SecretName == "" && config.Driver.LicensingConfig.ConfigMapName == "") {
```
tariq1890: Let's remove this if condition from here and move it outside this method. We should only call this method if that `if conditional` evaluates to true

controllers/object_controls.go
tariq1890: i don't see any reason for any changes to be made in this area of code. `Driver.CertConfig` does not have anything to do with VGPU Licensing config. Let's revert changes to any unrelated code

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
 				return nil, fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %w", err)
 			}
-			volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
-				cr.Spec.CertConfig.Name, destinationDir)
-			if err != nil {
-				return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
+
+			if cr.Spec.LicensingConfig.SecretName != "" {
+				volumeMounts, itemsToInclude, err := s.createSecretVolumeMounts(ctx, s.namespace,
+					cr.Spec.CertConfig.Name, destinationDir)
```
tariq1890: I don't think we should be referencing `cr.Spec.CertConfig` here
```suggestion
					cr.Spec.LicensingConfig.SecretName, destinationDir)
```

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
+				if err != nil {
+					return nil, fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
+				}
+				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
+				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createSecretVolume(cr.Spec.CertConfig.Name, itemsToInclude))
```
tariq1890: Same here
```suggestion
				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createSecretVolume(cr.Spec.LicensingConfig.SecretName, itemsToInclude))
```

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
+			} else {
+				volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
+					cr.Spec.CertConfig.Name, destinationDir)
```
tariq1890: Same here
```suggestion
					cr.Spec.LicensingConfig.Name, destinationDir)
```

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
+				if err != nil {
+					return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
+				}
+				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
+				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createConfigMapVolume(cr.Spec.CertConfig.Name, itemsToInclude))
```
tariq1890: same here
```suggestion
				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createConfigMapVolume(cr.Spec.LicensingConfig.Name, itemsToInclude))
```

tariq1890 (changes requested):

tariq1890:
@shivakunv Thanks for the work on this so far. Please add a unit test in `internal/state/driver_test.go` to cover the vGPU Licensing config scenarios as well

internal/state/driver_volumes.go
shivakunv: okay. Changed it for configMap as well

shivakunv: should not it be :  cr.Spec.LicensingConfig.**configMapName**  similar to secretName @tariq1890 

additionalCfgs.Volumes = append(additionalCfgs.Volumes, createConfigMapVolume(cr.Spec.LicensingConfig.Name, itemsToInclude))

shivakunv:
added UT . PTAL @tariq1890

internal/state/driver_volumes.go
shivakunv: // set any custom ssl key/certificate configuration provided
if cr.Spec.IsCertConfigEnabled() { 

we are passing cert name for custorm certificate configuration . 
Please check again @tariq1890

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1113,12 +1113,21 @@ type DriverCertConfigSpec struct {
 
 // DriverLicensingConfigSpec defines licensing server configuration for NVIDIA Driver container
 type DriverLicensingConfigSpec struct {
+	// nolint
+	// Deprecated: Use SecretName instead. for better security
```
```suggestion
	// Deprecated: ConfigMapName has been deprecated in favour of SecretName. Please use secrets to handle the LicensingConfig more securely.
```

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
+
+			if cr.Spec.LicensingConfig.SecretName != "" {
+				volumeMounts, itemsToInclude, err := s.createSecretVolumeMounts(ctx, s.namespace,
+					cr.Spec.LicensingConfig.SecretName, destinationDir)
+				if err != nil {
+					return nil, fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
+				}
+				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
+				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createSecretVolume(cr.Spec.LicensingConfig.SecretName, itemsToInclude))
+			} else {
```
tariq1890:
```suggestion
			} else if cr.Spec.LicensingConfig.ConfigMapName != "" {
```

tariq1890:
Please rebase the PR

internal/state/driver_volumes.go
shivakunv: LicensingConfig *DriverLicensingConfigSpec  does not have ConfigMapName 

modified for : 
`} else if cr.Spec.LicensingConfig.Name != "" {`

shivakunv:
PTAL @tariq1890

bundle/manifests/nvidia.com_nvidiadrivers.yaml
cdesiniotis, line 1: This file looks out-of-sync with `config/crd/bases/nvidia.com_nvidiadrivers.yaml`. Please make sure to copy the latest version of the CRD to both the helm chart and OLM bundle.

bundle/manifests/nvidia.com_clusterpolicies.yaml
cdesiniotis, line 1: This file looks out-of-sync with config/crd/bases/nvidia.com_clusterpolicies.yaml. Please make sure to copy the latest version of the CRD to both the helm chart and OLM bundle.

config/crd/bases/nvidia.com_clusterpolicies.yaml
```diff
@@ -685,11 +685,18 @@ spec:
                       licensing'
                     properties:
                       configMapName:
+                        description: |-
+                          nolint
```
cdesiniotis: This should not be present in the generated CRD.

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
```diff
@@ -26,6 +26,7 @@ spec:
     name: ""
   # vGPU licensing configuration
   licensingConfig:
+    secretName: ""
```
cdesiniotis, line 29: Let's add a comment on top of the `name` field indicating that it is deprecated and that users should use `secretName` instead.

config/samples/v1_clusterpolicy.yaml
```diff
@@ -91,6 +91,7 @@ spec:
     # vGPU licensing configuration
     licensingConfig:
       configMapName: ""
+      secretName: ""
```
cdesiniotis, line 93: Let's add a comment on top of the `configMapName` field indicating that it is deprecated and that users should use secretName instead.

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
cdesiniotis: Or actually, should we just remove the `name` field entirely from the sample CR? @tariq1890

config/samples/v1_clusterpolicy.yaml
cdesiniotis: Or actually, should we just remove the `configMapName` field entirely from the sample CR? @tariq1890

controllers/object_controls.go
```diff
@@ -3367,37 +3414,8 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
...
-				LocalObjectReference: corev1.LocalObjectReference{
-					Name: config.Driver.LicensingConfig.ConfigMapName,
-				},
-				Items: licenseItemsToInclude,
-			},
-		}
-		licensingConfigVol := corev1.Volume{Name: "licensing-config", VolumeSource: licensingConfigVolumeSource}
-		podSpec.Volumes = append(podSpec.Volumes, licensingConfigVol)
+	if config.Driver.IsVGPULicensingEnabled() {
+		applyLicensingConfig(obj, config, driverIndex)
```
cdesiniotis: A relevant change has been merged into main: https://github.com/NVIDIA/gpu-operator/pull/1740. We now use a reference to the container throughout this function instead of an index: https://github.com/NVIDIA/gpu-operator/blob/4201e397460b3ae9ceee6e9116a2075722c89992/controllers/object_controls.go#L3265

Please rebase and pass a reference to the Container instead of an index to `applyLicensingConfig`.

controllers/transforms_test.go
```diff
@@ -2520,3 +2525,150 @@ func TestGetRuntimeConfigFiles(t *testing.T) {
...
+					},
+					LicensingConfig: &gpuv1.DriverLicensingConfigSpec{
+						SecretName: "test-secret",
+					},
+				},
+			},
+			client: mockClient,
+			expectedDs: NewDaemonset().WithContainer(corev1.Container{
+				Name:            "nvidia-driver-ctr",
+				Image:           "nvcr.io/nvidia/driver:570.172.08-",
```
cdesiniotis, line 2688: Not a blocker -- Is it possible to add the relevant NFD labels to the Node object so that the OS version suffix gets appended to the image path here? Taking a look at the code that constructs the driver image tag https://github.com/NVIDIA/gpu-operator/blob/4201e397460b3ae9ceee6e9116a2075722c89992/controllers/object_controls.go#L700-L710 I believe we need to add `nfdOSReleaseIDLabelKey` and `nfdOSVersionIDLabelKey` labels to the mock Node object.

deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
cdesiniotis, line 1: Same comment as above. Make sure this CRD is up-to-date.

deployments/gpu-operator/crds/nvidia.com_nvidiadrivers.yaml
cdesiniotis, line 1: Same comment as above. Make sure this CRD is up-to-date.

deployments/gpu-operator/templates/nvidiadriver.yaml
```diff
@@ -48,7 +48,11 @@ spec:
   certConfig:
     name: {{ .Values.driver.certConfig.name }}
   {{- end }}
-  {{- if .Values.driver.licensingConfig.configMapName }}
+  {{- if .Values.driver.licensingConfig.secretName }}
+  licensingConfig:
+    name: {{ .Values.driver.licensingConfig.secretName }}
```
cdesiniotis: Shouldn't the field be `secretName`?
```suggestion
    secretName: {{ .Values.driver.licensingConfig.secretName }}
```

deployments/gpu-operator/values.yaml
```diff
@@ -207,6 +207,7 @@ driver:
   # vGPU licensing configuration
   licensingConfig:
     configMapName: ""
+    secretName: ""
```
cdesiniotis, line 209: Let's add a comment indicating that the `configMapName` is deprecated in favor of `secretName`. 

Also, @tariq1890 should we remove the deprecated field from the helm values?

scripts/gpu-operator-nvaie.sh
cdesiniotis: We should just use a secret going forward.

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
tariq1890: Removing it makes sense

config/samples/v1_clusterpolicy.yaml
tariq1890: Yes, agree with removing it

controllers/transforms_test.go
tariq1890: The `parseOSRelease` method isn't UT friendly unfortunately. We need to come up with a better long-term solutions and update all the unit tests affected by this

deployments/gpu-operator/templates/nvidiadriver.yaml
tariq1890: good catch

controllers/transforms_test.go
cdesiniotis: Ack. Let's keep the unit tests as is then in this PR.

bundle/manifests/nvidia.com_nvidiadrivers.yaml
tariq1890: Done.

bundle/manifests/nvidia.com_clusterpolicies.yaml
tariq1890: Done

config/crd/bases/nvidia.com_clusterpolicies.yaml
tariq1890: Done

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
tariq1890: Done. Please check

config/samples/v1_clusterpolicy.yaml
tariq1890: Removed. Please check

deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
tariq1890: Done

deployments/gpu-operator/templates/nvidiadriver.yaml
tariq1890: Good catch. I've fixed this. Please check

deployments/gpu-operator/values.yaml
tariq1890: Yes. I've removed the `configMapName` from the `values.yaml` and from the sample clusterpolicy yaml

.golangci.yml
```diff
@@ -23,6 +23,15 @@ linters:
       - third_party$
       - builtin$
       - examples$
+    rules:
+      - linters:
+          - staticcheck
+        path: (.+)\.go$
+        text: 'SA1019: config.Driver.LicensingConfig.ConfigMapName is deprecated(.+)'
```
rajathagasthya, line 30: **non-blocking:** Is there a way to make this exclusion a bit more generic instead of one per deprecated field? Like using wildcard matching and restricting the path to the API types file(s). Something like this (note: it's untested):

```
rules:
  - linters:
      - staticcheck
    path: 'path/to/specific/file\.go$'
    text: 'SA1019:'
```

tariq1890: I am sure there is. I just need to come up with a better regex

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
cdesiniotis: I don't see this being addressed.

controllers/object_controls.go
cdesiniotis: Thanks @tariq1890. I see this has been addressed now.

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
tariq1890: You're right. I missed this one

.golangci.yml
tariq1890: Basically there are two files where these `Deprecated` fields are referenced

1. `controllers/object_controls.go`
2. `internal/state/driver_volumes.go`

tariq1890: I actually prefer filtering out the specific lint error as opposed to specifying file paths. There are other lint checks that need to pass in those files

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
tariq1890: Can you check now?

cdesiniotis: I see it now. Thanks.

shivakunv:
@tariq1890 
```
Merging is blocked
1 review requesting changes by reviewers with write access. 
```
Please check .

shivakunv:
tested the code , It's working fine @tariq1890

.golangci.yml
rajathagasthya: That makes total sense, thanks!

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -72,7 +72,8 @@ metadata:
               },
               "licensingConfig": {
                 "nlsEnabled": true,
-                "configMapName": ""
+                "configMapName": "",
```
cdesiniotis: Should we remove the deprecated field from the example ClusterPolicy instance?

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -170,7 +171,8 @@ metadata:
             },
             "licensingConfig": {
               "nlsEnabled": true,
-              "name": ""
+              "name": "",
```
cdesiniotis: Should we remove the deprecated field from the example ClusterPolicy instance?

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
-			volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
-				cr.Spec.CertConfig.Name, destinationDir)
-			if err != nil {
-				return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
+
+			if cr.Spec.LicensingConfig.SecretName != "" {
+				volumeMounts, itemsToInclude, err := s.createSecretVolumeMounts(ctx, s.namespace,
+					cr.Spec.LicensingConfig.SecretName, destinationDir)
+				if err != nil {
+					return nil, fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
```
cdesiniotis:
```suggestion
					return nil, fmt.Errorf("ERROR: failed to create Secret VolumeMounts for licensing configuration: %w", err)
```

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
+				}
+				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
+				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createSecretVolume(cr.Spec.LicensingConfig.SecretName, itemsToInclude))
+			} else if cr.Spec.LicensingConfig.Name != "" {
+				volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
+					cr.Spec.LicensingConfig.Name, destinationDir)
+				if err != nil {
+					return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
```
cdesiniotis:
```suggestion
					return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for licensing configuration: %w", err)
```

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
 			if err != nil {
 				return nil, fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %w", err)
 			}
-			volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
-				cr.Spec.CertConfig.Name, destinationDir)
-			if err != nil {
-				return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
```
cdesiniotis, line 165: Was this deleted by accident?

internal/state/driver_volumes.go
```diff
@@ -159,13 +159,24 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
...
+
+			if cr.Spec.LicensingConfig.SecretName != "" {
```
cdesiniotis: This code block appears to be in the wrong scope, no? It appears to be under `if cr.Spec.IsCertConfigEnabled() { ... }` when it should not be.

cdesiniotis: I see there is a separate code block later on in this function for `cr.Spec.IsVGPULicensingEnabled() { .. }` -- all of the necessary logic should be contained to that block, right?

tariq1890: You're right. Good catch

internal/state/driver_volumes.go
tariq1890: checking

shivakunv: no : it's under condition  Line-171 ,  (  **Non-Precompiled** `if !cr.Spec.UsePrecompiledDrivers() {`) : 
```
} else if cr.Spec.LicensingConfig.Name != "" {
				volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
					cr.Spec.LicensingConfig.Name, destinationDir)
				if err != nil {
					return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
				}
				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createConfigMapVolume(cr.Spec.LicensingConfig.Name, itemsToInclude))
			}
```
original block 
```
// set any custom ssl key/certificate configuration provided
		if cr.Spec.IsCertConfigEnabled() {
			destinationDir, err := getCertConfigPath(pool.osRelease)
			if err != nil {
				return nil, fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %w", err)
			}
			volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
				cr.Spec.CertConfig.Name, destinationDir)
			if err != nil {
				return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
			}
			additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
			additionalCfgs.Volumes = append(additionalCfgs.Volumes, createConfigMapVolume(cr.Spec.CertConfig.Name, itemsToInclude))
		}
```

internal/state/driver_volumes.go
shivakunv: No , both have different logic, . The logic 
`if cr.Spec.IsCertConfigEnabled() {`
under non precompiled 
`if !cr.Spec.UsePrecompiledDrivers() {`

condition creates a separate volume for custom certificates  
`cr.Spec.CertConfig.Name`

_createSecretVolumeMounts , createConfigMapVolumeMounts_

```
if cr.Spec.LicensingConfig.SecretName != "" {
				volumeMounts, itemsToInclude, err := s.createSecretVolumeMounts(ctx, s.namespace,
					cr.Spec.LicensingConfig.SecretName, destinationDir)
				if err != nil {
					return nil, fmt.Errorf("ERROR: failed to create Secret VolumeMounts for custom certs: %w", err)
				}
				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createSecretVolume(cr.Spec.LicensingConfig.SecretName, itemsToInclude))
			} else if cr.Spec.LicensingConfig.Name != "" {
				volumeMounts, itemsToInclude, err := s.createConfigMapVolumeMounts(ctx, s.namespace,
					cr.Spec.LicensingConfig.Name, destinationDir)
				if err != nil {
					return nil, fmt.Errorf("ERROR: failed to create ConfigMap VolumeMounts for custom certs: %w", err)
				}
				additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, volumeMounts...)
				additionalCfgs.Volumes = append(additionalCfgs.Volumes, createConfigMapVolume(cr.Spec.LicensingConfig.Name, itemsToInclude))
			}
```

In the configMapName logic, we are using `cr.Spec.CertConfig.Name`  and in secretName , we are using `secretName` . For consistency, we should use the same `cr.Spec.CertConfig.Name ` . 

_Could you please clarify whether it's pkay to remove the entire code block for both configMapName and secretName_  ?

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1120,12 +1120,20 @@ type DriverCertConfigSpec struct {
 
 // DriverLicensingConfigSpec defines licensing server configuration for NVIDIA Driver container
 type DriverLicensingConfigSpec struct {
+	// Deprecated: ConfigMapName has been deprecated in favour of SecretName. Please use secrets to handle the LicensingConfig more securely
```
tariq1890: I know I had suggested this before. But on giving it more thought, the API docs should be in plain english as opposed to using pascal case like `LicensingConfig`
```suggestion
	// Deprecated: ConfigMapName has been deprecated in favour of SecretName. Please use secrets to handle the licensing server configuration more securely
```

internal/state/driver_volumes.go
tariq1890: Shiva, the code blocks concerning `CertConfig` should not be touched in this PR at all. 

If we want to transition to managing `CertConfig` as a secret, that has to be done in a separate PR. Let us just focus on `LicensingConfig` in this PR

api/nvidia/v1/clusterpolicy_types.go
shivakunv: done

tariq1890: thanks

internal/state/driver_volumes.go
shivakunv: Thanks. removed

cdesiniotis (approved):

PR #1545, opened 2025-07-20
[dcgm-exporter] Add option hostNetwork for dcgmExporter

Alja9:
Is it possible to add the `hostnetwork` option to the `dcgm-exporter` via `gpu-operator` ? 
because the `dcgm-exporter` itself already provides that option :
https://github.com/NVIDIA/dcgm-exporter/blob/8fd69339fe2138459d3618508c3ebb1f43b51f79/deployment/templates/daemonset.yaml#L49-L50

Why is this necessary? because the label on the `dcgm-exporter` metrics does not have a node label to display the node name, while the label provided by the `dcgm-exporter` that is possible to display the node name is only `hostname`.

If `hostnetwork: false` then in the value `hostname` label will use the pod name, 
but if `hostnetwork: true` then in the value `hostname` label will use the node name.

And this is useful for monitoring tools that require access to node-level network interfaces or need to expose ports directly on the node.

### DIFF
--- a/bundle/manifests/nvidia.com_clusterpolicies.yaml
+++ b/bundle/manifests/nvidia.com_clusterpolicies.yaml
@@ -337,6 +337,9 @@ spec:
                       - name
                       type: object
                     type: array
+                  hostNetwork:
+                    description: Run the DCGM Exporter pod on the host’s network, sharing the node’s network interfaces and IP address
+                    type: boolean
                   image:
                     description: NVIDIA DCGM Exporter image name
                     pattern: '[a-zA-Z0-9\-]+'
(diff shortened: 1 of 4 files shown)


tariq1890:
@Alja9 Thank you for creating this PR and we apologise for missing this earlier. It looks like this PR https://github.com/NVIDIA/gpu-operator/pull/1962 has already been merged with the same changes.

We appreciate you taking the time to contribute to gpu-operator!

Alja9:
Hi @tariq1890 

Thanks for check this PR and for letting me know, I appreciate it !

PR #1552, opened 2025-07-23
Bump version to 25.3.2

cdesiniotis:


versions.mk
```diff
@@ -17,7 +17,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= v25.3.1
+VERSION ?= v25.3.2
 
 GOLANG_VERSION ?= 1.24.4
```
tariq1890: Thoughts on updating the go version too?

tariq1890: Can be a separate PR too. Your call

cdesiniotis: Added a separate commit to this PR

ArangoGutierrez (approved):

PR #1554, opened 2025-07-24
Add pod labels and annotations for DCGM Exporter

maciej-tatarski:


rahulait:
Thank you for this PR @maciej-tatarski. We have added support for annotations in https://github.com/NVIDIA/gpu-operator/pull/2292

Fix will be available in release v26.7.0. We don't have plans to add custom pod labels. But if there is a usecase for which it should be evaluated, please feel free to create a github issue detailing where its required and we can work towards getting that implemented as well. Going to close this PR for now.

PR #1573, opened 2025-07-31
Added MIG configs for RTX pro 6000

JunAr7112:


assets/state-mig-manager/0400_configmap.yaml
```diff
@@ -53,6 +53,78 @@ data:
           mig-devices:
             "7g.40gb": 1
 
+      # RTX-PRO-6000-96GB
+      all-1g.24gb+gfx:
+        - device-filter: ["0x2bb510DE"]
+          devices: all
```
cdesiniotis: To align with our other `all-*` configs, let's not specify a device filter here. Note, the same MIG profile can be supported on more than one device.
```suggestion
        - devices: all
```

assets/state-mig-manager/0400_configmap.yaml
```diff
@@ -53,6 +53,78 @@ data:
           mig-devices:
             "7g.40gb": 1
 
+      # RTX-PRO-6000-96GB
```
cdesiniotis, line 56: To be consistent with the rest of this file, let's keep the comment on this line and remove it from the other `all-*` configs directly below this one.

assets/state-mig-manager/0400_configmap.yaml
```diff
@@ -53,6 +53,78 @@ data:
...
+      # RTX-PRO-6000-24GB
+      all-1g.24gb+me.all:
+        - device-filter: ["0x2bb510DE"]
+          devices: all
+          mig-enabled: true
+          mig-devices:
+            "1g.24gb+me.all": 1
+      
+      # RTX-PRO-6000-96GB
+      all-1g.24gb-me:
```
cdesiniotis, line 69: I don't see a config for `all-1g.24gb`.

cdesiniotis: Ah nvm, I see later on that a config already exists for it and we added the devid for RTX Pro 6000 there.

assets/state-mig-manager/0400_configmap.yaml
```diff
@@ -301,6 +373,14 @@ data:
             "1g.24gb": 2
             "2g.47gb": 1
             "3g.95gb": 1
+        
+        # RTX-PRO-6000-96GB
+        - device-filter: ["0x2bb510DE"]
```
cdesiniotis: Let's capitilize the device id string to align with the other configs

assets/state-mig-manager/0400_configmap.yaml
JunAr7112: Alright I'll switch to this.

assets/state-mig-manager/0400_configmap.yaml
JunAr7112: Alright I will remove

PR #1578, opened 2025-08-01
Use CDI by default as the mechanism for injecting GPUs into workload containers

cdesiniotis:


cdesiniotis:
The e2e test failures should be fixed by https://github.com/NVIDIA/gpu-operator/pull/1580

controllers/object_controls.go
```diff
@@ -1223,6 +1225,21 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
...
+		return
+	}
+
+	// When CDI is enabled in GPU Operator, we leverage native CDI support in containerd / cri-o
+	// to inject GPUs into workloads. We do not configure 'nvidia' as the default runtime. The
+	// 'nvidia' runtime will be set as the runtime class for our management containers so that
+	// they get access to all GPUs.
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CrioConfigModeEnvName, "config")
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaRuntimeSetAsDefaultEnvName, "false")
```
shivamerla: Question: Could there be a race condition where the currently running device plugin still allows pods to be scheduled with the envvar strategy, while the toolkit update has already removed the default nvidia runtime?

shivamerla:
Great to see that we are switching to `cdi` as default finally!

controllers/object_controls.go
```diff
@@ -1327,6 +1337,11 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		setContainerEnv(mainContainer, "CONTAINERD_RUNTIME_CLASS", getRuntimeClass(config))
 	}
 
+	if runtime == gpuv1.CRIO.String() {
+		// We add the nvidia runtime to the cri-o config by default instead of installing the OCI prestart hook
+		setContainerEnv(mainContainer, CrioConfigModeEnvName, "config")
```
elezar: Question: Why are we setting this twice?

controllers/object_controls.go
cdesiniotis: Yes, I think this race condition is possible. When upgrading the gpu-operator, new revisions of the plugin and toolkit daemonsets will get rolled out. As we don't coordinate their rollouts, it is possible for the new toolkit pod to start on a node before the new plugin pod. If a GPU pod gets scheduled onto the node at this particular point in time, the old plugin will set the envvar instead of the CDI annotation / CRI fields, which means the pod will not have any GPUs injected. I am not sure how we can avoid this race condition...

elezar: Does that race condition not already exist? It is already possible for users to configure the toolkit as is shown here. How would one ensure ordering / coordination between the components?

elezar (commented):
Thanks @cdesiniotis 

When discussing switching to using native CDI where possible, we also mentioned how to handle containerd 1.6 where CDI is not supported, and we would need to leverage the NVIDIA Container Runtime. That is not called out explicitly here (or in the design doc) and we should remember to flesh out the steps a user would have to take in this case. (or do we just tell these users that they would have to switch to cdi.enabled = false`?)

controllers/object_controls.go
```diff
@@ -1223,6 +1225,21 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
...
+func transformToolkitForCDI(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) {
+	if !config.CDI.IsEnabled() {
+		return
+	}
+
+	// When CDI is enabled in GPU Operator, we leverage native CDI support in containerd / cri-o
+	// to inject GPUs into workloads. We do not configure 'nvidia' as the default runtime. The
+	// 'nvidia' runtime will be set as the runtime class for our management containers so that
+	// they get access to all GPUs.
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
```
tariq1890: instead of referring to the container by a certain index directly, we should retrieve the index of the desired container to modify by looping through the containers and selecting the container that matches the desired name

controllers/transforms_test.go
```diff
@@ -613,6 +614,10 @@ func TestTransformToolkit(t *testing.T) {
 						},
 					},
 					Env: []corev1.EnvVar{
+						{Name: CDIEnabledEnvName, Value: "true"},
+						{Name: CrioConfigModeEnvName, Value: "config"},
```
tariq1890:
```suggestion
						{Name: CRIOConfigModeEnvName, Value: "config"},
```

config/crd/bases/nvidia.com_clusterpolicies.yaml
```diff
@@ -136,13 +136,14 @@ spec:
...
+                    description: 'Deprecated: This field is no longer used. Setting
+                      cdi.enabled=true will configure CDI as the default mechanism
+                      for making GPUs accessible to containers.'
                     type: boolean
                   enabled:
-                    default: false
-                    description: Enabled indicates whether CDI can be used to make
-                      GPUs accessible to containers.
+                    default: true
+                    description: Enabled indicates whether CDI should be used as the
```
tariq1890: we should probably expand CDI here. Let's update the godoc in `clusterpolicy_types.go`

controllers/object_controls.go
cdesiniotis: Yes, this race condition does already exist. The only way that I know of to coordinate is to write a custom controller...

controllers/object_controls.go
cdesiniotis: You are right, this envvar is getting set here and in `transformForCDI()`. I believe I can remove it from `transformForCDI()` now that we are always defaulting to the `config` mode for cri-o here...

cdesiniotis: However the unit tests confirm that the resulting pod spec only has this envvar set once. It is because `setContainerEnv()` handles duplicates -- it will override the value for an existing key instead of blindly appending envvars to the list.

controllers/object_controls.go
shivamerla: Yes, users could manually edit these today and run into the same issue. The difference here is that we're toggling the default behavior automatically, which makes it more prone to occur during upgrades. If possible we have to rollout the device-plugin change first in the state machine followed by the toolkit.

cdesiniotis: > If possible we have to rollout the device-plugin change first in the state machine followed by the toolkit.

How would we enforce ordering?

elezar:
@cdesiniotis if I understood our discussions correctly, we're switching to NATIVE CDI here. Should we update the PR title / description and commit messages to indicate this.

controllers/object_controls.go
shivamerla: This is a bit tricky since toolkit/plugin DaemonSet pod updates can rollout on any node, so they may not end up on the same one. Another option is to continue setting the NVIDIA runtime as the default. In native CDI mode, the toolkit should effectively be a no-op for a couple of releases, giving us a smooth transition period. After that, we can remove the default runtime entirely.

controllers/transforms_test.go
cdesiniotis: Updated.

config/crd/bases/nvidia.com_clusterpolicies.yaml
cdesiniotis: Updated.

controllers/object_controls.go
cdesiniotis: I agree. This PR is already getting quite large, so I might clean this up in a follow-up. I would also need to update `TransformToolkit()` and `TransformDevicePlugin()` to retrieve the container at a particular index as well.

xiongzubiao:
This is really useful to support third-party OCI runtimes in CDI mode. Will it be included in the next release?

controllers/object_controls.go
```diff
@@ -1223,6 +1225,21 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
...
+	setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CRIOConfigModeEnvName, "config")
```
tariq1890: What if we changed the default crio config mode from `hook` to `config` in v1.18.0 of toolkit?

controllers/object_controls.go
```diff
@@ -1255,14 +1272,7 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
 	// update env required for CDI support
-	if config.CDI.IsEnabled() {
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCtrRuntimeCDIPrefixesEnvName, "nvidia.cdi.k8s.io/")
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CrioConfigModeEnvName, "config")
-		if config.CDI.IsDefault() {
-			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCtrRuntimeModeEnvName, "cdi")
-		}
-	}
+	transformToolkitForCDI(obj, config, n)
```
tariq1890: I think it reads better to move the `CDI.IsEnabled()` check outside of this method

```suggestion
if config.CDI.IsEnabled() {
    transformToolkitForCDI(obj, config, n)
}
```

deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
```diff
@@ -136,13 +136,15 @@ spec:
...
+                    description: 'Deprecated: This field is no longer used. Setting
+                      cdi.enabled=true will configure CDI as the default mechanism
+                      for making GPUs accessible to containers.'
                     type: boolean
                   enabled:
-                    default: false
-                    description: Enabled indicates whether CDI can be used to make
-                      GPUs accessible to containers.
+                    default: true
+                    description: Enabled indicates whether Container Device Interface
```
tariq1890:
```suggestion
                    description: Enabled indicates whether the Container Device Interface
```

controllers/object_controls.go
```diff
@@ -1381,6 +1396,20 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 	return nil
 }
 
+func transformDevicePluginForCDI(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) {
+	if !config.CDI.IsEnabled() {
```
tariq1890: same comment as the one for `transformToolkitForCDI`

controllers/object_controls.go
cdesiniotis: @tariq1890 had an idea that would enforce ordering -- what if we deleted the other operand pods (e.g. device-plugin, gfd, dcgm-exporter) in the toolkit's initContainer? This is similar to what we do in the driver's initContainer in order to ensure that all the other operands (which depend on the driver) start AFTER the driver pod is in running and ready.

On each node where the new toolkit is rolled out on, we have it's initContainer delete the other operand pods, which forces the device-plugin to also upgrade on that node (if it hadn't already). This would address the race condition pointed out in this thread.

@shivamerla thoughts? Is the added complexity of adding this logic in the toolkit's initContainer worth it in your opinion? Also note, this does not prevent all possible race conditions. For example, it is still possible for the new device-plugin to run on a node before the new toolkit -- this could be problematic if a new GPU pod got scheduled and the underlying runtime (containerd) does not have CDI enabled, as the new plugin will not be setting the `NVIDIA_VISIBLE_DEVICES` envvar anymore.

cdesiniotis:
> When discussing switching to using native CDI where possible, we also mentioned how to handle containerd 1.6 where CDI is not supported, and we would need to leverage the NVIDIA Container Runtime. That is not called out explicitly here (or in the design doc) and we should remember to flesh out the steps a user would have to take in this case. (or do we just tell these users that they would have to switch to cdi.enabled = false`?)

Users have two options, both of which I have tested on a system with containerd 1.6 running.

Option 1) set `nvidia` as the default runtime:
```
toolkit:
  env:
  - name: "NVIDIA_RUNTIME_SET_AS_DEFAULT"
    value: "true"
```

Option 2) disable CDI:
```
cdi:
  enabled: false
```

We just need to decide which option (or both) to call out in our documentation / release notes. The one benefit of option 1 is that users would still be using CDI (in our runtime as opposed to containerd).

deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml
cdesiniotis: Updated.

docker/Dockerfile
```diff
@@ -121,6 +121,13 @@ COPY --from=sample-builder /build/vectorAdd /usr/bin/vectorAdd
 # Once new sample images are published that contain the compat libs, we can update the below.
 COPY --from=builder /usr/local/cuda/compat /usr/local/cuda/compat
 
+# The distroless image does not contain the ld.so.cache file. The update-ldcache
+# createContainer hook will skip updating the container's ldcache if this file
+# does not exist. This can lead to issues running the CUDA vectorAdd sample
+# if the NVIDIA libraries are not present in the default dynamic linker search
+# path(s).
+COPY --from=sample-builder /etc/ld.so.cache /etc/
```
elezar, line 129: Does setting the `LD_LIBRARY_PATH` not make more sense? (although we may not know where this is).

controllers/object_controls.go
```diff
@@ -1327,6 +1327,11 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		setContainerEnv(mainContainer, "CONTAINERD_RUNTIME_CLASS", getRuntimeClass(config))
 	}
 
+	if runtime == gpuv1.CRIO.String() {
+		// We add the nvidia runtime to the cri-o config by default instead of installing the OCI prestart hook
+		setContainerEnv(mainContainer, CRIOConfigModeEnvName, "config")
```
elezar, line 1355: Do we now do this twice, or is the `mainContainer` different from `obj.Spec.Template.Spec.Containers[0]`?

controllers/object_controls.go
cdesiniotis: Updated this to not refer to a certain container by index.

controllers/object_controls.go

controllers/object_controls.go
cdesiniotis: I have updated this so that we only set this envvar here.

controllers/object_controls.go

docker/Dockerfile
cdesiniotis: Are you suggesting we set `LD_LIBRARY_PATH` to all possible directories where the driver libraries can be found at in the container? Is the set of possible directories known up front?

I noticed on my OpenShift node that libraries got installed at `/usr/lib64` while on an Ubuntu node driver libraries got installed at `/usr/lib/x86_64-linux-gnu`.

controllers/object_controls.go
elezar: OK .So this means that we DON'T SUPPORT hook-based injection for cri-o -- unless a user explicitly selects it? (to be clear, given the changes in v1.18.0, I would recommend this). Should we issue a warning if hook is selected (perhaps in the tookit-container)?

controllers/object_controls.go
```diff
@@ -1433,8 +1433,8 @@ func TransformDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 		return err
 	}
 
-	// set RuntimeClass for supported runtimes
-	setRuntimeClass(&obj.Spec.Template.Spec, n.runtime, config.Operator.RuntimeClass)
+	// set RuntimeClassName
```
elezar: Nit: I don't see this comment as useful. It's basically just the same as the function name. We could consider renaming the function to `setRuntimeClassName` and removing this comment entirely.

controllers/object_controls.go
```diff
@@ -1223,8 +1225,33 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
 	return envVars
 }
 
+func transformToolkitCtrForCDI(container *corev1.Container, config *gpuv1.ClusterPolicySpec) {
+	// When CDI is enabled in GPU Operator, we leverage native CDI support in containerd / cri-o
+	// to inject GPUs into workloads. We do not configure 'nvidia' as the default runtime. The
+	// 'nvidia' runtime will be set as the runtime class for our management containers so that
+	// they get access to all GPUs.
```
elezar, line 1231: Should we add a note that a user COULD override this by setting the envvar directly?

controllers/object_controls.go
```diff
@@ -1223,8 +1225,33 @@ func getProxyEnv(proxyConfig *apiconfigv1.Proxy) []corev1.EnvVar {
...
+	setContainerEnv(container, CDIEnabledEnvName, "true")
+	setContainerEnv(container, NvidiaRuntimeSetAsDefaultEnvName, "false")
+
+	if config.CDI.IsDefault() {
+		setContainerEnv(container, NvidiaCtrRuntimeModeEnvName, "cdi")
+	}
```
elezar: Note to self: This will not trigger the automatic generation of CDI specs in the toolkit.

controllers/object_controls.go
```diff
@@ -1256,16 +1283,12 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
 
 	// update env required for CDI support
 	if config.CDI.IsEnabled() {
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), CDIEnabledEnvName, "true")
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), NvidiaCtrRuntimeCDIPrefixesEnvName, "nvidia.cdi.k8s.io/")
```
elezar, line 1260: Since we remove this, do we also expect the `nvidia` runtime to handle annotation-based injections that use `cdi.k8s.io/` annotations?

controllers/object_controls.go
```diff
@@ -1293,13 +1316,13 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
 		for _, env := range config.Toolkit.Env {
-			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
+			setContainerEnv(mainContainer, env.Name, env.Value)
 		}
 	}
 
 	// configure runtime
 	runtime := n.runtime.String()
-	err = transformForRuntime(obj, config, runtime, "nvidia-container-toolkit-ctr")
+	err = transformForRuntime(obj, config, runtime, mainContainerName)
```
elezar, line 1326: Something for a follow-up: Does it make sense to rather have `transformForRuntime` accept a pointer to a container instead of a name? We then don't need to iterate over the containers there to find the right container. (Alternatively we could consider an implementation that uses generics).

controllers/object_controls.go
```diff
@@ -1385,8 +1408,30 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+	mainContainerName := "nvidia-device-plugin"
+	for i, ctr := range obj.Spec.Template.Spec.Containers {
+		if ctr.Name == mainContainerName {
+			mainContainer = &obj.Spec.Template.Spec.Containers[i]
+			break
+		}
+	}
+	if mainContainer == nil {
+		return fmt.Errorf("failed to find main container %q", mainContainerName)
+	}
```
elezar, line 1434: Something for follow-up: It may clean things up to have a `findContainerByName` function that returns the container we want.

controllers/object_controls.go
```diff
@@ -1398,10 +1443,10 @@ func TransformDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 	if err != nil {
 		return err
 	}
-	obj.Spec.Template.Spec.Containers[0].Image = image
+	mainContainer.Image = image
```
elezar, line 1447: As a note: This is a simple but useful cleanup. The one comment is whether we want to always call this `mainContainer` or something more specific to make global searches simpler?

controllers/object_controls.go
```diff
@@ -1418,13 +1463,13 @@ func TransformDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
...
-		obj.Spec.Template.Spec.Containers[0].Args = config.DevicePlugin.Args
+		mainContainer.Args = config.DevicePlugin.Args
 	}
 
 	// add env to allow injection of /dev/nvidia-fs and /dev/infiniband devices for GDS
 	if config.GPUDirectStorage != nil && config.GPUDirectStorage.IsEnabled() {
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), GDSEnabledEnvName, "true")
-		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), MOFEDEnabledEnvName, "true")
+		setContainerEnv(mainContainer, GDSEnabledEnvName, "true")
+		setContainerEnv(mainContainer, MOFEDEnabledEnvName, "true")
```
elezar, line 1473: Note that we're going to have to do the same for GDR copy (and also merge https://github.com/NVIDIA/k8s-device-plugin/pull/1339).

deployments/gpu-operator/values.yaml
```diff
@@ -14,7 +14,6 @@ psa:
 
 cdi:
   enabled: false
-  default: false
```
elezar, line 17: Question: What impact does removing this have on downgrades?

deployments/gpu-operator/values.yaml
```diff
@@ -13,7 +13,7 @@ psa:
   enabled: false
 
 cdi:
-  enabled: false
+  enabled: true
```
elezar, line 16: Is it needed to handle the case where this was NOT set (i.e. `null`) and a user upgrades?

tests/holodeck.yaml
elezar, line 1: As a nit, I think we could make this change independent of this PR if we wanted to push it though.

docker/Dockerfile
elezar: Yes, that is the tricky part. We have discussed always mouting the libraries to the same folder in any container, but have not done this yet. If importing the ld.cache works, I think I'm ok with doing this for the time being.

controllers/object_controls.go
elezar: What about this as an alternative: Since the device plugin REQUIRES the NVIDIA Container Toolkit to start up and have access to the GPUs, we could check whether it has started with the expected toolkit configuration and if not, then fail? One option would be to add an envvar to the management CDI spec that we could check for, for example. 
We do in fact already inject such an envvar which we could consider using: https://github.com/NVIDIA/nvidia-container-toolkit/blob/8739e483f1800964adbb920335db7e7d32a88d02/pkg/nvcdi/driver-nvml.go#L117-L120

docker/Dockerfile
elezar: Could we use `NVIDIA_CTK_LIBCUDA_DIR` to set `LD_LIBRARY_PATH` for the validator? See https://github.com/NVIDIA/nvidia-container-toolkit/blob/8739e483f1800964adbb920335db7e7d32a88d02/pkg/nvcdi/driver-nvml.go#L117-L120

controllers/object_controls.go
cdesiniotis: Correct, hook-based injection will not be used for cri-o unless a user explicitly selects it, by setting `CRIO_CONFIG_MODE=hook`. I am not sure that this means hook-based injection is no longer "supported" though.

> Should we issue a warning if hook is selected (perhaps in the tookit-container)?

If we want to start discouraging the use of it (and maybe "deprecate" it?) then I wouldn't be opposed to this.

deployments/gpu-operator/values.yaml
cdesiniotis: I don't think this should impact downgrades. Older releases of the operator (where the `cdi.default` field is not a no-op) handle the case where this field is omitted from the ClusterPolicy spec -- the default value is `false` in this case: https://github.com/NVIDIA/gpu-operator/blob/v25.3.3/api/nvidia/v1/clusterpolicy_types.go#L2074-L2079

deployments/gpu-operator/values.yaml
cdesiniotis: No. This is not strictly required in this case, because when the `cdi.enabled` field is not set we default to `true` in the code.

I would argue this is needed in the case where `cdi.enabled=false` in the older operator version. During the upgrade, we want to toggle this to `true` -- if the user wants to continue to NOT use CDI, they will need to provide an override during the helm upgrade.

controllers/object_controls.go
cdesiniotis: Hmm I am trying to understand the use case / different possibilities here.

By default, the nvidia runtime will not be invoked for pods that request a GPU -- as nvidia is not configured as the default runtime. The CDI device passed via the CRI field is what will trigger the runtime, like containerd, to inject the requested GPU(s). 

If a user sets `runtimeClassName=nvidia` in their pod spec, then the nvidia runtime would be invoked. By default, the nvidia runtime is configured with `annotation-prefixes = ["cdi.k8s.io/"]` (right?) so it would detect the GPU specified in the annotation (that our device-plugin set in its AllocateResonse) and inject it using CDI. I guess containerd / cri-o would also inject the GPU(s), but I assume things should just "work" even if the CDI device injection is performed twice.

Both of these cases should work, correct? Am I missing anything?

controllers/object_controls.go
cdesiniotis: I agree, I can make the proposed change in a follow-up.

controllers/object_controls.go
cdesiniotis: Maybe `toolkitMainContainer` in the `TransformToolkit` function and `devicePluginMainContainer` in the `TransformDevicePlugin` function?

controllers/object_controls.go
cdesiniotis: Agreed. I removed the comments throughout and updated the function to be named `setRuntimeClassName`

controllers/object_controls.go
cdesiniotis: It wouldn't hurt. I have gone ahead and added a comment explaining this.

docker/Dockerfile
cdesiniotis: Is `NVIDIA_CTK_LIBCUDA_DIR` the path of libcuda.so's parent directory on the host AND in the container?

I guess we could leverage this. Since the NVIDIA_CTK_LIBCIDA_DIR envvar is not known until runtime, I assume we would have to set `LD_LIBRARY_PATH=$NVIDIA_CTK_LIBCUDA_DIR` in the entrypoint command of the validator container? Or is there another way to leverage this envvar?

controllers/object_controls.go
cdesiniotis: Are you suggesting we add a new envvar that represents the current toolkit configuration (maybe a hash of the `toml` file)? And the device-plugin, on startup, would recompute the hash and compare it? If I am understanding your proposal correctly then that means we would also need to mount the toolkit configuration file into the device-plugin.

tests/holodeck.yaml
```diff
@@ -27,6 +27,7 @@ spec:
   containerRuntime:
     install: true
     name: containerd
+    version: 1.7.27
```
tariq1890, line 30: Do we want to start pointing to containerd 2.x versions? Or is the intention to test with the minimum support version of a runtime with native CDI?

cdesiniotis: The latter. Technically CDI support was introduced in containerd 1.7.0, but I was having difficulties installing versions prior to 1.7.27 (I don't recall why exactly).

cdesiniotis: In an ideal world, we would be testing multiple active branches of containerd, e.g. 1.7 / 2.0 / 2.X

docker/Dockerfile
elezar: The intent is that `NVIDIA_CTK_LIBCUDA_DIR` is the path to the directory containing `libcuda.so` *in the container*. This is not determined at runtime, but is encoded in the CDI spec. Setting the `LD_LIBRARY_PATH=$NVIDIA_CTK_LIBCUDA_DIR:$LD_LIBRARY_PATH` as you specify in the entrypoint was what I was thinking.

Like I said ... if adding a dummy ldcache works for now, I'd be happy to look at this in a follow-up.

controllers/object_controls.go
elezar: I would strongly suggest deprecating it in this context. The feature drift between the `legacy` hook and the new runtime (or CDI) is significant. I would also argue that crio users are more likely to have more restrictive configurations such as SELinux enabled.

controllers/object_controls.go
elezar: (I really remember typing a response here, but it seems to have gone missing).

Making the same modifications twice should be fine, but there may be some unexpected sideeffects specifically if hooks are run multiple times. The creation of symlinks *should* be idempotent since we do check whether the symlink that we're creating is already valid. In the case of something like `update-ldcache`, there may be some side-effects like multiple `/etc/ld.so.conf.d` files with the same contents.

To be clear, this is not a configuration that I can test easily given my available machines, so if you have a test system that I could have access to (or that we can pair on), that may be useful.

controllers/object_controls.go
elezar: Yes, that's what I was referring to. Not critical for this change though.

deployments/gpu-operator/values.yaml
elezar: OK, so we would need to explicitly document this override? (I am a little rusty on helm upgrade semantics).

controllers/object_controls.go
elezar: Sorry, I mean that we can detect if the device plugin was started with the new toolkit if the device plugin container includes the `NVIDIA_CTK_LIBCUDA_DIR` environment variable. This will only be possible if the toolkit was installed and is running before the device plugin is started.

If we add an init container to the device plugin that blocks on the presence of this envvar (when CDI is enabled?) would that at least ensure the desired ordering? (I don't know whether this approach is perfect, but it's possible with what's already available).

deployments/gpu-operator/values.yaml
cdesiniotis: Yes, we can call this out explicitly in our release notes -- we will definitely mention that the default value of this field is updated in the new release, and we can document how to keep the prior default (false) via an override file or the helm command line.

controllers/object_controls.go
cdesiniotis: Yes, we should do this.

I have updated my PR so that the operator sets `config` as the crio config mode by default. I mentioned this here: https://github.com/NVIDIA/gpu-operator/pull/1578#discussion_r2361236018

I have also raised https://github.com/NVIDIA/nvidia-container-toolkit/pull/1314 to update the default crio config mode that is hardwired in the toolkit container.

cc @elezar

PR #1580, opened 2025-08-04
Cleanup extraneous envvars from values.yaml

cdesiniotis:
This PR fixes a regression introduced in https://github.com/NVIDIA/gpu-operator/commit/54c270cf3f91c39fb1d0b8e2597f389b70975228. When `cdi.enabled` and `cdi.default` are set to `true`, the controller will set `DEVICE_LIST_STRATEGY=ennvar,cdi-annotations` for the device plugin, however, this envvar is also set in `values.yaml` which will take precedence and overrides what the controller configured. While fixing this, I also discovered that many of the envvars specified in our values.yaml file are already set in the DaemonSet manifests in the `assets/` folder.

### DIFF
--- a/assets/state-driver/0500_daemonset.yaml
+++ b/assets/state-driver/0500_daemonset.yaml
@@ -60,7 +60,7 @@ spec:
           - name: ENABLE_GPU_POD_EVICTION
             value: "true"
           - name: ENABLE_AUTO_DRAIN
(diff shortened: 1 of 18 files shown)


tariq1890:
Great catch, @cdesiniotis. Thanks!

tariq1890 (approved):

PR #1583, opened 2025-08-05
Fix regression introduced in helm hook entrypoint commands

cdesiniotis:
After switching the operator image to the distroless base image, /bin/sh is no longer a valid file. The 'sh' binary is present in the image at /busybox/sh and is present in the executable search path.


tariq1890:
> lgtm, eventually both of these should be replaced with crd-apply-tool.

Agreed

PR #1584, opened 2025-08-06
[driver] add support for passing secrets as input

tariq1890:


controllers/nvidiadriver_controller.go
```diff
@@ -168,6 +169,31 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return reconcile.Result{}, nil
 	}
 
+	operatorNamespace := os.Getenv("OPERATOR_NAMESPACE")
```
cdesiniotis, line 172: As an observation, I see we already fetch this env in `main.go` https://github.com/NVIDIA/gpu-operator/blob/d71f30d0d2a1009b52ad5dd41ca8a1a878b7f11e/cmd/gpu-operator/main.go#L109 Thoughts on adding the operator namespace as a field of the `NVIDIADriverReconciler` so that we don't have to repeatedly source this env during every reconciliation?

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -200,6 +200,9 @@ spec:
     {{- if .Values.driver.kernelModuleConfig }}
     kernelModuleConfig: {{ toYaml .Values.driver.kernelModuleConfig | nindent 6 }}
     {{- end }}
+    {{- if .Values.driver.secretEnv }}
+    secretEnv: {{ toYaml .Values.driver.secretEnv | nindent 6 }}
```
cdesiniotis: Question -- is the `nindent 6` needed here if secretEnv is just a string?

controllers/transforms_test.go
```diff
@@ -1570,3 +1599,66 @@ func TestTransformNodeStatusExporter(t *testing.T) {
...
+	testCases := []struct {
+		description   string
+		ds            Daemonset
+		cpSpec        *gpuv1.ClusterPolicySpec
+		client        client.Client
+		expectedDs    Daemonset
+		errorExpected bool
+	}{
+		{
+			description: "driver spec with secret env",
```
cdesiniotis, line 1614: Out of scope for this PR -- we should add some more test cases here. At minimum it would be nice to have a test case where all / most of the driver related fields in ClusterPolicy are set.

controllers/nvidiadriver_controller.go
cdesiniotis, line 172: Out of scope -- I can't leave a review comment on lines that are not part of the diff, but looks like the code immediately before this is stale. 
```
	if instance.Spec.IsGDSEnabled() && instance.Spec.IsOpenKernelModulesRequired() && !instance.Spec.IsOpenKernelModulesEnabled() {
		err = fmt.Errorf("GPUDirect Storage driver '%s' is only supported with NVIDIA OpenRM drivers. Please set 'useOpenKernelModules=true' to enable OpenRM mode", instance.Spec.GPUDirectStorage.Version)
		logger.V(consts.LogLevelError).Error(nil, err.Error())
		instance.Status.State = nvidiav1alpha1.NotReady
		condErr = r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, err.Error())
		if condErr != nil {
			logger.V(consts.LogLevelDebug).Error(nil, condErr.Error())
		}
	}
```

deployments/gpu-operator/templates/clusterpolicy.yaml
tariq1890: Good catch. I'll remove

controllers/nvidiadriver_controller.go
tariq1890: Yes, I had the same thought as well. I was thinking of making this change in both the cluster policy and nvidia driver reconcilers, so we call the os.GetEnv only once. I was thinking of doing this in a follow up PR

controllers/transforms_test.go
tariq1890: Agreed

controllers/nvidiadriver_controller.go
tariq1890: Good catch! Yes, this should be changed

deployments/gpu-operator/templates/clusterpolicy.yaml
tariq1890: This is fixed

controllers/nvidiadriver_controller.go
tariq1890: These are three other places where we do this

- https://github.com/NVIDIA/gpu-operator/blob/d68374e891d6a14242dfd9e48eac88b34c80420d/internal/state/driver.go#L621
- https://github.com/NVIDIA/gpu-operator/blob/f9e0eb7b425813a4800da8401318049e6f90b1da/cmd/gpu-operator/main.go#L109
- https://github.com/NVIDIA/gpu-operator/blob/2e5685a9481c82df67ad07dfb506b2b1d0b9fd1e/controllers/state_manager.go#L762

cdesiniotis: Doing this in a follow-up is fine by me.

tariq1890:
@cdesiniotis Thanks for the review. I've updated the PR to now add secretEnv references to the GDRCopy and GDS containers when this field is set

controllers/object_controls.go
```diff
@@ -2781,6 +2781,14 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 			gdsContainer.VolumeMounts = append(gdsContainer.VolumeMounts, volumeMounts...)
 		}
 
+		secretName := config.Driver.SecretEnv
+		if len(secretName) > 0 {
+			err := createSecretEnvReference(n.ctx, n.client, secretName, n.operatorNamespace, gdsContainer)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to get attach secret %s to the driver container: %w", secretName, err)
```
cdesiniotis:
```suggestion
				return fmt.Errorf("ERROR: failed to attach secret %s to the driver container: %w", secretName, err)
```

controllers/object_controls.go
```diff
@@ -2860,6 +2868,14 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 			gdrcopyContainer.VolumeMounts = append(gdrcopyContainer.VolumeMounts, volumeMounts...)
 		}
 
+		secretName := config.Driver.SecretEnv
+		if len(secretName) > 0 {
+			err := createSecretEnvReference(n.ctx, n.client, secretName, n.operatorNamespace, gdrcopyContainer)
+			if err != nil {
+				return fmt.Errorf("ERROR: failed to get attach secret %s to the driver container: %w", secretName, err)
```
cdesiniotis:
```suggestion
				return fmt.Errorf("ERROR: failed to attach secret %s to the driver container: %w", secretName, err)
```

controllers/object_controls.go
```diff
@@ -3372,6 +3388,14 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 		podSpec.Volumes = append(podSpec.Volumes, createConfigMapVolume(config.Driver.CertConfig.Name, itemsToInclude))
 	}
 
+	secretName := config.Driver.SecretEnv
+	if len(secretName) > 0 {
+		err := createSecretEnvReference(n.ctx, n.client, secretName, n.operatorNamespace, driverContainer)
+		if err != nil {
+			return fmt.Errorf("ERROR: failed to get attach secret %s to the driver container: %w", secretName, err)
```
cdesiniotis:
```suggestion
			return fmt.Errorf("ERROR: failed to attach secret %s to the driver container: %w", secretName, err)
```

controllers/object_controls.go
tariq1890: Thanks. Will fix

tariq1890: Fixed

PR #1605, opened 2025-08-18
Added NSPECT to pipeline

JunAr7112:


.nvidia-ci.yml
```diff
@@ -136,10 +136,8 @@ release:ngc-gpu-operator:
     - .release:ngc
     - .target-gpu-operator
 
-raise-publishing-merge-request:
+.raise-merge-request:
```
cdesiniotis: Out of scope for this PR -- we should remove the `release:ngc-gpu-operator` job (a few lines above) now as that is the old publishing process.

.nvidia-ci.yml
```diff
@@ -149,6 +147,63 @@ raise-publishing-merge-request:
     - cnt-ngc-publish render --project-name "gpu-operator" --versions-file "build-info-${CI_PIPELINE_ID}.txt" --output gpu-operator.yaml
     - cnt-ngc-publish merge-request --files "gpu-operator.yaml"
 
+raise-publishing-merge-request:
```
cdesiniotis: I see in the toolkit Evan has named this job `publish-images-to-ngc` https://github.com/NVIDIA/nvidia-container-toolkit/blob/77e8f4ba6f6d8525d7956d98e3d0e4f08e5ad7ef/.nvidia-ci.yml#L276, do we want to follow suite?

.nvidia-ci.yml
```diff
@@ -149,6 +147,63 @@ raise-publishing-merge-request:
...
+  extends:
+    - .raise-merge-request
+  rules:
+    - if: $CI_COMMIT_TAG
+  needs:
+    - job: release:ngc-gpu-operator
+
+raise-staging-merge-request:
```
cdesiniotis: nit: what about renaming this job to `raise-dummy-publishing-merge-request` to indicate that this job is soley for testing purposes? Or `publish-images-dummy` if we name similar to the toolkit repo.

Also, let's add a comment above this job to indicate it's intent.
```
# Create a dummy MR that exercises the publishing logic
```

.nvidia-ci.yml
```diff
@@ -149,6 +147,63 @@ raise-publishing-merge-request:
...
+    name: "${CNT_NGC_PUBLISH_IMAGE}"
+    pull_policy: always
+  variables:
+    PROJECT_NAME: "gpu-operator"
+    REPO_URL: "https://github.com/NVIDIA/gpu-operator.git"
+  script:
+    - |
+      cnt-ngc-publish nspect --versions-file "build-info-${CI_PIPELINE_ID}.txt"
+
+update-nspect-staging:
```
cdesiniotis, line 145: Let's add a comment above this job to indicate it's intent:
```
# Update the nspect staging environment in order to exercise the nspect publishing logic
```

.nvidia-ci.yml
cdesiniotis: Oops, this is incorrect. Disregard this comment. The `release:ngc-gpu-operator` job is required still.

.nvidia-ci.yml
```diff
@@ -149,6 +147,63 @@ raise-publishing-merge-request:
...
+  needs:
+    - job: release:ngc-gpu-operator
+
+raise-staging-merge-request:
+  extends:
+    - .raise-merge-request
+  needs:
+    - job: release:staging-gpu-operator
+  before_script:
+    - export NGC_PUBLISHING_PROJECT_PATH="dl/container-dev/ngc-automation"
```
tariq1890: gitlab CI supports setting envars via the `env:` yaml field. Let's use that instead of an `export` statement

.nvidia-ci.yml
tariq1890: What about `release:ngc-publish` or `release:ngc-image-publish` ?

.nvidia-ci.yml
tariq1890: Sorry, `variables:`, not `env:`

.nvidia-ci.yml
JunAr7112: Alright I will have job names: release:ngc-image-publish and release:dummy-image-publish with template .publish-images.

.nvidia-ci.yml
JunAr7112: I'll add the comment, and use the format release:dummy-image-publish so it matches with Tariq's format suggestion.

.nvidia-ci.yml
JunAr7112: Alright

.nvidia-ci.yml
JunAr7112: So right now, NGC_PUBLISHING_PROJECT_PATH is set to the Publishing Configs repo as a GROUP lvl variable, that takes precedence over a job variable. We will need to export to ensure the automation path takes precedence.

.common-ci.yml
```diff
@@ -114,7 +114,6 @@ trigger-pipeline:
     - regctl registry login "${CI_REGISTRY}" -u "${CI_REGISTRY_USER}" -p "${CI_REGISTRY_PASSWORD}"
     - '[ ${CI_REGISTRY} = ${OUT_REGISTRY} ] || echo "Logging in to output registry ${OUT_REGISTRY}"'
     - '[ ${CI_REGISTRY} = ${OUT_REGISTRY} ] || regctl registry login "${OUT_REGISTRY}" -u "${OUT_REGISTRY_USER}" -p "${OUT_REGISTRY_TOKEN}"'
-
```
cdesiniotis: Let's revert this change.

.nvidia-ci.yml
```diff
@@ -123,23 +123,25 @@ scan:gpu-operator-arm64:
...
-    - cat build-info-${CI_PIPELINE_ID}.txt
-  artifacts:
-    paths:
-      - build-info-${CI_PIPELINE_ID}.txt
 
 release:ngc-gpu-operator:
   extends:
     - .release:ngc
-    - .target-gpu-operator
+    - .target-gpu-operator
```
cdesiniotis: Let's remove the trailing spaces added here

.nvidia-ci.yml
```diff
@@ -123,23 +123,25 @@ scan:gpu-operator-arm64:
...
-    - .target-gpu-operator
+    - .target-gpu-operator  
 
-raise-publishing-merge-request:
+.set-publish-images-values:
+  before_script:
+    - |
+      if [-n "${CI_COMMIT_TAG}"]; then
+        echo "${CI_COMMIT_SHORT_SHA} ${CI_COMMIT_TAG}" > build-info-${CI_PIPELINE_ID}.txt
+        export NSPECT_CLIENT_ID="${NSPECT_CLIENT_ID}" NSPECT_CLIENT_SECRET="${NSPECT_CLIENT_SECRET}"
```
cdesiniotis: For readability, let's define one envvar per line:
```suggestion
        export NSPECT_CLIENT_ID="${NSPECT_CLIENT_ID}" 
        export NSPECT_CLIENT_SECRET="${NSPECT_CLIENT_SECRET}"
```

.nvidia-ci.yml
```diff
@@ -123,23 +123,25 @@ scan:gpu-operator-arm64:
...
+      else
+        echo "${CI_COMMIT_SHORT_SHA} ${CI_COMMIT_SHORT_SHA}" > build-info-${CI_PIPELINE_ID}.txt
+        export NGC_PUBLISHING_PROJECT_PATH="${NGC_PUBLISHING_TEST_PROJECT_PATH}" NSPECT_CLIENT_ID="${NSPECT_STAGING_CLIENT_ID}" NSPECT_CLIENT_SECRET="${NSPECT_STAGING_CLIENT_SECRET}"
```
cdesiniotis: For readability, let's define one envvar per line:
```suggestion
        export NGC_PUBLISHING_PROJECT_PATH="${NGC_PUBLISHING_TEST_PROJECT_PATH}" 
        export NSPECT_CLIENT_ID="${NSPECT_STAGING_CLIENT_ID}" 
        export NSPECT_CLIENT_SECRET="${NSPECT_STAGING_CLIENT_SECRET}"
```

.nvidia-ci.yml
```diff
@@ -123,23 +123,25 @@ scan:gpu-operator-arm64:
...
-      - build-info-${CI_PIPELINE_ID}.txt
 
 release:ngc-gpu-operator:
   extends:
     - .release:ngc
-    - .target-gpu-operator
+    - .target-gpu-operator  
 
-raise-publishing-merge-request:
+.set-publish-images-values:
```
cdesiniotis:
```suggestion
.ngc-publish-variables:
```

.nvidia-ci.yml
cdesiniotis: It is still a bit dubious that we have to explicitly export these variables here, while we are able to define other variables, like `GITLAB_ACCESS_TOKEN`, in the `variables` section. If possible, it would be preferred to define all variables in the `variables` section.

.nvidia-ci.yml
```diff
@@ -123,23 +123,25 @@ scan:gpu-operator-arm64:
...
     OUT_REGISTRY: "${NGC_REGISTRY}"
-  script:
-    - echo "${VERSION} ${OUT_IMAGE_VERSION}" > build-info-${CI_PIPELINE_ID}.txt
-    - echo "Printing source and target image tags"
-    - cat build-info-${CI_PIPELINE_ID}.txt
-  artifacts:
-    paths:
-      - build-info-${CI_PIPELINE_ID}.txt
 
 release:ngc-gpu-operator:
```
cdesiniotis: This job is no longer required now, right? I believe we can remove `release:ngc-gpu-operator` and `.release:ngc`

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
     - cnt-ngc-publish render --project-name "gpu-operator" --versions-file "build-info-${CI_PIPELINE_ID}.txt" --output gpu-operator.yaml
     - cnt-ngc-publish merge-request --files "gpu-operator.yaml"
 
+# Raise an MR to publish the image to NGC
+release:ngc-image-publish:
```
cdesiniotis: I don't want to be too picky here but the name of this job doesn't feel right to me. "release" and "publish" feel redundant. What about `release:ngc-image` and `release:dummy-image`? Open to other suggestions.

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+  extends:
+    - .publish-images
+    - .set-publish-images-values
+  rules:
+    - if: $CI_COMMIT_TAG
+  needs:
+    - job: release:ngc-gpu-operator
```
cdesiniotis: We just depend on the job that pushes the image to nvstaging, right?
```suggestion
    - job: release:staging-gpu-operator
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+    - .update-nspect
+    - .set-publish-images-values
+  variables:
+    ENV: "stage"
+    RELEASE_VERSION: "v1.0.0"
+
+# Update the nspect production environment with the new release
+update-nspect-publishing:
+  needs:
+    - job: release:ngc-gpu-operator
```
cdesiniotis: We just depend on the job that pushes the image to nvstaging, right?
```suggestion
    - job: release:staging-gpu-operator
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+  script:
+    - |
+      echo "${VERSION} ${OUT_IMAGE_VERSION}" > build-info-${CI_PIPELINE_ID}.txt
+      cnt-ngc-publish nspect --versions-file "build-info-${CI_PIPELINE_ID}.txt"
+
+# Update the nspect staging environment to test the nspect publishing logic
+update-nspect-staging:
+  needs:
+    - job: release:staging-gpu-operator
+      artifacts: true
```
cdesiniotis: The staging job no longer generates an artifact.
```suggestion
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+    - .set-publish-images-values
+  variables:
+    ENV: "stage"
+    RELEASE_VERSION: "v1.0.0"
+
+# Update the nspect production environment with the new release
+update-nspect-publishing:
+  needs:
+    - job: release:ngc-gpu-operator
+      artifacts: true
```
cdesiniotis: The staging job no longer generates an artifact.
```suggestion
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+  extends:
+    - .update-nspect
+    - .set-publish-images-values
+  variables:
+    ENV: "stage"
+    RELEASE_VERSION: "v1.0.0"
+
+# Update the nspect production environment with the new release
+update-nspect-publishing:
```
cdesiniotis:
```suggestion
update-nspect:
```

.nvidia-ci.yml
cdesiniotis: Now that I think about this more, we actually want this job to run _**after**_ nspect is updated, right?. So I believe we actually want:
```suggestion
    - job: update-nspect
```

And for readability sake, let's move the nspect job definitions so that they come before the image publishing job definitions.

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+  needs:
+    - job: release:ngc-gpu-operator
+
+# Create a dummy MR that exercises the publishing logic
+release:dummy-image-publish:
+  extends:
+    - .publish-images
+    - .set-publish-images-values
+  needs:
+    - job: release:staging-gpu-operator
```
cdesiniotis, line 131:
```suggestion
    - job: update-nspect-staging
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
     - cnt-ngc-publish render --project-name "gpu-operator" --versions-file "build-info-${CI_PIPELINE_ID}.txt" --output gpu-operator.yaml
     - cnt-ngc-publish merge-request --files "gpu-operator.yaml"
 
+# Raise an MR to publish the image to NGC
+release:ngc-image-publish:
+  extends:
+    - .publish-images
+    - .set-publish-images-values
```
cdesiniotis: Shouldn't we add this to the `.publish-images` template instead?
```suggestion
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+  rules:
+    - if: $CI_COMMIT_TAG
+  needs:
+    - job: release:ngc-gpu-operator
+
+# Create a dummy MR that exercises the publishing logic
+release:dummy-image-publish:
+  extends:
+    - .publish-images
+    - .set-publish-images-values
```
cdesiniotis: Shouldn't we add this to the `.publish-images` template instead?
```suggestion
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+      cnt-ngc-publish nspect --versions-file "build-info-${CI_PIPELINE_ID}.txt"
+
+# Update the nspect staging environment to test the nspect publishing logic
+update-nspect-staging:
+  needs:
+    - job: release:staging-gpu-operator
+      artifacts: true
+  extends:
+    - .update-nspect
+    - .set-publish-images-values
```
cdesiniotis: Shouldn't we add this to the `.update-nspect` template instead?
```suggestion
```

.nvidia-ci.yml
```diff
@@ -149,6 +151,64 @@ raise-publishing-merge-request:
...
+    RELEASE_VERSION: "v1.0.0"
+
+# Update the nspect production environment with the new release
+update-nspect-publishing:
+  needs:
+    - job: release:ngc-gpu-operator
+      artifacts: true
+  extends:
+    - .update-nspect
+    - .set-publish-images-values
```
cdesiniotis: Shouldn't we add this to the `.update-nspect` template instead?
```suggestion
```

.nvidia-ci.yml
tariq1890: Agreed, I think we should look into this further and leverage `variables`. If `variables` is not an option, then we should at least get more clarity on why explicit `export` statements are needed here

.nvidia-ci.yml
tariq1890: How about `release:ngc-image-push`? Or `release:ngc-image` could work

I'd rather not use the term `dummy`. I prefer `staging` or `test` instead

tariq1890: If it really serves the purpose of a "dummy image", then we could name it as "mock"

cdesiniotis: What about `ngc-image-publish` and `mock-image-publish`?

.common-ci.yml
JunAr7112: Reverted

.nvidia-ci.yml
JunAr7112: Removed

.nvidia-ci.yml
JunAr7112: Updated format here.

.nvidia-ci.yml
JunAr7112: Updated format.

.nvidia-ci.yml
JunAr7112: Sure we can use this naming convention.

.nvidia-ci.yml
JunAr7112: I tried running yesterday with NSPECT credentials in variable sections, but it seems to default to different credential values:

INFO:nspect_stream.SSAToken:Getting token for https://hsxntwlenuyxjgvcdt5ifkp9x78g9spcncmtwuyoxws.stg.ssa.nvidia.com/token
INFO:nspect_stream.SSAToken:Scopes: public.api:read public.api:write
INFO:nspect_stream.SSAToken:SSA client id: nvssa-prd-pa******************************jo
INFO:nspect_stream.SSAToken:SSA client secret: ssap-3**********Vn
ERROR:nspect_stream.SSAToken:Failed to fetch SSA client token

.nvidia-ci.yml
JunAr7112: Yeah I'll remove them.

.nvidia-ci.yml
JunAr7112: Ok I can switch the dependency to nspect.

.nvidia-ci.yml
JunAr7112: Sure, I don't think there are any dependencies in terms of artifacts anymore, but we can set the needs to release:staging.

.nvidia-ci.yml
JunAr7112: Switched name.

.nvidia-ci.yml
JunAr7112: Sure I can move this extends to the template.

.nvidia-ci.yml
JunAr7112: Ok I can also create a separate extends section in the template.

.common-ci.yml
```diff
@@ -114,7 +114,7 @@ trigger-pipeline:
...
+
```
tariq1890: This looks like extraneous whitespace, please remove

.nvidia-ci.yml
tariq1890: It seems like this hasn't been addressed.

.nvidia-ci.yml
```diff
@@ -116,30 +116,60 @@ scan:gpu-operator-arm64:
...
+update-nspect-staging:
   extends:
-    - .release:ngc
-    - .target-gpu-operator
+    - .update-nspect
+  except:
+    - tags
+  variables:
+    ENV: "stage"
+    RELEASE_VERSION: "v1.0.0"
```
tariq1890: We shouldn't be hard-coding the release version here to `v1.0.0`

.common-ci.yml
JunAr7112: ok

JunAr7112: Also for the below comment (can't seem to respond to it), I updated the NGC-IMAGE-PUBLISH and MOCK-IMAGE-PUBLISH to depend on update-nspect:

ngc-image-publish:
  extends:
    - .publish-images
  rules:
    - if: $CI_COMMIT_TAG
  needs:
    - job: update-nspect

.nvidia-ci.yml
JunAr7112: Hi Tariq, when we were discussing earlier, we didn't want to create a new version in STAGING every time we tested the pipeline. So I was just going to keep the test version hard coded to v1.0.0 since that already exists in the repo.

tariq1890: What if we use a test version? Like `${COMMIT_SHA}-${LATEST_RELEASE}` ?

tariq1890: It seems odd to hard-code `v1.0.0` here. I would rather it be called "test" or "devel" since these are mock nspect versions

.nvidia-ci.yml
JunAr7112: I updated the NGC-IMAGE-PUBLISH and MOCK-IMAGE-PUBLISH to depend on update-nspect:

ngc-image-publish:
extends:
- .publish-images
rules:
- if: $CI_COMMIT_TAG
needs:
- job: update-nspect

mock-image-publish:
  extends:
    - .publish-images
  except:
    - tags
  needs:
    - job: update-nspect-staging

.nvidia-ci.yml
JunAr7112: Ok I can switch to "test". Sometime we should probably delete v1.0.0 from Staging repo

.nvidia-ci.yml
```diff
@@ -116,30 +116,60 @@ scan:gpu-operator-arm64:
...
-raise-publishing-merge-request:
-  stage: ngc-publish
+# Update the nspect production environment with the new release
+update-nspect:
+  extends:
+    - .update-nspect
   rules:
     - if: $CI_COMMIT_TAG
+  variables:
+    OSRB_BUG_ID: "${OSRB_ID}"
```
cdesiniotis: Question: Is this line necessary if we set `OSRB_BUG_ID` at the project level? I would prefer if this var was named `OSRB_BUG_ID` instead of `OSRB_ID` at the project level so that it is more clear what the intent of the variable is. If we rename, we can manually clear this variable in the staging job by setting `OSRB_BUG_ID=""` there.

cdesiniotis (commented):
One small comment about the naming of the OSRB bug variable. But otherwise this looks good! Thanks @JunAr7112 for the patience on this.

.nvidia-ci.yml
JunAr7112: Updated this variable to be "${OSRB_BUG_ID}" as well as at the project level. Updated the automation to verify whether ENV is stage or prod.

cdesiniotis (approved):

PR #1624, opened 2025-08-25
[dcgm-exporter] volume mount /run/nvidia if using hpc job files

heyitsanthony:
dcgm-exporter reads files from a user-defined directory to determine gpu-to-job mappings when hpc job mapping is enabled. Currently gpu-operator will only volume mount /run/nvidia for initialization, but not the long running dcgm-exporter pod, so it is not possible to configure a job mapping directory that maps back to the host.

This patch mounts /run/nvidia as readonly into the dcgm-exporter pod if hpc job mapping is enabled. This lets hpc job mappings be communicated from the host file system to dcgm-exporter via files in a user-defined directory under /run/nvidia.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


controllers/object_controls.go
```diff
@@ -1660,6 +1663,22 @@ func TransformDCGMExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
...
+	hasHPCMappingEnv := getContainerEnv(&obj.Spec.Template.Spec.Containers[0], DCGMHPCJobMappingDirEnvName) != ""
+	hasHPCMappingArg := slices.ContainsFunc(
+		obj.Spec.Template.Spec.Containers[0].Args,
+		func(s string) bool { return strings.HasPrefix(s, "--hpc-job-mapping-dir") })
+	if hasHPCMappingEnv || hasHPCMappingArg {
+		mode := corev1.MountPropagationHostToContainer
+		volMount := corev1.VolumeMount{
+			Name:             "run-nvidia",
+			ReadOnly:         true,
+			MountPath:        "/run/nvidia",
```
heyitsanthony, line 1676: NB: it would probably be better to use the hpc job dir with DirectoryOrCreate here since dcgm-exporter will log errors if the directory does not already exist.

heyitsanthony:
Fixed by #1894, closing

PR #1652, opened 2025-08-29
Allow resource requests and limits to be configurable for all containers in driver pod

shivakunv:
The resource requests / limits specified in the CRs only get configured with the main container and not any of the additional containers that can run as part of the driver pod. Below are the list of all other possible containers that can run in the driver pod:

OCP driver toolkit container
Runs only on OpenShift
NVIDIA peer-mem container
Runs only when `spec.driver.rdma.enabled=true` in the ClusterPolicy CR or `spec.rdma.enabled=true` in the NVIDIADriver CR
GDS driver container
Runs only when `spec.gds.enabled=true` in the ClusterPolicy CR or `spec.gds.enabled=true` in the NVIDIADriver CR
GDRCopy driver container
Runs only when `spec.gdrcopy.enabled=true` in the ClusterPolicy CR or `spec.gdrcopy.enabled=true `in the NVIDIADriver CR
The task is to update our controller code so that the resource requests / limits specified by users in the ClusterPolicy / NVIDIADriver CRs get applied to ALL containers in the driver pod, not just the main container.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -2789,6 +2789,12 @@ func transformPeerMemoryContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
 			}
 			obj.Spec.Template.Spec.Containers[i].VolumeMounts = append(obj.Spec.Template.Spec.Containers[i].VolumeMounts, volumeMounts...)
 		}
+		if config.Driver.Resources != nil {
+			obj.Spec.Template.Spec.Containers[i].Resources = corev1.ResourceRequirements{
+				Requests: config.Driver.Resources.Requests,
+				Limits:   config.Driver.Resources.Limits,
+			}
+		}
 	}
 	return nil
 }
@@ -2877,6 +2883,12 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 		if err != nil {
 			return fmt.Errorf("ERROR: failed to transform the Driver Toolkit Container: %s", err)
 		}
+		if config.Driver.Resources != nil {
+			gdsContainer.Resources = corev1.ResourceRequirements{
+				Requests: config.Driver.Resources.Requests,
+				Limits:   config.Driver.Resources.Limits,
+			}
+		}
 	}
 	return nil
 }
@@ -2963,6 +2975,12 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 		err = transformOpenShiftDriverToolkitContainer(obj, config, n, "nvidia-gdrcopy-ctr")
 		if err != nil {
 			return fmt.Errorf("ERROR: failed to transform the Driver Toolkit Container: %w", err)
+		}
+		if config.Driver.Resources != nil {
+			gdrcopyContainer.Resources = corev1.ResourceRequirements{
+				Requests: config.Driver.Resources.Requests,
+				Limits:   config.Driver.Resources.Limits,
+			}
 		}
 	}
 	return nil
@@ -3120,6 +3138,14 @@ func transformOpenShiftDriverToolkitContainer(obj *appsv1.DaemonSet, config *gpu
 		}
 	}
 	obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, volSharedDir)
+
+	// set resource limits
+	if config.Driver.Resources != nil {
+		driverToolkitContainer.Resources = corev1.ResourceRequirements{
+			Requests: config.Driver.Resources.Requests,
+			Limits:   config.Driver.Resources.Limits,
+		}
+	}
 	return nil
 }
 
--- a/internal/state/driver_test.go
+++ b/internal/state/driver_test.go
@@ -187,6 +187,10 @@ func TestDriverSpec(t *testing.T) {
 		ImagePullPolicy:  "Always",
 		ImagePullSecrets: []string{"secret-a", "secret-b"},
 		Resources: &nvidiav1alpha1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("200m"),
+				corev1.ResourceMemory: resource.MustParse("100Mi"),
+			},
 			Limits: corev1.ResourceList{
 				"memory": resource.MustParse("200Mi"),
 				"cpu":    resource.MustParse("500m"),
@@ -623,6 +627,16 @@ func getMinimalDriverRenderData() *driverRenderData {
 				LivenessProbe:  getDefaultContainerProbeSpec(),
 				ReadinessProbe: getDefaultContainerProbeSpec(),
 				DriverType:     nvidiav1alpha1.GPU,
+				Resources: &nvidiav1alpha1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("200m"),
+						corev1.ResourceMemory: resource.MustParse("100Mi"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("500m"),
+						corev1.ResourceMemory: resource.MustParse("300Mi"),
+					},
+				},
 			},
 			AppName:          "nvidia-gpu-driver-ubuntu22.04-7c6d7bd86b",
 			Name:             "nvidia-gpu-driver-ubuntu22.04",
--- a/manifests/state-driver/0500_daemonset.yaml
+++ b/manifests/state-driver/0500_daemonset.yaml
@@ -321,9 +321,8 @@ spec:
             mountPath: /etc/pki/ca-trust/extracted/pem
             readOnly: true
           {{- end}}
-        {{- with .Driver.Spec.Resources }}
-        resources:
-          {{ . | yaml | nindent 10 }}
+        {{- if .Driver.Spec.Resources }}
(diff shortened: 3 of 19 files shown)


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1282,6 +1282,11 @@ type GPUDirectRDMASpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Use MOFED drivers directly installed on the host to enable GPUDirect RDMA"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	UseHostMOFED *bool `json:"useHostMofed,omitempty"`
+	// Optional: Define resources requests and limits for each pod
```
```suggestion

	// Optional: Define resources requests and limits for each pod
```

controllers/object_controls.go
```diff
@@ -2796,6 +2796,12 @@ func transformPeerMemoryContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
 			}
 			obj.Spec.Template.Spec.Containers[i].VolumeMounts = append(obj.Spec.Template.Spec.Containers[i].VolumeMounts, volumeMounts...)
 		}
+		if config.Driver.Resources != nil {
```
tariq1890, line 2792: Shouldn't it be this?
```suggestion
		if config.Driver.GPUDirectRDMA.Resources != nil {
```

controllers/object_controls.go
```diff
@@ -2796,6 +2796,12 @@ func transformPeerMemoryContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
...
+			err := setResourceRequestsLimits(config.Driver.Resources, &obj.Spec.Template.Spec.Containers[i])
```
```suggestion
			err := setResourceRequestsLimits(config.Driver.GPUDirectRDMA.Resources, &obj.Spec.Template.Spec.Containers[i])
```

controllers/object_controls.go
```diff
@@ -2884,6 +2890,12 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 		if err != nil {
 			return fmt.Errorf("ERROR: failed to transform the Driver Toolkit Container: %s", err)
 		}
+		if config.Driver.Resources != nil {
```
tariq1890, line 2886:
```suggestion
		if config.Driver.GPUDirectStorage.Resources != nil {
```

controllers/object_controls.go
```diff
@@ -2884,6 +2890,12 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
...
+			err := setResourceRequestsLimits(config.Driver.Resources, gdsContainer)
```
```suggestion
			err := setResourceRequestsLimits(config.Driver.GPUDirectStorage.Resources, gdsContainer)
```

controllers/object_controls.go
```diff
@@ -2971,6 +2983,12 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 		if err != nil {
 			return fmt.Errorf("ERROR: failed to transform the Driver Toolkit Container: %w", err)
 		}
+		if config.Driver.Resources != nil {
```
tariq1890, line 2979:
```suggestion
		if config.Driver.GDRCopy.Resources != nil {
```

controllers/object_controls.go
```diff
@@ -2971,6 +2983,12 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
...
+			err := setResourceRequestsLimits(config.Driver.Resources, gdrcopyContainer)
```
```suggestion
			err := setResourceRequestsLimits(config.Driver.GDRCopy.Resources, gdrcopyContainer)
```

controllers/object_controls.go
```diff
@@ -3563,6 +3593,16 @@ func createSecretEnvReference(ctx context.Context, ctrlClient client.Client, sec
 	return nil
 }
 
+func setResourceRequestsLimits(resources *gpuv1.ResourceRequirements, container *corev1.Container) error {
```
tariq1890: It doesn't make a lot of sense to have a separate method to set resources to a container.

deployments/gpu-operator/templates/nvidiadriver.yaml
```diff
@@ -111,12 +114,15 @@ spec:
     {{- end }}
     {{- if .Values.gdrcopy.imagePullSecrets }}
     imagePullSecrets: {{ toYaml .Values.gdrcopy.imagePullSecrets | nindent 8 }}
-    {{- end }}
+    {{- end  }}
```
tariq1890: why add an extra space?

tariq1890:
Please add tests in `internal/state/driver_test.go` as well

controllers/object_controls.go
shivakunv: As mentioned in NVBug/Jira, we need to set resources from the driver  to the subresources.  
We can also add a condition to set resources for each driver container individually. If that’s not configured, it’ll just use the main driver container’s resources.

```
if config.Driver.GPUDirectRDMA.Resources != nil {
  err := setResourceRequestsLimits(config.Driver.GPUDirectRDMA.Resources, &obj.Spec.Template.Spec.Containers[i])
}
elif config.Driver.Resources != nil {
  err := setResourceRequestsLimits(config.Driver.Resources, &obj.Spec.Template.Spec.Containers[i])
}

```

controllers/object_controls.go
shivakunv: We are using the same set of code in different places for containers set resources with different names, so created a separate function to reduce repetitive code.

shivakunv:
> Please add tests in `internal/state/driver_test.go` as well

added

controllers/object_controls.go
tariq1890: you can reduce the method to one statement
```
conainer.Resources := &corev1.ResourceRequirements{
		Requests: resources.Requests,
		Limits:   resources.Limits,
	}
```

With this reduction, it does not make sense to create a separate method just for this

controllers/object_controls.go
tariq1890: If that is the case, why are you adding the `Resources` field to `GPUDirectStorageSpec`, `GPUDirectRDMASpec` etc? You are not using these fields at all. If the intent is to only uses the `Resources` field from `DriverSpec`, you don't need the `Resources` from the other structs

shivakunv: Thanks , modified accordingly . 

working fine. Test command: 

```
kubectl get pods -n nvidia-gpu-operator nvidia-driver-daemonset-dwh4j     -o yaml | yq '.spec.containers[] | select(.name | test("nvidia-gdrcopy-ctr"))
......
......
name: nvidia-gdrcopy-ctr
resources:
  limits:
    cpu: "2"
    memory: 20Gi
  requests:
    cpu: 500m
    memory: 1Gi
.............
............
```

shivakunv:
PTAL @tariq1890 @cdesiniotis

controllers/object_controls.go
```diff
@@ -3103,6 +3121,18 @@ func transformOpenShiftDriverToolkitContainer(obj *appsv1.DaemonSet, config *gpu
 		}
 	}
 	obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, volSharedDir)
+
+	// set resource limits
+	if config.Driver.Resources != nil {
+		mainContainer.Resources = corev1.ResourceRequirements{
+			Requests: config.Driver.Resources.Requests,
+			Limits:   config.Driver.Resources.Limits,
+		}
```
cdesiniotis: The function we are currently in, `transformOpenShiftDriverToolkitContainer()`, should only be modifying the driver toolkit container. 
```suggestion
```

Resource requests / limits are applied for the main container in the `transformDriverContainer` function here:
https://github.com/NVIDIA/gpu-operator/blob/d12142c281c74894879b43298e66f4b6171e7100/controllers/object_controls.go#L3289-L3293

manifests/state-driver/0500_daemonset.yaml
```diff
@@ -521,6 +525,10 @@ spec:
             {{- end }}
         {{- end }}
         {{- end }}
+        {{- with .Driver.Spec.Resources }}
```
cdesiniotis: I don't see these changes being added for the `nvidia-peermem` container or the `openshift-driver-toolkit-ctr` in this template.

tariq1890: These Resource blocks are needed within every container block

internal/state/driver_test.go
```diff
@@ -829,3 +829,51 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
 		require.Equal(t, test.expected, result)
 	}
 }
+
+func TestDriverResources(t *testing.T) {
```
tariq1890: We don't need a new test case for this change. We already have existing test cases for all the possible containers of the driver daemonset. Those existing tests should be updated to ensure that resource blocks are actually in those container specs when the driver spec resources are provided

manifests/state-driver/0500_daemonset.yaml
shivakunv: added

internal/state/driver_test.go
shivakunv: removed and added test case for openshift , precompiled, secret , licensing, vgpu host manager 
@tariq1890 @cdesiniotis

PR #1672, opened 2025-09-05
bump go to v1.25.1 and golangci-lint to v2.4.0

tariq1890:


versions.mk
```diff
@@ -19,8 +19,8 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= v25.3.2
 
-GOLANG_VERSION ?= 1.24.5
+GOLANG_VERSION ?= 1.25.1
 
-GOLANGCI_LINT_VERSION ?= v2.1.6
+GOLANGCI_LINT_VERSION ?= v2.4.0q
```
cdesiniotis: typo here

tariq1890: wow, thanks :facepalm:

cdesiniotis:
@tariq1890 do you want this backported to `release-25.3`?

tariq1890:
I was going to ask the same question. I think we should as 1.25.1 includes some security fixes

PR #1707, opened 2025-09-18
tests: Add unit tests for DCGM exporter Service and ServiceMonitor

karthikvetrivel:
This PR adds unit tests for the ServiceMonitor and Service controller functions, which Prometheus ServiceMonitor custom resources and Kubernetes Service resources for GPU telemetry components.

**Changes**

1. Added TestServiceMonitor to validate ServiceMonitor lifecycle management
2. Added TestService to validate Service resource management

**Test Coverage**

-  Correct state transitions based on component enablement
- Handling of missing ServiceMonitor CRD
- Resource creation, updates, and deletion
- Different component states (dcgm-exporter, node-status-exporter, operator-metrics)

Test coverage bump from 18.7% to 19.7%

### DIFF
--- a/controllers/object_controls_test.go
+++ b/controllers/object_controls_test.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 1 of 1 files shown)


controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		description       string
+		stateName         string
+		crdPresent        bool
+		dcgmEnabled       *bool
+		nodeStatusEnabled *bool
+		dcgmSMEnabled     *bool
+		withEdits         bool
+		wantState         gpuv1.State
+		extraAssert       assertsFn
+	}{
```
cdesiniotis: Instead of separate fields for the various ClusterPolicy options, why not just define one field that contains the ClusterPolicy spec for the test case? For example, what about reducing this to:
```suggestion
	testCases := []struct {
		description       string
		stateName         string
		crdPresent        bool
		cpSpec            gpuv1.ClusterPolicySpec
		wantState         gpuv1.State
		extraAssert       assertsFn
	}{
```
This would eliminate the need for the code you have later on that updates the ClusterPolicy object for each test case.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			}
+			if tc.nodeStatusEnabled != nil {
+				cp.Spec.NodeStatusExporter.Enabled = tc.nodeStatusEnabled
+			}
+			// Configure DCGM SM
+			if tc.stateName == "state-dcgm-exporter" {
+				if tc.dcgmSMEnabled != nil {
+					cp.Spec.DCGMExporter.ServiceMonitor = &gpuv1.DCGMExporterServiceMonitorConfig{Enabled: tc.dcgmSMEnabled}
+				}
+			}
```
cdesiniotis: As noted above, I think we could remove the need for this code if we just defined the ClusterPolicy spec for each test case.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			// If edits are requested, seed CP config for edits
+			if tc.withEdits {
+				cp.Spec.DCGMExporter.ServiceMonitor = &gpuv1.DCGMExporterServiceMonitorConfig{
+					Enabled:          truePtr,
+					Interval:         promv1.Duration("15s"),
+					HonorLabels:      truePtr,
+					AdditionalLabels: map[string]string{"a": "b"},
+					Relabelings:      []*promv1.RelabelConfig{{Action: "keep"}},
+				}
+			}
```
cdesiniotis: As noted above, I think we could remove the need for this code if we just embedded this configuration in the test case itself.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+
+			// Build the ServiceMonitor resource template
+			sm := promv1.ServiceMonitor{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-sm", Labels: map[string]string{}},
+				Spec: promv1.ServiceMonitorSpec{
+					NamespaceSelector: promv1.NamespaceSelector{MatchNames: []string{"FILLED BY THE OPERATOR"}},
+					Endpoints:         []promv1.Endpoint{{}},
+				},
+			}
```
cdesiniotis: nit -- since this ServiceMonitor template does not differ between test cases, do we want to define this variable outside of the for loop?

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			// Build fake client, optionally seed CRD existence by registering type only
+			b := fake.NewClientBuilder().WithScheme(scheme)
+			if tc.crdPresent {
+				crd := &apiextensionsv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: ServiceMonitorCRDName}}
+				b = b.WithObjects(crd)
+			}
+			k8sClient := b.Build()
```
cdesiniotis: Not critical, but we could consider adding a `client` field for each test case and building the fake client in-line for each test case. For example, for test cases where the ServiceMonitor CRD does not exist we would have:
```
client: fake.NewClientBuilder().WithScheme(scheme).Build()
```
while for the test cases where the ServiceMonitor CRD does exist we would have:
```
client: fake.NewClientBuilder().WithScheme(scheme).WithObjects(serviceMonitorCRD).Build()
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+				found := &promv1.ServiceMonitor{}
+				err := c.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: name}, found)
+				require.NoError(t, err)
+				require.Equal(t, promv1.Duration("15s"), found.Spec.Endpoints[0].Interval)
+				require.Equal(t, true, found.Spec.Endpoints[0].HonorLabels)
+				require.Equal(t, "b", found.Labels["a"])
+				require.NotNil(t, found.Spec.Endpoints[0].RelabelConfigs)
+				require.Equal(t, 1, len(found.Spec.Endpoints[0].RelabelConfigs))
+			},
+		},
```
cdesiniotis: Instead of defining custom assert logic here, why don't we define the expected ServiceMonitor object here and compare it to the actual object that gets created when executing the test case?

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+				logger:            ctrl.Log.WithName("test"),
+			}
+
+			state, err := ServiceMonitor(controller)
+			require.NoError(t, err)
+			require.Equal(t, tc.wantState, state)
+
+			if tc.extraAssert != nil {
+				tc.extraAssert(t, k8sClient, "test-sm", "test-ns")
+			}
```
cdesiniotis: As suggested in a prior comment, for test cases where we expect a ServiceMonitor object to be created, would it be possible to just compare the two service monitor objects (expected vs actual)?

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		},
+		{
+			description:   "dcgm-exporter SM enabled, CRD present -> Ready and applies edits",
+			stateName:     "state-dcgm-exporter",
+			crdPresent:    true,
+			dcgmEnabled:   truePtr,
+			dcgmSMEnabled: truePtr,
+			withEdits:     true,
+			wantState:     gpuv1.Ready,
+			extraAssert: func(t *testing.T, c client.Client, name, ns string) {
```
tariq1890: Instead of defining a closure to define additional assert blocks, can we just define a separate test case to assert the extra fields? That would make it easier to read.

Personally speaking, I'd avoid using closures and just sticking to more procedural coding. It makes test cases easier to read

controllers/object_controls_test.go
```diff
@@ -1083,3 +1083,288 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			description: "create and preprocess",
+			enabled:     truePtr,
+			serviceCfg: &gpuv1.DCGMExporterServiceConfig{
+				Type:                  corev1.ServiceTypeNodePort,
+				InternalTrafficPolicy: &localPolicy,
+			},
+			wantState: gpuv1.Ready,
+			extraAssert: func(t *testing.T, c client.Client, name, ns string) {
+				found := &corev1.Service{}
+				err := c.Get(context.TODO(), client.ObjectKey{Namespace: ns, Name: name}, found)
```
tariq1890: We only use `context.TODO()` when we intend to pass in a ctx object later on. Since this most likely isn't a TODO, I would suggest using `context.Background()` instead

karthikvetrivel:
Hi team,

I've pushed a new commit addressing the review feedback. More broadly, I've made the test more modular, simpler, and removed all assert blocks. Functionality remains the same. Please take another look when you get a chance.

Thanks!

cdesiniotis:
@karthikvetrivel thanks. To help out the reviewers, can you respond to the individual review comments? That helps us know which ones have been addressed and which ones not.

tariq1890:
nit - We aren't really adding a new feature here, so we should remove the `feat(test)` from the commit message and PR title

cdesiniotis:
Also, please sign your commits. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

controllers/object_controls_test.go
karthikvetrivel: I shifted away from using asserts in the new test.

controllers/object_controls_test.go
karthikvetrivel: Good point. I switched to context.Background() in the new commit.

controllers/object_controls_test.go
karthikvetrivel: Done, I pass in ClusterPolicy spec into each test case now.

controllers/object_controls_test.go
karthikvetrivel: Updated to use an expected ServiceMonitor object and compare it.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Yep! Great idea. I pass a client field for each test case now.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Yep, I embed the configuration now in the test case. Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated to pass in a ClusterPolicy spec for each test case.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
 		require.Equal(t, test.expected, result)
 	}
 }
+
+func TestServiceMonitor(t *testing.T) {
+	const (
+		testNS   = "test-ns"
```
tariq1890:
```suggestion
		testNamespace   = "test-namespace"
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		testSM   = "test-sm"
```
tariq1890:
```suggestion
		testServiceMonitor   = "test-service-monitor"
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+	// Create controller with given spec and state
+	newController := func(k8s client.Client, scheme *runtime.Scheme, spec gpuv1.ClusterPolicySpec, state string) ClusterPolicyController {
+		cp := &gpuv1.ClusterPolicy{Spec: spec}
+		return ClusterPolicyController{
+			client:            k8s,
+			ctx:               context.Background(),
+			singleton:         cp,
+			scheme:            scheme,
+			operatorNamespace: testNS,
+			resources:         []Resources{{ServiceMonitor: smTemplate}},
```
tariq1890: move the intialisation fo this slice to its own statement

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		cpSpec      gpuv1.ClusterPolicySpec
+		wantState   gpuv1.State
+		expectedSM  *promv1.ServiceMonitor
+	}{
+		{
+			description: "dcgm-exporter disabled, CRD missing -> Ready",
+			stateName:   "state-dcgm-exporter",
+			client:      fake.NewClientBuilder().WithScheme(scheme).Build(),
+			cpSpec: gpuv1.ClusterPolicySpec{
+				DCGMExporter: gpuv1.DCGMExporterSpec{Enabled: ptr.To(false)},
```
tariq1890, line 1143: Great use of `ptr.To` here!

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+	// CRD object for tests that need ServiceMonitor CRD present
+	serviceMonitorCRD := &apiextensionsv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: ServiceMonitorCRDName}}
+
+	tests := []struct {
+		description string
+		stateName   string
+		client      client.Client
+		cpSpec      gpuv1.ClusterPolicySpec
+		wantState   gpuv1.State
+		expectedSM  *promv1.ServiceMonitor
```
tariq1890:
```suggestion
		expectedServiceMonitor  *promv1.ServiceMonitor
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+	}{
+		{
+			description: "dcgm-exporter disabled, CRD missing -> Ready",
+			stateName:   "state-dcgm-exporter",
+			client:      fake.NewClientBuilder().WithScheme(scheme).Build(),
```
tariq1890: why do we need an instance of the `client` per test case? I think we can just instantiate a client once and reuse it in all the test-cases

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
 		require.Equal(t, test.expected, result)
 	}
 }
+
+func TestServiceMonitor(t *testing.T) {
+	const (
+		testNS   = "test-ns"
+		testSM   = "test-sm"
+		filledNS = "FILLED BY THE OPERATOR"
```
tariq1890:
```suggestion
		filledNamespace = "FILLED BY THE OPERATOR"
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			operatorNamespace: testNS,
+			resources:         []Resources{{ServiceMonitor: smTemplate}},
+			stateNames:        []string{state},
+			idx:               0,
+			logger:            ctrl.Log.WithName("test"),
+		}
+	}
+
+	// Retrieve ServiceMonitor from client
+	getSM := func(k8s client.Client) (*promv1.ServiceMonitor, error) {
```
tariq1890: I don't think we need this function closure. Let's just invoke these lines directly instead of using the function var `getSM`

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+
+	// CRD object for tests that need ServiceMonitor CRD present
+	serviceMonitorCRD := &apiextensionsv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: ServiceMonitorCRDName}}
+
+	tests := []struct {
+		description string
+		stateName   string
+		client      client.Client
+		cpSpec      gpuv1.ClusterPolicySpec
+		wantState   gpuv1.State
```
tariq1890:
```suggestion
		expectedState   gpuv1.State
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{Name: testSM, Labels: map[string]string{}},
+		Spec: promv1.ServiceMonitorSpec{
+			NamespaceSelector: promv1.NamespaceSelector{MatchNames: []string{filledNS}},
+			Endpoints:         []promv1.Endpoint{{}},
+		},
+	}
+
+	// Create controller with given spec and state
+	newController := func(k8s client.Client, scheme *runtime.Scheme, spec gpuv1.ClusterPolicySpec, state string) ClusterPolicyController {
+		cp := &gpuv1.ClusterPolicy{Spec: spec}
```
tariq1890:
```suggestion
		clusterPolicy := &gpuv1.ClusterPolicy{Spec: spec}
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			require.Equal(t, tc.expectedSM.Namespace, found.Namespace)
+			require.Equal(t, tc.expectedSM.Labels, found.Labels)
+			require.Equal(t, tc.expectedSM.Spec, found.Spec)
+		})
+	}
+}
+
+func TestService(t *testing.T) {
+	const (
+		testNS      = "test-ns"
```
tariq1890:
```suggestion
		testNamespace      = "test-namespace"
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			operatorNamespace: testNS,
+			resources:         []Resources{{Service: svcTemplate}},
+			stateNames:        []string{"state-dcgm-exporter"},
+			idx:               0,
+			logger:            ctrl.Log.WithName("test"),
+		}
+	}
+
+	// Helper to retrieve Service from client
+	getService := func(t *testing.T, k8s client.Client) (*corev1.Service, error) {
```
tariq1890: similar comment as getSM()

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+
+	localPolicy := corev1.ServiceInternalTrafficPolicyLocal
+
+	tests := []struct {
+		description   string
+		client        client.Client
+		cpSpec        gpuv1.ClusterPolicySpec
+		wantState     gpuv1.State
+		expectService bool
+		expectType    corev1.ServiceType
```
tariq1890:
```suggestion
		expectedType    corev1.ServiceType
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		expectPolicy  *corev1.ServiceInternalTrafficPolicy
```
tariq1890:
```suggestion
		expectedPolicy  *corev1.ServiceInternalTrafficPolicy
```

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		expectIP      string // For ClusterIP preservation test
```
tariq1890:
```suggestion
		expectedIP      string // For ClusterIP preservation test
```

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated!

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: I could be wrong, but I thought that the controller-runtime fake client stores objects in-memory. This means that one tests creating a ServiceMonitor would leak into the next if you reuse the client. 

Test 1 creates a `ServiceMonitor`; Test 2 expects “none exists” (asserts NotFound). If you reuse the client, Test 2 now finds the object and fails. I created a mock test to validate this idea and found this to be the case

`--- FAIL: TestServiceMonitor_SharedClient_StateLeak (0.00s)
    --- PASS: TestServiceMonitor_SharedClient_StateLeak/first_creates_ServiceMonitor (0.00s)
    --- FAIL: TestServiceMonitor_SharedClient_StateLeak/second_expects_NotFound_but_leaks_from_first (0.00s)`

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Great point, updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: `cp` was a convention used earlier in `object_controls_test.go` when created to pass into a ClusterPolicyController. 

https://github.com/NVIDIA/gpu-operator/blob/c88ce54070f8c3e56c4a3c8578c6c20f49a02ace/controllers/object_controls_test.go#L197

Since it's passed into `ClusterPolicyController` immediately after, I don't feel strongly about changing it. However, if we do, we should change all instances in this file.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,301 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			if tc.expectedServiceMonitor == nil {
+				require.True(t, apierrors.IsNotFound(err))
+				return
+			}
+			require.NoError(t, err)
+
+			require.Equal(t, tc.expectedServiceMonitor.Name, found.Name)
+			require.Equal(t, tc.expectedServiceMonitor.Namespace, found.Namespace)
+			require.Equal(t, tc.expectedServiceMonitor.Labels, found.Labels)
+			require.Equal(t, tc.expectedServiceMonitor.Spec, found.Spec)
```
cdesiniotis, line 1255: Question -- could we not replace this with the below?
```suggestion
			require.Equal(t, tc.expectedServiceMonitor, found)
```

cdesiniotis:
@karthikvetrivel please squash your commits into one commit.

controllers/object_controls_test.go
tariq1890: I understand that `cp` was used before this PR, but we should use PRs as an opportunity to improve our coding standards wherever possible. I don't have a strong opinion on changing every instance (could be done either or in a follow-up PR dedicated to minor code-cleanups), but it would be better to use `clusterPolicy` instead of `cp` in the code we are introducing

controllers/object_controls_test.go
tariq1890: Understood. If the in-memory object list is the only variable here, we should make that the struct field rather than the whole fake client object

controllers/object_controls_test.go
karthikvetrivel: It fails the test then. The created object has controller-set fields, API-managed metadata (e.g., ResourceVersion, possibly defaults), and other labels. A deep equality on the whole object would require accounting for all of those. 

I think it's sufficient to target assertions on just the fields we control.

karthikvetrivel:
@cdesiniotis Done! Squashed into a single commit.

controllers/object_controls_test.go
cdesiniotis: Ah right, make sense.

controllers/object_controls_test.go
karthikvetrivel: I agree 100%. I can lead a small follow-up PR for minor code-cleanups like this because there are many such abbreviations. For example, `ds` for `DaemonSet` pops up with `object_controls` often.

controllers/object_controls_test.go
karthikvetrivel: I see the vision. I implemented an `existingObjects []client.Object` field in the struct and moved the fake client object to the test cases. 

Let me know if you have any other questions/thoughts for the PR.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+			idx:               0,
+			logger:            ctrl.Log.WithName("test"),
+		}
+	}
+
+	localPolicy := corev1.ServiceInternalTrafficPolicyLocal
+
+	tests := []struct {
+		description     string
+		existingObjects []client.Object
```
tariq1890: Let's rename this to `k8sObjects`, `existingObjects` is vague

controllers/object_controls_test.go
tariq1890: Yes, please create the follow-up PR to change the var names in the other areas, but I ask that you rename `cp` to `clusterPolicy` in the code blocks you've added

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+	}
+
+	// CRD object for tests that need ServiceMonitor CRD present
+	serviceMonitorCRD := &apiextensionsv1.CustomResourceDefinition{ObjectMeta: metav1.ObjectMeta{Name: ServiceMonitorCRDName}}
+
+	tests := []struct {
+		description            string
+		stateName              string
+		existingObjects        []client.Object
+		cpSpec                 gpuv1.ClusterPolicySpec
```
tariq1890:
```suggestion
		clusterPolicySpec                 gpuv1.ClusterPolicySpec
```

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
karthikvetrivel: Updated.

controllers/object_controls_test.go
```diff
@@ -1083,3 +1084,311 @@ func TestGetSanitizedKernelVersion(t *testing.T) {
...
+		filledNamespace    = "FILLED BY THE OPERATOR"
+	)
+
+	// Create scheme with required types
+	scheme := runtime.NewScheme()
+	require.NoError(t, promv1.AddToScheme(scheme))
+	require.NoError(t, apiextensionsv1.AddToScheme(scheme))
+	require.NoError(t, gpuv1.AddToScheme(scheme))
+
+	smTemplate := promv1.ServiceMonitor{
```
tariq1890: Is this object used as a template? If not I'd just rename it to `serviceMonitor`

karthikvetrivel: Good catch! Yes, it was previously a template but no longer is. Updated.

controllers/object_controls_test.go
tariq1890: thanks

karthikvetrivel:
@tariq1890 Pushed with all of your naming edits.

tariq1890:
Screenshot 2025-09-22 at 4 36 48 PM

Not sure why the coveralls-bot hasn't commented, but I see in coveralls.io that the coverage is increased by 0.6%

PR #1710, opened 2025-09-19
Add support for drop-in configs

elezar:
This change adds explicit support for drop-in configs as supported by containerd and cri-o.

See:
* https://github.com/NVIDIA/nvidia-container-toolkit/pull/1280
* https://github.com/NVIDIA/nvidia-container-toolkit/pull/1303

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -57,16 +57,18 @@ import (
 const (
 	// DefaultContainerdConfigFile indicates default config file path for containerd
 	DefaultContainerdConfigFile = "/etc/containerd/config.toml"
+	// DefaultContainerdDropInConfigFile indicates default drop-in config file path for containerd
+	DefaultContainerdDropInConfigFile = "/etc/containerd/conf.d/99-nvidia.toml"
 	// DefaultContainerdSocketFile indicates default containerd socket file
 	DefaultContainerdSocketFile = "/run/containerd/containerd.sock"
 	// DefaultDockerConfigFile indicates default config file path for docker
 	DefaultDockerConfigFile = "/etc/docker/daemon.json"
 	// DefaultDockerSocketFile indicates default docker socket file
 	DefaultDockerSocketFile = "/var/run/docker.sock"
-	// DefaultCRIOConfigFile indicates default config file path for cri-o.
-	// Note, config files in the drop-in directory, /etc/crio/crio.conf.d,
-	// have a higher priority than the default /etc/crio/crio.conf file.
-	DefaultCRIOConfigFile = "/etc/crio/crio.conf.d/99-nvidia.conf"
+	// DefaultCRIOConfigFile indicates default config file path for cri-o. .
+	DefaultCRIOConfigFile = "/etc/crio/config.toml"
+	// DefaultCRIODropInConfigFile indicates the default path to the drop-in config file for cri-o
+	DefaultCRIODropInConfigFile = "/etc/crio/crio.conf.d/99-nvidia.conf"
 	// TrustedCAConfigMapName indicates configmap with custom user CA injected
 	TrustedCAConfigMapName = "gpu-operator-trusted-ca"
 	// TrustedCABundleFileName indicates custom user ca certificate filename
@@ -95,6 +97,8 @@ const (
 	DefaultRuntimeSocketTargetDir = "/runtime/sock-dir/"
 	// DefaultRuntimeConfigTargetDir represents target directory where runtime socket dirctory will be mounted
 	DefaultRuntimeConfigTargetDir = "/runtime/config-dir/"
+	// DefaultRuntimeDropInConfigTargetDir represents target directory where drop-in config directory will be mounted
+	DefaultRuntimeDropInConfigTargetDir = "/runtime/config-dir.d/"
 	// ValidatorImageEnvName indicates env name for validator image passed
 	ValidatorImageEnvName = "VALIDATOR_IMAGE"
 	// ValidatorImagePullPolicyEnvName indicates env name for validator image pull policy passed
@@ -1355,12 +1359,18 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		setContainerEnv(mainContainer, CRIOConfigModeEnvName, "config")
 	}
 
+	// For runtime config files we have top-level configs and drop-in files.
+	// These are supported as follows:
+	//   * Docker only supports top-level config files.
+	//   * Containerd supports drop-in files, but required modification to the top-level config
+	//   * Crio supports drop-in files at a predefined location. The top-level config may be read
+	//     but should not be updated.
+
 	// setup mounts for runtime config file
-	runtimeConfigFile, err := getRuntimeConfigFile(mainContainer, runtime)
+	topLevelConfigFile, dropInConfigFile, err := getRuntimeConfigFiles(mainContainer, runtime)
 	if err != nil {
-		return fmt.Errorf("error getting path to runtime config file: %v", err)
+		return fmt.Errorf("error getting path to runtime config file: %w", err)
 	}
-	sourceConfigFileName := path.Base(runtimeConfigFile)
 
 	var configEnvvarName string
 	switch runtime {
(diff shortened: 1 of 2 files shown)


controllers/object_controls.go
```diff
@@ -1344,15 +1350,41 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+		mainContainer.VolumeMounts = append(mainContainer.VolumeMounts, volMountConfig)
+
+		configVol := corev1.Volume{Name: volMountConfigName, VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: sourceConfigDir, Type: newHostPathType(corev1.HostPathDirectoryOrCreate)}}}
+		obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, configVol)
+	}
+
+	// Handle the drop-in configs
+	if runtimeConfigFiles.dropInConfigFile != "" {
+		sourceConfigFileName := path.Base(runtimeConfigFiles.dropInConfigFile)
+		sourceConfigDir := path.Dir(runtimeConfigFiles.dropInConfigFile)
```
ArangoGutierrez: Why sourceConfig* if these are the drop-in file/folder names, shouldn't this be `dropinConfigDir` and `dropinConfigFileName` ?

elezar: This doesn't actually matter. The variables are scoped to the conditional block. This also calls out that we may have some comonality here so that we *could* factor this out into a function.

elezar:
Please note. This is a very quick first draft.

controllers/object_controls.go
```diff
@@ -1344,15 +1350,44 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+	// Note that we probably want to implement drop-in file support in the
+	// kata manager in any case -- in which case it will be good to use a
+	// similar implementation.
+	if runtimeConfigFiles.dropInConfigFile != "" && containerName != "nvidia-kata-manager" {
+		sourceConfigFileName := path.Base(runtimeConfigFiles.dropInConfigFile)
+		sourceConfigDir := path.Dir(runtimeConfigFiles.dropInConfigFile)
+		// TODO: This should be a constant.
+		containerConfigDir := "/runtime/config-dir.d/"
+		setContainerEnv(mainContainer, "RUNTIME_DROP_IN_CONFIG", containerConfigDir+sourceConfigFileName)
+		setContainerEnv(mainContainer, "RUNTIME_DROP_IN_CONFIG_HOST_PATH", runtimeConfigFiles.dropInConfigFile)
```
cdesiniotis: Is `RUNTIME_DROP_IN_CONFIG_HOST_PATH` an actual envvar that the toolkit container processes? I did a quick search and can't find any mention of this variable. As a consequence, when testing out this PR I see that the toolkit container is setting 
```
imports = ["/runtime/config-dir.d/*.toml"]
```
in the top-level containerd config instead of 
```
imports = ["/run/toolkit/config/*.toml"]
```

controllers/transforms_test.go
```diff
@@ -647,15 +647,19 @@ func TestTransformToolkit(t *testing.T) {
 						{Name: "CONTAINERD_RUNTIME_CLASS", Value: "nvidia"},
 						{Name: "RUNTIME_CONFIG", Value: "/runtime/config-dir/config.toml"},
 						{Name: "CONTAINERD_CONFIG", Value: "/runtime/config-dir/config.toml"},
+						{Name: "RUNTIME_DROP_IN_CONFIG", Value: "/runtime/config-dir.d/99-nvidia.toml"},
```
cdesiniotis, line 668: Corresponding test cases need to be added to `TestTransformForRuntime()` as well.

controllers/object_controls.go
cdesiniotis: Ah nevermind, I see this is being added in https://github.com/NVIDIA/nvidia-container-toolkit/pull/1303.

controllers/transforms_test.go
cdesiniotis: I have pushed a commit for this.

controllers/object_controls.go
```diff
@@ -2357,30 +2392,62 @@ func TransformNodeStatusExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPol
...
 	case gpuv1.Containerd.String():
-		runtimeConfigFile = DefaultContainerdConfigFile
+		topLevelConfigFile := DefaultContainerdConfigFile
+		// TODO: We should also read RUNTIME_CONFIG here
 		if value := getContainerEnv(c, "CONTAINERD_CONFIG"); value != "" {
-			runtimeConfigFile = value
+			topLevelConfigFile = value
+		}
+		// TODO: This should be a sane default.
+		dropInConfigFile := "/run/toolkit/config/99-nvidia.toml"
```
cdesiniotis: Question -- what about `/run/nvidia/toolkit/config/99-nvidia.toml`? The toolkit container creates the `toolkit.pid` file at  `/run/nvidia/toolkit/toolkit.pid`

controllers/object_controls.go
```diff
@@ -2357,30 +2392,62 @@ func TransformNodeStatusExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPol
...
 		}
+		return runtimeConfigFiles{
+			topLevelConfigFile: topLevelConfigFile,
+			// Docker does not support drop-in files.
+			dropInConfigFile: "",
+		}, nil
 	case gpuv1.Containerd.String():
-		runtimeConfigFile = DefaultContainerdConfigFile
+		topLevelConfigFile := DefaultContainerdConfigFile
+		// TODO: We should also read RUNTIME_CONFIG here
```
cdesiniotis: Ack. Should `RUNTIME_CONFIG`, if defined, take precedence over `CONTAINERD_CONFIG`/`CRIO_CONFIG`?

controllers/object_controls.go
```diff
@@ -2357,30 +2392,62 @@ func TransformNodeStatusExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPol
...
+			dropInConfigFile = value
 		}
+		return runtimeConfigFiles{
+			topLevelConfigFile: topLevelConfigFile,
+			dropInConfigFile:   dropInConfigFile,
+		}, nil
 	case gpuv1.CRIO.String():
-		runtimeConfigFile = DefaultCRIOConfigFile
+		// TODO: We should still allow the top-level config to be specified
+		// TODO: This should be DefaultCRIOConfigFile, but this is not the correct value.
```
cdesiniotis: Ack. We should update the value of `DefaultCRIOConfigFile` to `/etc/crio/config.toml` and define a separate variable for defining the default path for the drop-in file.

controllers/object_controls.go
cdesiniotis: I have updated this to `/run/nvidia/toolkit/config/99-nvidia.toml`.

controllers/object_controls.go

tariq1890:
Why can't we mount the parent dir of the top-level config and the drop-in config dir?

It seems like a lot of the container-toolkit implementation details are spilling into this PR? Any way to minimise?

cdesiniotis:
> Why can't we mount the parent dir of the top-level config and the drop-in config dir?
> 
> It seems like a lot of the container-toolkit implementation details are spilling into this PR? Any way to minimise?

We are mounting the parent dir for both config files.

controllers/object_controls.go
```diff
@@ -2396,30 +2424,60 @@ func TransformNodeStatusExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPol
 	return nil
 }
 
+type runtimeConfigFiles struct {
```
tariq1890: I don't think we need this struct. The "topLevelConfigFile" and "dropInConfigFile" can be the return arguments, whereas the `envvarname` can be resolved separately

cdesiniotis: That is fair. I have updated this and incorporated your feedback.

controllers/object_controls.go
cdesiniotis: @tariq1890 and I discussed this offline. I have updated the host path to `/etc/containerd/conf.d/99-nvidia.toml`

controllers/object_controls.go
tariq1890: I think the latter should have higher precedence as it is more specific than `RUNTIME_CONFIG`. If you're in agreement with this, can we resolve the TODO in this PR?

cdesiniotis: Sure. Let me update this...

cdesiniotis: Updated so we now read `RUNTIME_CONFIG` as well, with the more specific envvars, e.g. `CONTAINERD_CONFIG` / `CRIO_CONFIG`, taking precedence. I also added additional unit test cases for this.

tariq1890: Thanks!

tariq1890:
> Coverage increased (+0.3%) to 21.388%

🚀 🚀 🚀

PR #1724, opened 2025-09-23
bump DCGM to 4.4.1-2 and DCGM exporter to 4.4.1-4.5.2

JunAr7112:


deployments/gpu-operator/values.yaml
```diff
@@ -289,7 +289,7 @@ dcgmExporter:
   enabled: true
   repository: nvcr.io/nvidia/k8s
   image: dcgm-exporter
-  version: 4.3.1-4.4.0-ubuntu22.04
+  version: 4.4.1-4.5.2-ubuntu22.04
```
cdesiniotis: Can you add an additional commit to this PR that switches to using the distroless tag for the dcgm-exporter image? The tag here needs to be updated as well as the digests in the OLM bundle. cc @tariq1890 
```suggestion
  version: 4.4.1-4.5.2-distroless
```

JunAr7112: Alright I added a commit for distroless and updated the manifests

cdesiniotis: Can you update the commit message to indicate that we are switching `dcgm-exporter` to distroless, and not dcgm?

JunAr7112: Sure

PR #1728, opened 2025-09-24
[stage-mig-manager] update container command & args to support distroless

tariq1890:
This change is needed to resolve the follow container crashloop error:

```
/bin/sh: /bin/entrypoint.sh: not found
```


assets/state-mig-manager/0600_daemonset.yaml
```diff
@@ -39,9 +39,8 @@ spec:
       - name: nvidia-mig-manager
         image: "FILLED BY THE OPERATOR"
         imagePullPolicy: IfNotPresent
-        command: [/bin/sh, -c]
-        args:
-          - /bin/entrypoint.sh
+        command: ["/bin/sh"]
```
cdesiniotis, line 42: I am curious why `/bin/sh -c` works for other operands but not mig-manager? I made similar changes across all the components in https://github.com/NVIDIA/gpu-operator/pull/1496.

An alternative that should also work is `sh -c`

tariq1890: I tried `sh -c` , that didn't work either. I don't quite understand why I get the `not found` error. It seems like busybox treats the `-c` flag a bit differently

cdesiniotis: Before this change, what is the error message you are getting?

cdesiniotis: Oh you already put it in the PR description, my bad...

PR #1732, opened 2025-09-26
Refactor transformForRuntime to accept container pointer; add kata-manager test

karthikvetrivel:
- Change signature to `transformForRuntime(obj, config, runtime, *corev1.Container)` to avoid name-based lookup.
- Update callers in controllers/object_controls.go (Toolkit, Kata manager) to pass the resolved container pointer.
- Adjust existing tests to new signature.
- Add test ensuring `nvidia-kata-manager` skips drop-in config while setting top-level config and socket.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1327,7 +1327,7 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n 
 
 	// configure runtime
 	runtime := n.runtime.String()
-	err = transformForRuntime(obj, config, runtime, mainContainerName)
+	err = transformForRuntime(obj, config, runtime, mainContainer)
 	if err != nil {
 		return fmt.Errorf("error transforming toolkit daemonset : %w", err)
 	}
@@ -1335,16 +1335,10 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n 
 	return nil
 }
 
-func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, containerName string) error {
-	var mainContainer *corev1.Container
-	for i, ctr := range obj.Spec.Template.Spec.Containers {
-		if ctr.Name == containerName {
-			mainContainer = &obj.Spec.Template.Spec.Containers[i]
-			break
-		}
-	}
+func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, container *corev1.Container) error {
+	mainContainer := container
 	if mainContainer == nil {
-		return fmt.Errorf("failed to find main container %q", containerName)
+		return fmt.Errorf("failed to find container for runtime transform")
 	}
 
 	setContainerEnv(mainContainer, "RUNTIME", runtime)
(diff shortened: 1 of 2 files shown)


controllers/object_controls.go
```diff
@@ -1327,24 +1327,18 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
-func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, containerName string) error {
-	var mainContainer *corev1.Container
-	for i, ctr := range obj.Spec.Template.Spec.Containers {
-		if ctr.Name == containerName {
-			mainContainer = &obj.Spec.Template.Spec.Containers[i]
-			break
-		}
-	}
+func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, container *corev1.Container) error {
+	mainContainer := container
```
cdesiniotis: Let's just refer to `container` throughout this function.

controllers/object_controls.go
```diff
@@ -1327,24 +1327,18 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
 	if mainContainer == nil {
-		return fmt.Errorf("failed to find main container %q", containerName)
+		return fmt.Errorf("failed to find container for runtime transform")
```
cdesiniotis:
```suggestion
		return fmt.Errorf("no container to transform")
```

controllers/object_controls.go
karthikvetrivel: Updated.

controllers/object_controls.go
karthikvetrivel: Updated.

karthikvetrivel:
@cdesiniotis I implemented the changes. Let me know if this looks good and I'll squash the commits. 

EDIT: Also fixed linter issue.

tariq1890:
Looks like Coverage is decreased here. Can you add a test case to maintain/increase the unit test coverage?

controllers/object_controls.go
```diff
@@ -1327,36 +1327,29 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
-		if ctr.Name == containerName {
-			mainContainer = &obj.Spec.Template.Spec.Containers[i]
-			break
-		}
-	}
-	if mainContainer == nil {
-		return fmt.Errorf("failed to find main container %q", containerName)
+func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, container *corev1.Container) error {
+	if container == nil {
+		return fmt.Errorf("no container to transform")
```
```suggestion
		return errors.New("no container to transform")
```

tariq1890: Actually, I am not too sure if we even need this if-conditional block. L1256 already performs this check

karthikvetrivel: Agreed! I removed the `container == nil` guard in transformForRuntime since all current call sites ensure a non-nil container.

karthikvetrivel:
@tariq1890 @cdesiniotis I implemented the recent comments (added test & removed conditional) and squashed the commits. 

Please let me know if you have any other questions!

tariq1890 (approved):

PR #1742, opened 2025-09-30
Fix bug where user-supplied envvars are not applied for dcgm-exporter

cdesiniotis:
This fixes a regression introduced in https://github.com/NVIDIA/gpu-operator/commit/54c270cf3f91c39fb1d0b8e2597f389b70975228 where user-supplied environment variables are not honored for the DCGM-Exporter daemonset on non-OCP environments.


controllers/object_controls.go
```diff
@@ -1729,6 +1729,12 @@ func TransformDCGMExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), "DCGM_EXPORTER_COLLECTORS", MetricsConfigMountPath)
 	}
 
+	if len(config.DCGMExporter.Env) > 0 {
+		for _, env := range config.DCGMExporter.Env {
+			setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
+		}
+	}
```
elezar: Nit: the length check is redundant since a loop over an empty slice is always a no-op:

```suggestion
	for _, env := range config.DCGMExporter.Env {
		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), env.Name, env.Value)
	}
```

controllers/object_controls.go
elezar: It took me a while to see that this `return` was what was causing the environment to not be applied. Does it make sense to factor the required changes into a function and conditionally call that so that we can still update the envs at the end of the function.

One minor concern is that we may need to make changes to the actual container for Openshift at some point, and then we could end up in a situation where user-supplied input no longer overrides the "automatic" environment variables.

controllers/transforms_test.go
```diff
@@ -915,6 +915,7 @@ func TestTransformDCGMExporter(t *testing.T) {
 				Args:            []string{"--fail-on-init-error=false"},
 				Env: []corev1.EnvVar{
 					{Name: "DCGM_REMOTE_HOSTENGINE_INFO", Value: "nvidia-dcgm:5555"},
+					{Name: "foo", Value: "bar"},
```
elezar, line 1055: Are we able to test the OpenShift case too? (probably out-of-scope for this PR).

controllers/object_controls.go
cdesiniotis: Updated as you have suggested.

controllers/object_controls.go
cdesiniotis: Yes definitely makes sense. I have refactored this so that the openshift logic is contained in its own function.

controllers/transforms_test.go
cdesiniotis: Added a test case for openshift.

PR #1744, opened 2025-10-01
[Dockerfile] use the official golang image in the build stage

tariq1890:


docker/Dockerfile
```diff
@@ -14,25 +14,7 @@
...
-        ppc64el | ppc64le) ARCH='ppc64le' ;; \
-        aarch64 | arm64) ARCH='arm64' ;; \
-        *) echo "unsupported architecture" ; exit 1 ;; \
-    esac; \
-    wget -nv -O - https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz \
-    | tar -C /usr/local -xz
-
-ENV GOPATH=/go
-ENV PATH=$GOPATH/bin:/usr/local/go/bin:$PATH
+FROM golang:${GOLANG_VERSION} AS builder
```
elezar: Question: Is there a reason that we don't just add the verison here to make updating this simpler?

tariq1890: Great question. I think it now makes sense to add the version here and leverage dependabot to auto-update the go image

tariq1890: Done, please check now

docker/Dockerfile
```diff
@@ -12,27 +12,7 @@
...
-        ppc64el | ppc64le) ARCH='ppc64le' ;; \
-        aarch64 | arm64) ARCH='arm64' ;; \
-        *) echo "unsupported architecture" ; exit 1 ;; \
-    esac; \
-    wget -nv -O - https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz \
-    | tar -C /usr/local -xz
-
-ENV GOPATH=/go
-ENV PATH=$GOPATH/bin:/usr/local/go/bin:$PATH
+FROM golang:1.25.1 AS builder
```
elezar, line 15: nit: We can add a follow-up to pull the GO_VERSION in `versions.mk` like we do in the device plugin or container toolkit.

tariq1890: Will do.

PR #1746, opened 2025-10-02
Check if NVIDIA kernel modules are loaded to avoid modprobe

karthikvetrivel:
This PR is a part of this endeavor:

> GPU Driver container should avoid re-installing drivers on spurious container restarts

Relevant PRs: 

- https://github.com/NVIDIA/gpu-driver-container/pull/454
- https://github.com/NVIDIA/k8s-driver-manager/pull/103

### DIFF
--- a/api/nvidia/v1/clusterpolicy_types.go
+++ b/api/nvidia/v1/clusterpolicy_types.go
@@ -499,6 +499,15 @@ type DriverSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:select:auto,urn:alm:descriptor:com.tectonic.ui:select:open,urn:alm:descriptor:com.tectonic.ui:select:proprietary"
 	KernelModuleType string `json:"kernelModuleType,omitempty"`
 
+	// DriverType defines the type of NVIDIA driver to be deployed.
+	// Accepted values are gpu, vgpu, and vgpu-host-manager.
+	// +kubebuilder:validation:Enum=gpu;vgpu;vgpu-host-manager
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Driver Type"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.description="Driver Type"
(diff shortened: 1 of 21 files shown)


manifests/state-driver/0500_daemonset.yaml
```diff
@@ -148,6 +148,18 @@ spec:
...
+            value: {{ .Driver.Spec.Version | quote }}
+        {{- end }}
+        {{- if .Driver.Spec.KernelModuleType }}
+          - name: KERNEL_MODULE_TYPE
+            value: {{ .Driver.Spec.KernelModuleType | quote }}
+        {{- end }}
+        {{- if .Driver.Spec.DriverType }}
+          - name: DRIVER_TYPE
+            value: {{ .Driver.Spec.DriverType | quote }}
+        {{- end }}
```
cdesiniotis, line 162: Question -- do we not need to make similar changes to the DaemonSet that gets rendered by the ClusterPolicy controller?

cdesiniotis: Also, are we expecting our controller to set the `DRIVER_VERSION` variable or are we just relying on this variable being set in our driver-container image itself? (I think we should just do the latter).

karthikvetrivel: > Also, are we expecting our controller to set the `DRIVER_VERSION` variable or are we just relying on this variable being set in our driver-container image itself? (I think we should just do the latter).

For the init container, the controller must set it. For the driver container, we rely on the image.

cdesiniotis: Got it, makes sense that these variables need to always be added to the init container. I think my first question still stands -- do we need to make similar changes to the ClusterPolicy controller? Or is this already being set there?

Also, I notice that you only added the `DRIVER_TYPE` variable to the test files in `internal/state/testdata/golden`. Shouldn't we also be adding `DRIVER_VERSION` by default to the initContainer that gets rendered?

karthikvetrivel: I just pushed a commit--we do need to make similar changes in the ClusterPolicy controller. They're in now. 

>Also, I notice that you only added the DRIVER_TYPE variable to the test files in internal/state/testdata/golden. Shouldn't we also be adding DRIVER_VERSION by default to the initContainer that gets rendered? 

Looking into this now!

karthikvetrivel: DRIVER_VERSION is conditionally rendered in the template--the golden files only include DRIVER_VERSION when the test fixture sets .Spec.Version. 

I've now updated the test fixtures to set the version in `getMinimalDriverRenderData()`, which means that the golden files now include.

PR #1747, opened 2025-10-02
Opt-out of drop-in config files if RUNTIME_DROP_IN_CONFIG=""

cdesiniotis:
This change allows users to opt-out of using drop-in config files when configuring containerd or crio. This is useful in cases where the top-level config is a templated config file.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -2444,9 +2444,10 @@ func getRuntimeConfigFiles(c *corev1.Container, runtime string) (string, string,
 		} else if value := getContainerEnv(c, "RUNTIME_CONFIG"); value != "" {
 			topLevelConfigFile = value
 		}
+		// Users can opt-out of using drop-in files by setting RUNTIME_DROP_IN_CONFIG to ""
 		dropInConfigFile := DefaultContainerdDropInConfigFile
-		if value := getContainerEnv(c, "RUNTIME_DROP_IN_CONFIG"); value != "" {
-			dropInConfigFile = value
+		if containerContainsEnv(c, "RUNTIME_DROP_IN_CONFIG") {
+			dropInConfigFile = getContainerEnv(c, "RUNTIME_DROP_IN_CONFIG")
 		}
 		return topLevelConfigFile, dropInConfigFile, nil
 	case gpuv1.CRIO.String():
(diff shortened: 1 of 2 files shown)


controllers/object_controls.go
```diff
@@ -2457,9 +2458,10 @@ func getRuntimeConfigFiles(c *corev1.Container, runtime string) (string, string,
 		} else if value := getContainerEnv(c, "RUNTIME_CONFIG"); value != "" {
 			topLevelConfigFile = value
 		}
+		// Users can opt-out of using drop-in files by setting RUNTIME_DROP_IN_CONFIG to ""
```
tariq1890, line 2461: @elezar @cdesiniotis  I think we should reconsider this. Treating the absence of a value or empty string as an input is bug-prone and generally counterintuitive.

I think it's better just to use a bool flag called `RUNTIME_DROP_IN_CONFIG_ENABLED` and have users explicitly opt-out.

tariq1890: Once we implement CRI-based config retrieval, hopefully we wouldn't have to deal with these flags as much. So I think it's reasonable to have the bool flag `RUNTIME_DROP_IN_CONFIG_ENABLED` as a stopgap.

elezar: Note that the `RUNTIME_DROP_IN_CONFIG` will be used even when we switch to CRI-based config retrieval. The use of a drop-in or a fully-rendered config is separate from the config source.

My concern with adding yet another config option is that this is something that we need to maintain compatibility with.

As a follow-up question: Would using a reserved value like "disabled" for `RUNTIME_DROP_IN_CONFIG` to disable this instead fo `""` be more agreeable?

ArangoGutierrez: I agree with Evan, and I think that an empty value is ok, not to read as disabled, simply that we have a default  (empty value) behaviour, and a non-empty value is then a call to modify the default, which in the case of `RUNTIME_DROP_IN_CONFIG` is to modify the way a runtime config is handled

tariq1890: > to disable this instead fo "" be more agreeable

I don't recommend interpreting an empty string as input for disablement of a feature. I prefer being more explicit when it comes to this.

elezar: Just to clarify: `RUNTIME_DROP_IN_CONFIG` is not a feature. This is the *path* to the drop in config file that should be used.

However, thinking about this again, it is clear that the current implementation in the NVIDIA Container Toolkit container is NOT compatible with older GPU Operator versions out the box. In this case, the `RUNTIME_DROP_IN_CONFIG` envvar will not be specified -- which should mean that the default behaviour is to NOT use a drop in file.

I assume that our expectation is that the newer Container Toolkit works out of the box with at least the current GPU Operator version, or is that not a guarantee that we want to make?

cdesiniotis: I would be okay with v1.18.0 only being compatible with GPU Operator starting with the next release (25.10.0+).

cdesiniotis:
Closing this PR in favor of https://github.com/NVIDIA/gpu-operator/pull/1758. I am open to revisiting this PR if the need to disable drop-in files comes up again.

PR #1757, opened 2025-10-06
Fix helm template for empty validator env lists

rajathagasthya:
The clusterpolicy.yaml template was rendering invalid YAML when validator's env fields were empty lists. The template would render parent keys (`plugin:`) without any child fields, which is not a valid YAML.

This commit explicity renders `env: []` when it is empty, making the default CR valid.

**Results**

When `env` is empty:
```
$ helm template deployments/gpu-operator/ | grep validator -A 10
  validator:
    repository: nvcr.io/nvidia
    image: gpu-operator
    version: "main-latest"
    imagePullPolicy: IfNotPresent
    plugin:
      env: []

  mig:
    strategy: single
  psa:
```
When `env` is non-empty:
```
$ helm template deployments/gpu-operator/ --set 'validator.plugin.env[0].name=TEST,validator.plugin.env[0].value=false' | grep validator -A 10
  validator:
    repository: nvcr.io/nvidia
    image: gpu-operator
    version: "main-latest"
    imagePullPolicy: IfNotPresent
    plugin:
      env:
        - name: TEST
          value: false

  mig:
```


deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -89,42 +89,56 @@ spec:
     plugin:
       {{- if .Values.validator.plugin.env }}
       env: {{ toYaml .Values.validator.plugin.env | nindent 8 }}
+      {{- else }}
```
cdesiniotis, line 92: Instead of an `if-else` statement, why don't we leverage the [default](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) keyword to set `env` to an empty list if the field in the helm values is not defined? For example, the below appears to work as intended:
```
  env: {{ toYaml .Values.validator.plugin.env | default (list) | nindent 8 }}
```

cdesiniotis (commented):
Thanks @rajathagasthya for fixing this issue. I have left a suggestion. Also, there are other `env` fields, like `validator.env` that need to be addressed as well.

deployments/gpu-operator/templates/clusterpolicy.yaml
rajathagasthya: Unfortunately, this results in the following invalid YAML when `env` is an empty list (due to the unconditional `nindent`):
```
    repository: nvcr.io/nvidia
    image: gpu-operator
    version: "main-latest"
    imagePullPolicy: IfNotPresent
    plugin:
      env:
        []
```
I got this working with the `ternary` keyword:
```
env: {{ ternary (list) (toYaml .Values.validator.plugin.env | nindent 8) (empty .Values.validator.plugin.env) }}
```
I'm not sure how much of an improvement this is though. This can be a bit harder to read and reason compared to an explicit `if-else`, but I don't really have a strong opinion either way.

cdesiniotis: Got it, thanks for trying out the suggestion.

I would slightly prefer the `if-else` statements to avoid unnecessary cleverness / complexity.

rajathagasthya: I agree with that. I'll keep these changes and also include the other `env` fields. Thanks!

rajathagasthya: I went through the other `env` fields and they actually don't need this change. This `if/else` block is required when `env` is the only field under a parent key. When there are other keys in the parent, not having `env: []` is okay.

For example, the below does not work:
```
validator:
  plugin: 
    # no plugin.env field
```
because it's interpreted by YAML as:
```
validator:
  plugin: null
```

But not having `validator.env` does not violate the spec because of other fields:
```
validator:
  repository: "repo"
  # no validator.env field
```

cdesiniotis: Ah I see! Thanks for clarifying. This is also highlighted in the GitHub issue raised for this.

PR #1758, opened 2025-10-06
Allow config sources to be specified for containerd

cdesiniotis:
By default the toolkit container extracts the current containerd config by running the `containerd config dump` command and if that fails we read the existing config file.

This change allows ordering of these sources to be defined via the RUNTIME_CONFIG_SOURCE environment variable. This further allows an explicit path to be specified for the file config source so that there is a separation between the config file being modified by the toolkit container and the config used as the source for drop-in files, for example.

Depends on https://github.com/NVIDIA/nvidia-container-toolkit/pull/1251

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1403,6 +1403,25 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, configVol)
 	}
 
+	// Handle any additional runtime config sources
+	const runtimeConfigSourceFile = "file"
+	if runtimeConfigSources := getContainerEnv(container, "RUNTIME_CONFIG_SOURCE"); runtimeConfigSources != "" {
+		var sources []string
+		for _, runtimeConfigSource := range strings.Split(runtimeConfigSources, ",") {
+			parts := strings.SplitN(runtimeConfigSource, "=", 2)
+			if len(parts) == 1 || parts[0] != runtimeConfigSourceFile {
+				sources = append(sources, runtimeConfigSource)
+				continue
+			}
+			// We transform the host path to a container path by prepending "/host" to the file
+			// path. This works because the toolkit container has the host's root filesystem
+			// mounted as read-only at "/host"
+			sourceConfigFile := filepath.Join("/host", parts[1])
+			sources = append(sources, runtimeConfigSourceFile+"="+sourceConfigFile)
+		}
+		setContainerEnv(container, "RUNTIME_CONFIG_SOURCE", strings.Join(sources, ","))
+	}
+
 	// setup mounts for runtime socket file
 	runtimeSocketFile, err := getRuntimeSocketFile(container, runtime)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/object_controls.go
```diff
@@ -1403,6 +1403,24 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, configVol)
 	}
 
+	// Handle any additional runtime config sources
+	if runtimeConfigSources := getContainerEnv(container, "RUNTIME_CONFIG_SOURCE"); runtimeConfigSources != "" {
+		var sources []string
+		for _, runtimeConfigSource := range strings.Split(runtimeConfigSources, ",") {
+			parts := strings.SplitN(runtimeConfigSource, "=", 2)
+			if len(parts) == 1 || parts[0] != "file" {
```
tariq1890: Let's move `"file"` to constant variable. The variable could be called `runtimeConfigSourceFile`

cdesiniotis: Sure. Updated.

tariq1890 (approved):

controllers/object_controls.go
```diff
@@ -1403,6 +1403,25 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+		for _, runtimeConfigSource := range strings.Split(runtimeConfigSources, ",") {
+			parts := strings.SplitN(runtimeConfigSource, "=", 2)
+			if len(parts) == 1 || parts[0] != runtimeConfigSourceFile {
+				sources = append(sources, runtimeConfigSource)
+				continue
+			}
+			// We transform the host path to a container path by prepending "/host" to the file
+			// path. This works because the toolkit container has the host's root filesystem
+			// mounted as read-only at "/host"
+			sourceConfigFile := "/host" + parts[1]
```
elezar: One nit is that this ASSUMES that `parts[1]` is an absolute path.

ArangoGutierrez: ++

ArangoGutierrez (approved):
https://github.com/NVIDIA/nvidia-container-toolkit/pull/1251 is now merged

controllers/object_controls.go
cdesiniotis: I have updated to using `filepath.Join()` to be more robust. Although note, I would only expect the toolkit to be able to find the source config if `parts[1]` is an absolute path.

controllers/object_controls.go
```diff
@@ -1403,6 +1403,25 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+		for _, runtimeConfigSource := range strings.Split(runtimeConfigSources, ",") {
+			parts := strings.SplitN(runtimeConfigSource, "=", 2)
+			if len(parts) == 1 || parts[0] != runtimeConfigSourceFile {
+				sources = append(sources, runtimeConfigSource)
+				continue
+			}
+			// We transform the host path to a container path by prepending "/host" to the file
+			// path. This works because the toolkit container has the host's root filesystem
+			// mounted as read-only at "/host"
+			sourceConfigFile := filepath.Join("/host", parts[1])
```
ArangoGutierrez, line 1419: Nice!

PR #1775, opened 2025-10-10
Add permissions for nodefeatures in GFD's role object

cdesiniotis:
These permissions are required when GFD is configured to use the Node Feature API instead of feature files.


cdesiniotis:
I just realized this is not sufficient. We will also have to add these permissions to the gpu-operator cluster role as well -- otherwise it won't be able to create the GFD cluster role.

cdesiniotis:
I realized that NodeFeatures are a namespaced resource. I have moved the permissions to the role objects for gpu-operator and GFD respectively.

PR #1781, opened 2025-10-13
e2e: create separate module for Go/Ginkgo test migration

karthikvetrivel:
## Summary

This PR is the foundation and first step for migrating GPU Operator e2e tests from bash to Go/Ginkgo.  

This PR restructures the E2E testing infrastructure by introducing a dedicated Go module for end-to-end tests and removes unused dependencies from the main module.

### E2E Test Module
- Added `tests/e2e/go.mod` with isolated dependencies for E2E tests
- Added `tests/e2e/go.sum` with dependency checksums
- Separates E2E test dependencies from the main operator codebase

## Motivation

The gpu-operator's `go.mod` dependency tree has grown very large, primarily due to Helm dependencies that were added to enable writing E2E tests in Go. While these dependencies don't affect the compiled operator binary, they appear in the project's `go.mod`/`go.sum` files, causing two major issues:

1. **CVE False Positives** - Security scanners flag numerous CVEs from Helm's transitive dependencies, which are almost always false positives since these deps aren't in the runtime binary
2. **Large Dependency Graph** - Makes dependency management and auditing more complex than necessary

While multi-module Go repos are generally discouraged, this separation is justified because:
- E2E tests are not a deliverable artifact
- Keeps the main operator dependency tree focused and manageable
- Eliminates CVE noise from test-only dependencies
- The operator binary remains unaffected


bundle/manifests/nvidia.com_clusterpolicies.yaml
```diff
@@ -687,7 +687,7 @@ spec:
                       configMapName:
                         description: 'Deprecated: ConfigMapName has been deprecated
                           in favour of SecretName. Please use secrets to handle the
-                          LicensingConfig more securely'
+                          licensing server configuration more securely'
```
tariq1890, line 690: This diff should not be showing up in this PR. Can you make sure you've rebased on top of main correctly?

karthikvetrivel: This diff is generated when I run `make check`.  I just double checked and I'm rebased correctly.

EDIT: `make check` still completes successfully if this diff isn't present. The diff is just generated when the command is run.

tariq1890: You're right It looks like `main` is out of sync. Our CI checks should have caught this

tariq1890: I raised this PR to address this: https://github.com/NVIDIA/gpu-operator/pull/1782

tests/e2e/go.mod
```diff
@@ -0,0 +1,146 @@
+module github.com/NVIDIA/gpu-operator/tests/e2e
+
+go 1.24.3
```
rahulait, line 3: Should we also bump golang version to 1.25.2?

karthikvetrivel: The parent module `go.mod` uses 1.24.3 so we should update both or keep this one the same. @tariq1890 thoughts?

tariq1890: My preference is to keep the minimum supported go version here. I believe `go mod tidy` will automatically resolve that for us

karthikvetrivel: Sounds good, I will keep as is.

bundle/manifests/nvidia.com_clusterpolicies.yaml
tariq1890: The PR has been merged. Can you rebase?

karthikvetrivel: @tariq1890 Rebased. Let me know how the rest of the PR looks.

tests/e2e/go.mod
```diff
@@ -0,0 +1,146 @@
...
+
+require (
+	github.com/NVIDIA/gpu-operator v1.11.1
```
tariq1890: Would it work if we did this instead?
```suggestion
	github.com/NVIDIA/gpu-operator v0.0.0
```

bundle/manifests/nvidia.com_clusterpolicies.yaml
tariq1890: Thanks!

tests/e2e/go.mod
karthikvetrivel: I believe so because of the `replace` below. Updated.

PR #1782, opened 2025-10-13
check bundle/ and deployments/ dirs for CRD syncs

tariq1890:
This PR fixes the out-of-sync CRD manifests in the helm and OLM bundle directories. It turns out that the `make check` job wasn't checking uncommitted diffs in the directories mentioned earlier, so the `main` branch had stale CRD manifests for a while.

h/t @karthikvetrivel for calling this out!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,7 +241,7 @@ validate-helm-values: cmds
 
 validate-generated-assets: manifests generate generate-clientset sync-crds
 	@echo "- Verifying that the generated code and manifests are in-sync..."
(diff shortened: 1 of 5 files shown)


tariq1890:
@rajathagasthya Great question. We prefer to have the actual files here as these directories are used to create the helm charts/OLM bundles themselves. Our automation uses these directories to generate the artifacts be it pre-release or release

rajathagasthya:
Makes sense, thanks for the explanation!

PR #1783, opened 2025-10-13
Add GDRCopy validation to validator daemonset

cdesiniotis:
When GDRCopy is enabled, this ensures that the GDRCopy driver is loaded prior the the k8s-device-plugin from starting up.

### DIFF
--- a/cmd/nvidia-validator/main.go
+++ b/cmd/nvidia-validator/main.go
@@ -63,6 +63,9 @@ type Driver struct {
 // NvidiaFs GDS Driver component
 type NvidiaFs struct{}
 
+// GDRCopy driver component
+type GDRCopy struct{}
+
 // CUDA represents spec to run cuda workload
 type CUDA struct {
 	ctx        context.Context
@@ -150,6 +153,8 @@ const (
 	driverStatusFile = "driver-ready"
 	// nvidiaFsStatusFile indicates status file for nvidia-fs driver readiness
 	nvidiaFsStatusFile = "nvidia-fs-ready"
+	// gdrCopyStatusFile indicates status file for GDRCopy driver (gdrdrv) readiness
+	gdrCopyStatusFile = "gdrcopy-ready"
 	// toolkitStatusFile indicates status file for toolkit readiness
 	toolkitStatusFile = "toolkit-ready"
 	// pluginStatusFile indicates status file for plugin readiness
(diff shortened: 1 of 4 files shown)


cmd/nvidia-validator/main.go
```diff
@@ -944,6 +958,38 @@ func (n *NvidiaFs) runValidation(silent bool) error {
...
+
+	// create driver status file
+	err = createStatusFile(outputDirFlag + "/" + gdrCopyStatusFile)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (g *GDRCopy) runValidation(silent bool) error {
```
karthikvetrivel, line 982: I see this is in draft, so this is just a nit: might be worth tightening the `gdrdrv` module regex to avoid potential false positives--maybe something like this: `args := []string{"-c", "lsmod | grep -E '^gdrdrv\\s'"}`

This only matches when gdrdrv appears at the start of a line (first column)

cdesiniotis: Did you mean `\s` instead of `\\s` at the end of the regular expression?

karthikvetrivel: No--with a single backslash, we'll get an `unknown escape` error here. I believe Go will parse `\\s` to `\s`.

cdesiniotis: Gotcha. I have updated as you have suggested. 

Here are the logs from this container:
```
$ kubectl logs -n gpu-operator nvidia-operator-validator-c28kr -c gdrcopy-validation
time="2025-10-14T18:21:51Z" level=info msg="version: v25.3.4, commit: a476e1e6-dirty"
running command sh with args [-c lsmod | grep -E '^gdrdrv\s']
gdrdrv 28672 0 - Live 0xffffffffc2370000 (OE)
```

cdesiniotis:
@tariq1890 requesting your review. I am also not planning to address the failing coveralls check as the code introduced is in the validator, which doesn't have any unit testing currently.

controllers/object_controls.go
```diff
@@ -2335,6 +2336,12 @@ func TransformValidatorComponent(config *gpuv1.ClusterPolicySpec, podSpec *corev
 				podSpec.InitContainers = append(podSpec.InitContainers[:i], podSpec.InitContainers[i+1:]...)
 				return nil
 			}
+		case "gdrcopy":
+			if config.GDRCopy == nil || !config.GDRCopy.IsEnabled() {
```
tariq1890: can we add a helper method `config.IsGDRCopyEnabled()` to further simplify?
```suggestion
			if !config.IsGDRCopyEnabled() {
```

cdesiniotis: Updated as suggested.

PR #1797, opened 2025-10-17
include the go submodules in the validate-submodules make target

tariq1890:
The go mod tidy checks should include the submodules in this repository as well


Makefile
```diff
@@ -230,6 +230,16 @@ validate-modules:
 	@echo "- Checking if the vendor dir is in sync..."
 	go mod vendor
 	@git diff --exit-code -- vendor
+	@echo "- [tools] Verifying that the dependencies in the have expected content..."
```
rahulait:
```suggestion
	@echo "- [tools] Verifying that the dependencies have expected content..."
```

tariq1890: Thanks for catching that!

Makefile
```diff
@@ -230,6 +230,16 @@ validate-modules:
...
+	go -C $(PROJECT_DIR)/tools mod verify
+	@echo "- [tools] Checking for any unused/missing packages in go.mod..."
+	go -C $(PROJECT_DIR)/tools mod tidy
+	@git diff --exit-code -- $(PROJECT_DIR)/tools/go.sum $(PROJECT_DIR)/tools/go.mod
+	@echo "- [tests/e2e] Verifying that the dependencies in the have expected content..."
```
rahulait:
```suggestion
	@echo "- [tests/e2e] Verifying that the dependencies have expected content..."
```

Makefile
```diff
@@ -230,6 +230,16 @@ validate-modules:
...
+	go -C $(PROJECT_DIR)/tests/e2e mod verify
+	@echo "- [tests/e2e] Checking for any unused/missing packages in go.mod..."
+	go -C $(PROJECT_DIR)/tests/e2e mod tidy
+	@git diff --exit-code -- $(PROJECT_DIR)/tests/e2e/go.sum $(PROJECT_DIR)/tests/e2e/go.mod
```
rahulait, line 245: Would it be better to have global var `TOOLS_DIR` and `TESTS_DIR` and use that instead of `$(PROJECT_DIR)/tools` and `$(PROJECT_DIR)/tests/e2e`?

So like:
```
TOOLS_DIR := $(PROJECT_DIR)/tools
TESTS_DIR := $(PROJECT_DIR)/tests/e2e

validate-modules: validate-project validate-tools validate-tests

...
validate-tools:
  go -C $(TOOLS_DIR) mod verify
  ...
```

so something like this.

tests/e2e/go.mod
```diff
@@ -3,15 +3,13 @@ module github.com/NVIDIA/gpu-operator/tests/e2e
 go 1.24.3
 
 require (
-	github.com/NVIDIA/gpu-operator v0.0.0
```
karthikvetrivel, line 6: This dependency is needed required because `helpers/clusterpolicy.go` and `helpers/nvidiadriver.go `import CRD types and clientsets from it, but `go mod tidy` removes it because these helpers are never actually instantiated in any tests.  

I can write a quick test that uses these helpers so that the dep won't be removed?

karthikvetrivel: See this repo: https://github.com/NVIDIA/gpu-operator/pull/1788--I'm adding a test that uses these helpers now.

karthikvetrivel: Resolving this. I'll add a test in my PR linked in the comment above and rebase on top of this merged PR.

tariq1890: Thanks @karthikvetrivel !

Makefile
tariq1890: Yes, agreed! That is indeed much better

PR #1798, opened 2025-10-17
[DRAFT]node-feature-rules: Add 0x2321 as CC-capable device

manuelh-dev:


deployments/gpu-operator/templates/nodefeaturerules.yaml
```diff
@@ -89,6 +89,15 @@ spec:
...
             device: {op: In, value: ["2322"]}
+    - name: "NVIDIA H100L 94GB"
+      labels:
+        "nvidia.com/gpu.H100L": "true"
+        "nvidia.com/gpu.family": "hopper"
+      matchFeatures:
+        - feature: pci.device
+          matchExpressions:
+            vendor: {op: In, value: ["10de"]}
+            device: {op: In, value: ["2321"]}
```
manuelh-dev, line 100: Ideally we won't add devices piece meal. Can "whitelist" : 0x23**, 0x2b**, GBXXX -- Blackwell, GBXXX -- Hopper. Some may need to be excluded via "blacklist" then: exclude 2b00	TA1090SA [THOR].

assets/state-cc-manager/0500_daemonset.yaml
```diff
@@ -34,7 +34,9 @@ spec:
               fieldRef:
                 fieldPath: spec.nodeName
           - name: CC_CAPABLE_DEVICE_IDS
```
manuelh-dev, line 36: I wonder where this CC_CAPABLE_DEVICE_IDS variable is being referenced.

deployments/gpu-operator/values.yaml
```diff
@@ -511,7 +511,9 @@ ccManager:
   imagePullSecrets: []
   env:
     - name: CC_CAPABLE_DEVICE_IDS
-      value: "0x2339,0x2331,0x2330,0x2324,0x2322,0x233d"
```
elezar, line 514: As mentioned offline: The envvars from the values file should probably be removed so that a user can properly override them. The defaults should be specified in the `daemonset` template instead.

assets/state-cc-manager/0500_daemonset.yaml
elezar: Here: https://github.com/NVIDIA/k8s-cc-manager/blob/45f968d36e9c4bc39e497bd87aa6f5265648d0dd/cmd/main.go#L113

deployments/gpu-operator/templates/nodefeaturerules.yaml
elezar: One thing to note is that `matchExpressions` don't allow wildcards (as far as I am aware). Is there another component that could / should create thes labels instead of a nodefeature rule directly?

manuelh-dev: By the way, we could do something like the following:
```
- name: "NVIDIA Hopper GPU Family"
  labels:
    "nvidia.com/gpu.family": "hopper"
  matchFeatures:
    - feature: pci.device
      matchExpressions:
        vendor: {op: In, value: ["10de"]}
        device: {op: InRegexp, value: ["^23[0-9a-f]{2}$"]}
```
```
- name: "NVIDIA Blackwell GPU Family"
  labels:
    "nvidia.com/gpu.family": "blackwell"
  matchFeatures:
    - feature: pci.device
      matchExpressions:
        vendor: {op: In, value: ["10de"]}
        device: {op: InRegexp, value: ["^2b[0-9a-f]{2}$"]}
```

assets/state-cc-manager/0500_daemonset.yaml
manuelh-dev: Thank you! It looks like we may need a change in the k8s-cc-manager as well then. If we want to allow all `23` and `2b` Hopper/Blackwell GPUs, we may rather not want to pass a list of specific GPUs.

deployments/gpu-operator/values.yaml
manuelh-dev: Thank you! Offline we had a discussion that the change were some of the dev defaults were removed in `values.yaml` was the following: https://github.com/NVIDIA/gpu-operator/pull/1580 - the ccManager envvars may have potentially been missed to remove.

manuelh-dev:
Closing this pull request as we will address this in a more generic way via https://github.com/NVIDIA/gpu-operator/pull/1973

PR #1802, opened 2025-10-20
Replace kubectl usage with utility program to apply CRDs

rajathagasthya:
kubectl is heavyweight and often has CVEs, so we are forced to bump its version even if the final gpu-operator image does not contain kubectl bits. This change lets us remove that dependency and use client-go functions to manage CRDs.


cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func main() {
+	config := config{}
+
+	// Create the top-level CLI
+	c := cli.NewApp()
+	c.Name = "apply-crds"
+	c.Usage = "Tools for managing Custom Resource Definitions (CRDs) for NVIDIA GPU Operator"
+	c.Version = "0.1.0"
```
elezar: Should we set the version in the same way that we set the other CLI versions. This *should* give us the git commit SHA as well, which could e useful for debugging.

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+		logger.SetLevel(logLevel)
+		return nil
+	}
+
+	// Common flags for both apply and delete subcommands
+	commonFlags := []cli.Flag{
+		&cli.StringSliceFlag{
+			Name:     "crds-path",
+			Usage:    "Path to CRD manifest file or directory (can be specified multiple times, directories are searched recursively)",
+			Required: true,
```
elezar: Did we want an env here too?

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// Define the subcommands
+	c.Commands = []*cli.Command{
+		{
+			Name:  "apply",
+			Usage: "Apply CRDs from the specified path",
+			Flags: commonFlags,
+			Action: func(c *cli.Context) error {
+				config.crdsPaths = c.StringSlice("crds-path")
```
elezar: Why not set the `Destination` for the flag directly? (Note we recently updated to urfave v3 in `nvidia-container-toolkit` and this simplifies string slices quite a bit in that the destination can be an actual `[]string` and not a `cli.StringSlice`.

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	err := c.Run(os.Args)
+	if err != nil {
+		log.Errorf("%v", err)
+		log.Exit(1)
+	}
+}
+
+func runApply(cfg config) error {
+	logger.Infof("Applying CRDs from %d path(s): %v", len(cfg.crdsPaths), cfg.crdsPaths)
```
elezar: Do we use `klog` in the GPU Operator?

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	ctx := context.Background()
+	if err := crdutil.ProcessCRDs(ctx, cfg.crdsPaths, crdutil.CRDOperationApply); err != nil {
```
elezar: I know that this is out of scope for this PR, by why is the `ProcessCRDs` signature not:
```
ProcessCRDs(context.Context, Operation, ...string) error
```
?

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	logger.Info("Successfully applied CRDs")
+	return nil
+}
+
+func runDelete(cfg config) error {
+	logger.Infof("Deleting CRDs from %d path(s): %v", len(cfg.crdsPaths), cfg.crdsPaths)
+
+	ctx := context.Background()
```
elezar: Question: We already have a context in the CLI action. Why don't we just pass that in and use that?

elezar: (OK: I may be thinking in the CLI v3 sense here).

deployments/gpu-operator/templates/cleanup_crd.yaml
```diff
@@ -38,13 +38,9 @@ spec:
...
           - >
-              kubectl delete clusterpolicy cluster-policy;
-              kubectl delete crd clusterpolicies.nvidia.com;
-              kubectl delete crd nvidiadrivers.nvidia.com --ignore-not-found=true;
-            {{- if .Values.nfd.enabled -}}
-              kubectl delete crd nodefeatures.nfd.k8s-sigs.io --ignore-not-found=true;
-              kubectl delete crd nodefeaturegroups.nfd.k8s-sigs.io --ignore-not-found=true;
-              kubectl delete crd nodefeaturerules.nfd.k8s-sigs.io --ignore-not-found=true;
-            {{- end }}
+            apply-crds delete --crds-path /opt/gpu-operator/nvidia.com_clusterpolicies.yaml --crds-path /opt/gpu-operator/nvidia.com_nvidiadrivers.yaml;
```
elezar: I found the rework of args here: https://github.com/NVIDIA/gpu-operator/pull/1137/files#r2452418172 quite  clean. We really don't need to call the `apply-crds` command multiple times.

rajathagasthya: That does seem clean. I'm going to borrow it here, thanks.

cmd/apply-crds/main.go
rajathagasthya: No particular reason, really. Just didn't think of this :)

It's not too late if we want to change the signature there. @tariq1890 Thoughts?

cmd/apply-crds/main.go
rajathagasthya: I don't think it's necessary. I could it see `DEBUG` env above being useful, but I don't anticipate users wanting to set an env with these paths.

cmd/apply-crds/main.go
tariq1890: Currently klog is only used in the e2e tests submodule

cmd/apply-crds/main.go
tariq1890: Sounds good to me!

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	err := c.Run(os.Args)
+	if err != nil {
+		log.Errorf("%v", err)
+		log.Exit(1)
+	}
+}
+
+func runApply(cfg config) error {
```
tariq1890:
```suggestion
func runApply(ctx context.Context, cfg config) error {
```

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Define the subcommands
+	c.Commands = []*cli.Command{
+		{
+			Name:  "apply",
+			Usage: "Apply CRDs from the specified path",
+			Flags: commonFlags,
+			Action: func(c *cli.Context) error {
+				config.crdsPaths = c.StringSlice("crds-path")
+				return runApply(config)
```
tariq1890:
```suggestion
				return runApply(c.Context, config)
```

cmd/apply-crds/main.go
tariq1890: @elezar Agreed, the cliContext struct includes the (golang) Context, we can use that here

deployments/gpu-operator/templates/cleanup_crd.yaml
```diff
@@ -35,16 +35,12 @@ spec:
...
-          - >
-              kubectl delete clusterpolicy cluster-policy;
-              kubectl delete crd clusterpolicies.nvidia.com;
-              kubectl delete crd nvidiadrivers.nvidia.com --ignore-not-found=true;
-            {{- if .Values.nfd.enabled -}}
-              kubectl delete crd nodefeatures.nfd.k8s-sigs.io --ignore-not-found=true;
-              kubectl delete crd nodefeaturegroups.nfd.k8s-sigs.io --ignore-not-found=true;
-              kubectl delete crd nodefeaturerules.nfd.k8s-sigs.io --ignore-not-found=true;
-            {{- end }}
+            - /usr/bin/apply-crds
```
cdesiniotis: Shouldn't we be invoking the `delete` subcommand?

deployments/gpu-operator/templates/upgrade_crd.yaml
```diff
@@ -83,13 +83,12 @@ spec:
           image: {{ include "gpu-operator.fullimage" . }}
           imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
           command:
-          - sh
-          - -c
-          - >
-            kubectl apply -f /opt/gpu-operator/nvidia.com_clusterpolicies.yaml;
-            kubectl apply -f /opt/gpu-operator/nvidia.com_nvidiadrivers.yaml;
+            - /usr/bin/apply-crds
```
cdesiniotis: Shouldn't we be invoking the `apply` subcommand?

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,129 @@
...
+			logLevel = log.DebugLevel
+		}
+		logger.SetLevel(logLevel)
+		return nil
+	}
+
+	// Common flags for both apply and delete subcommands
+	commonFlags := []cli.Flag{
+		&cli.StringSliceFlag{
+			Name:        "crds-path",
```
cdesiniotis: nit: since the scope of this CLI is currently confined to just applying / deleting CRDs, what if we renamed this flag to `--filepath` with `-f` as the shorthand notation?

cmd/apply-crds/main.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func main() {
+	config := config{
+		crdsPaths: cli.NewStringSlice(),
+	}
+
+	// Create the top-level CLI
+	c := cli.NewApp()
+	c.Name = "apply-crds"
```
cdesiniotis: Can we come up with a better name for this utility? Especially since `apply` is the name of a subcommand, it feels redundant having `apply` in the top-level CLI name. `crdctl`? `manage-crds`?

deployments/gpu-operator/templates/cleanup_crd.yaml
rajathagasthya: Yes! :) I accidentally deleted it when I made formatting changes.

cmd/apply-crds/main.go
rajathagasthya: Good suggestion. Renamed to `manage-crds`.

PR #1811, opened 2025-10-21
Bump terminationGracePeriodSeconds to 120 for vGPU Manager daemonset

cdesiniotis:
This ensures the vGPU Manager has enough time to finish its cleanup logic when the pod is terminated. This fixes an issue observed on a HGX node (with 8 GPUs) where the vGPU Manager was killed before it finished disabling VFs on all 8 GPUs.

### DIFF
--- a/manifests/state-driver/0500_daemonset.yaml
+++ b/manifests/state-driver/0500_daemonset.yaml
@@ -59,6 +59,9 @@ spec:
         {{- .Driver.Spec.Labels | yaml | nindent 8 }}
         {{- end }}
     spec:
+      {{- if eq .Driver.Spec.DriverType "vgpu-host-manager" }}
+      terminationGracePeriodSeconds: 120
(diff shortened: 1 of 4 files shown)


manifests/state-driver/0500_daemonset.yaml
```diff
@@ -59,6 +59,9 @@ spec:
         {{- .Driver.Spec.Labels | yaml | nindent 8 }}
         {{- end }}
     spec:
+      {{- if eq .Driver.Spec.DriverType "vgpu-host-manager" }}
```
tariq1890, line 62: I think we'll need to update the test cases for this

cdesiniotis: I did update the test cases.

tariq1890: You're right. Resolving this comment

PR #1816, opened 2025-10-22
fix: add concurrency group for ci workflow

rahulait:
Adding concurrency group helps in preventing multiple concurrent workflow runs for the same PR. When a new commit is added to the PR, old workflow runs still running for the PR should get cancelled as they are no longer relevant and it should start again. This PR adds concurrency group with `cancel-in-progress` set which makes sure stale workflow runs gets cancelled and we don't waste GHA cycles for runs/steps which don't matter as the code has been updated.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,6 +21,10 @@ on:
       - main
       - release-*
 
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   ### Configuration checks ###
   helm-lint:


rajathagasthya (approved):

.github/renovate.json
```diff
@@ -4,20 +4,6 @@
   "prConcurrentLimit": 10,
   "dependencyDashboard": false,
   "extends": [
-    "config:recommended",
-    ":disableDependencyDashboard",
```
tariq1890, line 8: Is this change related to the PR?

tariq1890 (commented):
Thanks @rahulait! This is indeed very valuable

.github/renovate.json
rahulait: Not really, just saw that we have duplicate lines here and removed it in same PR. I can open another PR for it if you think we should separate these two out.

tariq1890: Yes, that would be preferable

rahulait: updated

cdesiniotis (approved):

Issue #1372, opened 2025-03-28
Nvidia GPU operator issue on openshift(4.17.20)


Nikhil-VW:
Hello Team,

Recently, while upgrading the OpenShift cluster version from 4.17.18 to 4.17.20 version (Kubernetes version - v1.30.10), the NVIDIA GPU Operator was upgraded to version V25.3 After the upgrade, the NVIDIA GPU Operator was down, which impacted the applications using the GPU. All the pods(dcgm,validator,dcgm-exporter) from the NVIDIA GPU Operator were in the "init" stage and also gpu cluster policy was in not ready stage . We later rolled back to the previous version, V24.9.2, which eventually resolved the issue.

Could the new version, V25.3, be unsupported by the latest OpenShift and Kubernetes version, or did it contain bugs that caused the issue? Can someone investigate this and let us know the major cause of the problem? We need to document the reason and provide it to the business department as part of the post-mortem analysis.

Let me know if you require any details.  

Thanks,

cdesiniotis:
Would you be able to describe which component was failing and provide logs from it? Alternatively, you can provide the full debug bundle by running:
```
curl -o must-gather.sh -L https://raw.githubusercontent.com/NVIDIA/gpu-operator/main/hack/must-gather.sh
chmod +x must-gather.sh
./must-gather.sh
```
This bundle can be submitted to us via email: [operator_feedback@nvidia.com](mailto:operator_feedback@nvidia.com)

Nikhil-VW:
Hello Christopher,

I have sent you the mail with attachment, please check ane let us know if you require any other details.

cdesiniotis:
It appears the must-gather logs you collected were for the working v24.9.2 installation. Would you be able to capture the must-gather logs from the failing v25.3.0 install?

cdesiniotis:
cc @empovit

kaifrazaVWITS:
Hi @empovit @cdesiniotis 
We had an issue regarding the newer version of GPU operator i.e version v25.3.0, as discussed we where facing issue regarding the newer version, what we did is we installed the sample gpu pod and the pod is status is in pending state, as you have requested to the the specific GPU debug and asked us to execute the command i.e 

curl -o must-gather.sh -L https://raw.githubusercontent.com/NVIDIA/gpu-operator/main/hack/must-gather.sh
chmod +x must-gather.sh
./must-gather.sh

After execution we get output logs so that they can help you in resolving the query. 

[...]

PR #1819, opened 2025-10-22
Use public packages to create dev/char symlinks

elezar:
This change switches to the nvdevices and nvmodules packages
from nvidia-container-toolkit to create symlinks in /dev/char
instead of importing the cmd package.

This depends on #1372
See https://github.com/NVIDIA/mig-parted/pull/262 for related work.


cmd/nvidia-validator/main.go
```diff
@@ -865,21 +866,27 @@ func createDevCharSymlinks(driverInfo driverInfo, disableDevCharSymlinkCreation
...
-		devchar.WithDevCharPath(hostDevCharPath),
-		devchar.WithCreateAll(true),
-		devchar.WithCreateDeviceNodes(true),
-		devchar.WithLoadKernelModules(loadKernelModules),
+	if loadKernelModules {
+		modules := nvmodules.New(
+			nvmodules.WithRoot(driverRootCtrPath),
+		)
+		if err := modules.LoadAll(); err != nil {
+			return fmt.Errorf("failed to load NVIDIA kernel modules: %v", err)
```
tariq1890, line 874:
```suggestion
			return fmt.Errorf("failed to load NVIDIA kernel modules: %w", err)
```

cmd/nvidia-validator/main.go
```diff
@@ -865,21 +866,27 @@ func createDevCharSymlinks(driverInfo driverInfo, disableDevCharSymlinkCreation
...
 	)
 	if err != nil {
-		return fmt.Errorf("error creating symlink creator: %w", err)
+		return err
 	}
 
-	err = creator.CreateLinks()
-	if err != nil {
+	if err := devices.CreateNVIDIAControlDevices(); err != nil {
+		return fmt.Errorf("failed to create device nodes for NVIDIA control devices: %v", err)
```
tariq1890, line 886:
```suggestion
		return fmt.Errorf("failed to create device nodes for NVIDIA control devices: %w", err)
```

PR #1384, opened 2025-04-05
Feat: Allow to configure kubelet root directory

apten-fors:
This pull request introduces a new configuration option `KubeletRootDir` to specify the root directory for the kubelet base directory. The changes include updates to the cluster policy types, CRD manifests, controllers, and tests to support this new configuration.

### Configuration Enhancements:

* Added `KubeletRootDir` field to the `HostPathsSpec` struct in `api/nvidia/v1/clusterpolicy_types.go` to represent the kubelet base directory.
* Updated CRD manifests (`bundle/manifests/nvidia.com_clusterpolicies.yaml`, `config/crd/bases/nvidia.com_clusterpolicies.yaml`, `deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml`) to include the new `KubeletRootDir` field.

### Controller Updates:

* Added `DefaultKubeletRootDir` constant and updated `preProcessDaemonSet` function in `controllers/object_controls.go` to transform the kubelet root directory if a custom one is configured. [[1]](diffhunk://#diff-634d1afac63d5d768c5ce2a20a947dc0842a3480923500e36c0d6da8dd220e84R178-R186) [[2]](diffhunk://#diff-634d1afac63d5d768c5ce2a20a947dc0842a3480923500e36c0d6da8dd220e84R751-R753)
* Implemented `transformForKubeletRootDir` function in `controllers/object_controls.go` to apply necessary transformations for custom kubelet root directory.

### Testing:

* Added `TestTransformForKubeletRootDir` function in `controllers/transforms_test.go` to validate the transformations for different kubelet root directory configurations.

### Deployment Configuration:

* Updated `deployments/gpu-operator/templates/clusterpolicy.yaml` and `deployments/gpu-operator/values.yaml` to include the new `kubeletRootDir` configuration. [[1]](diffhunk://#diff-3af3dbe02c9930f0b55cad1ac6cc0ce46dcf024b64a7bcceaaf3442753b11dddR18) [[2]](diffhunk://#diff-aecfc7542f3194d151b0a0812fa08dca7a2147da1c6bc06f41f19aa2b4f934afR35-R38)

Closes #1358 
Closes #745

PR #1820, opened 2025-10-22
Allow kubeletPath to be set for DCGM

Archimonde666:
As in : 
https://github.com/NVIDIA/dcgm-exporter/blob/694914191e72d0c8587ba27a41248049d7fe9494/deployment/templates/daemonset.yaml#L77


tariq1890:
Sorry for the delayed response, @Archimonde666 ! Unfortunately, this change is already covered by PR #1384 , so we will be closing this as a duplicate.

Thanks for the taking the time to contribute to gpu-operator!

PR #1826, opened 2025-10-23
Configure dependabot for tests/e2e Go module

rajathagasthya:
Dependencies in `tests/e2e` module also need to be updated regularly.


rajathagasthya:
Not sure if this is actually necessary or if dependabot can update sub-modules under a main module without extra configuration.

PR #1832, opened 2025-10-24
[refactor] Replace custom pointer helpers with k8s.io/utils/ptr.To()

karthikvetrivel:
Replace custom utility functions (`BoolPtr()`, `Int64Ptr()`, `newHostPathType()`) with the standard `ptr.To()` from `k8s.io/utils/ptr` package.

### DIFF
--- a/controllers/transforms_test.go
+++ b/controllers/transforms_test.go
@@ -377,8 +377,8 @@ func TestTransformForRuntime(t *testing.T) {
 			input: NewDaemonset().
 				WithContainer(corev1.Container{Name: "test-ctr"}),
 			expectedOutput: NewDaemonset().
-				WithHostPathVolume("containerd-config", filepath.Dir(DefaultContainerdConfigFile), newHostPathType(corev1.HostPathDirectoryOrCreate)).
-				WithHostPathVolume("containerd-drop-in-config", "/etc/containerd/conf.d", newHostPathType(corev1.HostPathDirectoryOrCreate)).
+				WithHostPathVolume("containerd-config", filepath.Dir(DefaultContainerdConfigFile), ptr.To(corev1.HostPathDirectoryOrCreate)).
+				WithHostPathVolume("containerd-drop-in-config", "/etc/containerd/conf.d", ptr.To(corev1.HostPathDirectoryOrCreate)).
 				WithHostPathVolume("containerd-socket", filepath.Dir(DefaultContainerdSocketFile), nil).
 				WithContainer(corev1.Container{
 					Name: "test-ctr",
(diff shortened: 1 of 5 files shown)


rajathagasthya (approved):

internal/utils/utils.go
```diff
@@ -67,6 +68,11 @@ func Int64Ptr(v int64) *int64 {
 	return &v
 }
 
+// HostPathTypePtr returns a pointer to the HostPathType passed in.
+func HostPathTypePtr(v corev1.HostPathType) *corev1.HostPathType {
```
tariq1890: Let's move these `ptr` methods to its own subpackage within the `utils` package. This way, they can be imported like this:

```
ptr.HostPathType(corev1.HostPathDirectory)
```

karthikvetrivel: Updated! 

EDIT: As an added bonus, this change allows the PR to pass coveralls!

controllers/transforms_test.go
```diff
@@ -444,8 +445,8 @@ func TestTransformForRuntime(t *testing.T) {
 					},
 				}),
 			expectedOutput: NewDaemonset().
-				WithHostPathVolume("containerd-config", filepath.Dir(DefaultContainerdConfigFile), newHostPathType(corev1.HostPathDirectoryOrCreate)).
-				WithHostPathVolume("containerd-drop-in-config", "/etc/containerd/conf.d", newHostPathType(corev1.HostPathDirectoryOrCreate)).
+				WithHostPathVolume("containerd-config", filepath.Dir(DefaultContainerdConfigFile), internalptr.HostPathType(corev1.HostPathDirectoryOrCreate)).
```
cdesiniotis: Question -- is there a technical reason for not using the methods provided in `k8s.io/utils/ptr`? For example, could we not replace this with:
```suggestion
				WithHostPathVolume("containerd-config", filepath.Dir(DefaultContainerdConfigFile), ptr.To(corev1.HostPathDirectoryOrCreate)).
```
which would eliminate the need for our own internal `ptr` package?

karthikvetrivel: Great catch, I think you're right. I don't think we need our own `ptr` package. 

I'm not sure why we had these original pointer methods. I guess that means we don't need `newHostPathType` at all.  I've updated the code.

cdesiniotis: Sweet, thanks!

Can you update the PR title? After getting another pair of eyes on this, we can squash then merge.

karthikvetrivel: Done. 

Me when I realized I could've used the `ptr` package: 
batman-gif

Thanks, Chris!

tariq1890: Thanks!

tariq1890:
LGTM. Please squash the commit history and we can merge

karthikvetrivel:
Squashed.

PR #1833, opened 2025-10-24
[refactor] Replace parseOSRelease env var hack with mockable function variable

karthikvetrivel:
Resolves TODO to mock parseOSRelease instead of using environment variable check.

**Design Decision:**
I used function variable pattern (vs interface) for simplicity. This matches the stdlib approach (e.g., `time.Now`) and keeps changes minimal.

**Testing:**
All existing tests pass with mock enabled. No production code behavior changed.

Note: I also added `defer f.Close()` to prevent file descriptor leaks

### DIFF
--- a/controllers/object_controls_test.go
+++ b/controllers/object_controls_test.go
@@ -146,6 +146,18 @@ func getModuleRoot(dir string) (string, error) {
 	return dir, nil
 }
 
+// mockOSRelease returns a mock parseOSRelease function for testing.
+// It allows tests to simulate different operating systems without filesystem access.
+func mockOSRelease(osID, version string) func() (map[string]string, error) {
+	return func() (map[string]string, error) {
+		return map[string]string{
+			"ID":         osID,
+			"VERSION_ID": version,
+			"NAME":       osID,
+		}, nil
+	}
+}
+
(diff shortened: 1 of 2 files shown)


controllers/object_controls_test.go
```diff
@@ -1393,3 +1407,68 @@ func TestService(t *testing.T) {
 		})
 	}
 }
+
+func TestParseOSReleaseFromFile(t *testing.T) {
+	parseFile := func(path string) (map[string]string, error) {
```
cdesiniotis: Why are we re-implementing this method as opposed to directly testing the `parseOSReleaseFromFile` method from `object_controls.go`? Testing code that does not actually get included in the final binary seems counter productive to me.

What about defining a variable for the path to the `os-release` file that we would override in our test code?

karthikvetrivel: Yes--good point. Updated to use a path to the `os-release` file.

karthikvetrivel:
@cdesiniotis I added this snippet:

	originalPath := osReleaseFilePath
	defer func() { osReleaseFilePath = originalPath }()
	
to make sure that `osReleaseFilePath` gets restored to its original value after the test completes. Now, this only matters if we have later tests that use it as well (which may be unlikely). Let me know if we should keep it.

PR #1838, opened 2025-10-27
[improvement] : add logic to retry if update fails due to conflict

rahulait:
When updating the resource, we should handle and retry on any conflicts. In this PR, we are wrapping the update call within retry block which will retry update in case of any conflicts.

### DIFF
--- a/internal/conditions/clusterpolicy.go
+++ b/internal/conditions/clusterpolicy.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -55,14 +56,12 @@ func (u *clusterPolicyUpdater) SetConditionsError(ctx context.Context, cr any, r
 	return u.setConditions(ctx, clusterPolicyCr, Error, reason, message)
 }
 
-func (u *clusterPolicyUpdater) setConditions(ctx context.Context, cr *nvidiav1.ClusterPolicy, statusType, reason, message string) error {
-	reqLogger := log.FromContext(ctx)
+// updateConditions updates the conditions of the ClusterPolicy CR
+func (u *clusterPolicyUpdater) updateConditions(ctx context.Context, cr *nvidiav1.ClusterPolicy, statusType, reason, message string) error {
 	// Fetch latest instance and update state to avoid version mismatch
 	instance := &nvidiav1.ClusterPolicy{}
-	err := u.client.Get(ctx, types.NamespacedName{Name: cr.Name}, instance)
-	if err != nil {
-		reqLogger.Error(err, "Failed to get ClusterPolicy instance for status update", "name", cr.Name)
-		return err
+	if err := u.client.Get(ctx, types.NamespacedName{Name: cr.Name}, instance); err != nil {
+		return fmt.Errorf("failed to get ClusterPolicy instance for status update: %w", err)
 	}
 
 	switch statusType {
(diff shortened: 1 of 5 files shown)


internal/conditions/clusterpolicy.go
```diff
@@ -57,45 +58,47 @@ func (u *clusterPolicyUpdater) SetConditionsError(ctx context.Context, cr any, r
...
+				Message: message,
+			})
+		default:
+			reqLogger.Error(nil, "Unknown status type provided", "statusType", statusType)
+			return fmt.Errorf("unknown status type provided: %s", statusType)
+		}
+		err = u.client.Status().Update(ctx, instance)
+		if err != nil {
+			reqLogger.Error(err, "Failed to update ClusterPolicy status", "name", cr.Name)
+		}
```
karthikvetrivel: Do you think it might be worth moving this error log? The error is logged on every retry attempt, including retriable conflict errors. This will spam logs with expected transient errors.

We could log only after all retries are exhausted.

internal/conditions/clusterpolicy.go
```diff
@@ -57,45 +58,47 @@ func (u *clusterPolicyUpdater) SetConditionsError(ctx context.Context, cr any, r
...
-			Message: message,
-		})
-	default:
-		reqLogger.Error(nil, "Unknown status type provided", "statusType", statusType)
-		return fmt.Errorf("unknown status type provided: %s", statusType)
-	}
-
-	return u.client.Status().Update(ctx, instance)
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		instance := &nvidiav1.ClusterPolicy{}
```
tariq1890: can we move the retried method to a variable? I prefer a single-line return statement for readability

internal/conditions/clusterpolicy.go
tariq1890: Good callout @karthikvetrivel! I agree, we could either
i) Log the retry attempt along with the error message

OR

ii) move the error log out of the method being retried

rahulait: sounds good, updated the PR

internal/conditions/clusterpolicy.go
rahulait: Updated.

internal/conditions/clusterpolicy.go
```diff
@@ -99,3 +99,18 @@ func (u *clusterPolicyUpdater) setConditions(ctx context.Context, cr *nvidiav1.C
...
+// with retry on conflict to handle version mismatches
+func (u *clusterPolicyUpdater) setConditions(ctx context.Context, cr *nvidiav1.ClusterPolicy, statusType, reason, message string) error {
+	reqLogger := log.FromContext(ctx)
+
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return u.updateConditions(ctx, cr, statusType, reason, message)
+	})
+
+	if err != nil {
+		reqLogger.Error(err, "Failed to update ClusterPolicy status after retries", "name", cr.Name)
```
rajathagasthya, line 111: nit: [let's not capitalize error strings.](https://go.dev/wiki/CodeReviewComments#error-strings)

internal/conditions/nvidiadriver.go
```diff
@@ -113,3 +113,18 @@ func (u *nvDriverUpdater) setConditions(ctx context.Context, cr *nvidiav1alpha1.
...
+// with retry on conflict to handle version mismatches
+func (u *nvDriverUpdater) setConditions(ctx context.Context, cr *nvidiav1alpha1.NVIDIADriver, statusType, reason, message string) error {
+	reqLogger := log.FromContext(ctx)
+
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return u.updateConditions(ctx, cr, statusType, reason, message)
+	})
+
+	if err != nil {
+		reqLogger.Error(err, "Failed to update NVIDIADriver status after retries", "name", cr.Name)
```
rajathagasthya, line 125: Ditto.

internal/conditions/clusterpolicy.go
rahulait: Since we are logging here and not formatting, does this apply here as well? The go doc says `This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.`

rajathagasthya: Makes sense. Let's keep as is.

PR #1841, opened 2025-10-28
modify Dockerfile to stop pulling cuda:12.9.1-ubi9 base image

tariq1890:
We currently depend on the 12.9.1 CUDA base image to fetch the cuda packages for backwards compatibility reasons as we have to support driver versions from branches as old as R535.

With this change, we stop pulling the `nvcr.io/nvidia/cuda:12.9.1-base-ubi9` so that this dependency is never inadvertently bumped by dependabot

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ARG CUDA_SAMPLES_VERSION=12.9
+
 FROM golang:1.25.3 AS builder
 
 ARG GOPROXY="https://proxy.golang.org,direct"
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -38,32 +40,29 @@ ARG VERSION="unknown"
...
 
+ENV PATH=/usr/local/cuda/bin:${PATH}
+
 WORKDIR /build
 
 ARG SAMPLE_NAME=vectorAdd
 
-ARG CUDA_SAMPLES_VERSION=v12.9
-RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/${CUDA_SAMPLES_VERSION} | \
+RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${CUDA_COMPAT_VERSION/-/.} | \
```
cdesiniotis: If this doesn't cause too much trouble, I would prefer if we kept `CUDA_SAMPLES_VERSION` as the build argument and construct the compat version string from it. From a maintenance perspective, it is the CUDA samples version that we are concerned about maintaining / bumping.

tariq1890: Sure, that makes sense. Will make the change

cdesiniotis (approved):

PR #1844, opened 2025-10-30
[improvement] : fix logging for controllers

rahulait:
This PR improves the error logging for various controllers. Changes include:
1. Changing `r.Log.V(consts.LogLevelDebug).Error` to `r.log.Error` as logr doesn't have verbosity levels for Error messages. Error messages are always logged. https://pkg.go.dev/github.com/go-logr/logr#hdr-Verbosity
2. Logging entire error object instead of just the message so that other relevant info also gets logged.
3. Used `%w` instead of `%v` when doing error wrapping so that entire error object gets wrapped and any helper functions like `apierrors.IsNotFound(err)` can look into wrapped error as well to identify the error correctly. Previously, since `%v` was used, these helper functions would have failed to correctly identify the wrapped error.
4. Changed `return ctrl.Result{RequeueAfter: time.Second * 5}, statusError` to `return ctrl.Result{}, statusError` as the fixed retry interval gets ignored if returned error is not nil. https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.3/pkg/reconcile#TypedReconciler

### DIFF
--- a/controllers/clusterpolicy_controller.go
+++ b/controllers/clusterpolicy_controller.go
@@ -41,8 +41,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
-	"github.com/NVIDIA/k8s-operator-libs/pkg/consts"
-
 	gpuv1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1"
 	"github.com/NVIDIA/gpu-operator/internal/conditions"
 )
@@ -97,11 +95,9 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// Fetch the ClusterPolicy instance
 	instance := &gpuv1.ClusterPolicy{}
-	var condErr error
-	err := r.Get(ctx, req.NamespacedName, instance)
-	if err != nil {
-		err = fmt.Errorf("failed to get ClusterPolicy object: %v", err)
-		r.Log.Error(nil, err.Error())
+	if err := r.Get(ctx, req.NamespacedName, instance); err != nil {
+		err = fmt.Errorf("failed to get ClusterPolicy object: %w", err)
+		r.Log.Error(err, "unable to fetch ClusterPolicy")
 		clusterPolicyCtrl.operatorMetrics.reconciliationStatus.Set(reconciliationStatusClusterPolicyUnavailable)
 		if apierrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
@@ -110,9 +106,8 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		condErr = r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, err.Error())
-		if condErr != nil {
-			r.Log.V(consts.LogLevelDebug).Error(nil, condErr.Error())
+		if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, err.Error()); condErr != nil {
+			r.Log.Error(condErr, "failed to set condition")
 		}
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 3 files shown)


controllers/clusterpolicy_controller.go
```diff
@@ -123,16 +118,13 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		instance.SetStatus(gpuv1.Ignored, clusterPolicyCtrl.operatorNamespace)
 		// do not change `clusterPolicyCtrl.operatorMetrics.reconciliationStatus` here,
 		// spurious reconciliation
-		return ctrl.Result{}, err
+		return ctrl.Result{}, nil
```
cdesiniotis, line 121: Can you describe the rationale for this change?

controllers/nvidiadriver_controller.go
```diff
@@ -79,47 +80,41 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
 	// Get the singleton NVIDIA ClusterPolicy object in the cluster.
 	clusterPolicyList := &gpuv1.ClusterPolicyList{}
-	err = r.List(ctx, clusterPolicyList)
-	if err != nil {
-		err = fmt.Errorf("error getting ClusterPolicy list: %v", err)
-		logger.V(consts.LogLevelError).Error(nil, err.Error())
+	if err := r.List(ctx, clusterPolicyList); err != nil {
+		err = fmt.Errorf("error getting ClusterPolicy list: %w", err)
+		logger.Error(err, "error getting ClusterPolicy list")
```
cdesiniotis: Did you mean
```suggestion
		wrappedError := fmt.Errorf("error getting ClusterPolicy list: %w", err)
		logger.Error(err, "error getting ClusterPolicy list")
```
to align with the similar code block from above?

controllers/nvidiadriver_controller.go
```diff
@@ -79,47 +80,41 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 		}
-		return reconcile.Result{}, fmt.Errorf("error getting ClusterPolicyList: %v", err)
+		return reconcile.Result{}, err
 	}
 
 	if len(clusterPolicyList.Items) == 0 {
-		err = fmt.Errorf("no ClusterPolicy object found in the cluster")
-		logger.V(consts.LogLevelError).Error(nil, err.Error())
+		err := fmt.Errorf("no ClusterPolicy object found in the cluster")
+		logger.Error(err, "no ClusterPolicy object found in the cluster")
```
cdesiniotis: Question: What will get printed here? I assume our message will be printed twice?

controllers/clusterpolicy_controller.go
rahulait: Yeah, this code was like:
```
var err error
<some steps>
if err != nil {
    do something
}

if somecheck {
    return ctrl.Result{}, err
}
```

So, err was always nil when it reached this part of code. Hence, just changed it to nil and also the scope of err variable is now limited with the changes, so its not available here.

controllers/upgrade_controller.go
```diff
@@ -218,7 +218,7 @@ func (r *UpgradeReconciler) removeNodeUpgradeStateLabels(ctx context.Context) er
 			delete(node.Labels, upgradeStateLabel)
 			err = r.Update(ctx, node)
 			if err != nil {
-				r.Log.V(consts.LogLevelError).Error(
+				r.Log.Error(
```
cdesiniotis: nit: can we update this to just be one line?

controllers/clusterpolicy_controller.go
cdesiniotis: Ack, thank you for the explanation.

controllers/nvidiadriver_controller.go
rahulait: Yes, msg and error keys will have the same message. I updated it now to have different values. Here is what it will log now:
```
{"level":"error","ts":1762442249.3265705,"msg":"failed to get ClusterPolicy object","controller":"nvidia-driver-controller","object":{"name":"test1"},"namespace":"","name":"test1","reconcileID":"c1345edc-ac0c-4417-841f-d71f101c5071","error":"no ClusterPolicy object found in the cluster"}
{"level":"error","ts":1762442249.3296852,"msg":"Reconciler error","controller":"nvidia-driver-controller","object":{"name":"test1"},"namespace":"","name":"test1","reconcileID":"c1345edc-ac0c-4417-841f-d71f101c5071","error":"no ClusterPolicy object found in the cluster"}
```

controllers/nvidiadriver_controller.go
rahulait: Yeah, we can do that here as well. Updated the code.

controllers/upgrade_controller.go
rahulait: updated.

tariq1890 (approved):

Issue #1795, opened 2025-10-16
NVIDIADriver CR does not honor `driver.enabled: false`


chiragjn:
**Describe the bug**
We are managing drivers on the node, so we have the following values

```
driver:
  enabled: false
  version: "570.148.08"
  usePrecompiled: true
  kernelModuleType: "auto"
  nvidiaDriverCRD:
    enabled: true
    deployDefaultCR: true
    nodeSelector: {}
  upgradePolicy:
    autoUpgrade: true
    maxParallelUpgrades: 1
    gpuPodDeletion:
      force: false
      timeoutSeconds: 600
      deleteEmptyDir: true
```

yet we see because of the `NVIDIADriver` CR the operator will still try to manage the drivers. 

We are able to get around the issue by setting `nvidiaDriverCRD.enabled: false`. We expected since it is nested under the `driver` section, `driver.enabled` will take precedence


**Expected behavior**
`driver.enabled: false` should disable all drivers management

**Environment (please provide the following information):**
 - GPU Operator Version: v25.3.4
 - OS: Talos Linux
 - Kernel Version: N/A
 - Container Runtime Version: N/A
 - Kubernetes Distro and Version: N/A


**Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/)** (optional if deemed irrelevant)

 - [x] kubernetes pods status: `kubectl get pods -n OPERATOR_NAMESPACE`
```
NAME                                                              READY   STATUS         RESTARTS     AGE
gpu-feature-discovery-rftgb                                       0/1     Init:0/1       0            29s
gpu-operator-6458f8894-ht5ww                                      1/1     Running        0            26m
gpu-operator-6bc456d4d5-p6znp                                     0/1     Completed      3            62d
nvidia-container-toolkit-daemonset-zwzwq                          0/1     Init:0/1       0            29s
nvidia-dcgm-exporter-wgtbd                                        0/1     Init:0/1       0            29s
nvidia-device-plugin-daemonset-n99td                              0/1     Init:0/1       0            29s
nvidia-gpu-driver-talosv1.5.0-civo-5499bdbfcf-kkdqb               0/1     ErrImagePull   0            60s
tfy-gpu-operator-node-feature-discovery-gc-6759585f99-w7k84       1/1     Running        0            5h25m
tfy-gpu-operator-node-feature-discovery-master-75b57cc786-7mt5f   1/1     Running        1 (8d ago)   12d
[...]

PR #1861, opened 2025-11-05
[bugfix] : don't render nvidiadriver CR if driver is disabled

rahulait:
fixes https://github.com/NVIDIA/gpu-operator/issues/1795

If driver is disabled, we should not render default config for nvidiaDriver. This PR adds that check.

### DIFF
--- a/deployments/gpu-operator/templates/nvidiadriver.yaml
+++ b/deployments/gpu-operator/templates/nvidiadriver.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.driver.enabled }}
 {{- if and .Values.driver.nvidiaDriverCRD.enabled .Values.driver.nvidiaDriverCRD.deployDefaultCR }}
 apiVersion: nvidia.com/v1alpha1
 kind: NVIDIADriver
@@ -123,4 +124,5 @@ spec:
     args: {{ toYaml .Values.gdrcopy.args | nindent 6 }}
     {{- end  }}
   {{- end }}
+{{- end }}
 {{- end }}


deployments/gpu-operator/templates/nvidiadriver.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.driver.nvidiaDriverCRD.enabled .Values.driver.nvidiaDriverCRD.deployDefaultCR }}
+{{- if and .Values.driver.enabled .Values.driver.nvidiaDriverCRD.enabled .Values.driver.nvidiaDriverCRD.deployDefaultCR }}
```
tariq1890: As per the documentation, the `and` function only takes two arguments

https://helm.sh/docs/chart_template_guide/function_list#and

tariq1890: Let's do it this way #1862

rahulait: It says two or more: `Returns the boolean AND of two or more arguments` unless I am reading it wrong. Also, I did tested it locally and it worked fine with more than 2 arguments.

tariq1890: Ah yes, you're right. It does take more than two arguments. I missed that from the documentation.

However, I prefer the top-level if condition as it is more readable

rahulait: ACK, updated.

tariq1890: Thank you!

cdesiniotis (approved):

PR #1866, opened 2025-11-06
Fix misspelled metric name in Prometheus rules for OpenShift

empovit:
Following https://github.com/rhobs/observability-operator/pull/926, fix the misspelled temperature units in the the metrics translation rules for OpenShift.


cdesiniotis:
Thanks @empovit!

cdesiniotis:
@empovit is it okay if we include this in our upcoming 25.10.1 patch release? Or do we need to wait longer until a new version of https://github.com/rhobs/observability-operator is released?

empovit:
@cdesiniotis it's OK to include in 25.10.1, thanks for asking!

Issue #1661, opened 2025-09-04
NVIDIADriver CR stuck NotReady after nodeSelector change


h323:
**Describe the bug**
When editing the nodeSelector field of an NVIDIADriver custom resource (CR), the resource enters a permanent NotReady state if the change does not result in any driver pods being updated. The GPU Operator keeps reconciling without making progress, repeatedly logging that the object is not ready.

**To Reproduce**
 1. Deploy an NVIDIADriver custom resource with a specific nodeSelector.
 2. Wait until it reaches the Ready state.
 3. Edit the nodeSelector so that the selector changes, but the set of matching nodes remains the same (for example, replace one matching label with another equivalent label).
 4. Observe that the NVIDIADriver CR goes into NotReady and remains stuck.

Additional scenario:
If the nodeSelector is changed to target a subset of the original nodes, Kubernetes will remove pods from the non-matching nodes, but the remaining pods are not updated. The CR still becomes stuck in NotReady.

**Expected behavior**
The GPU Operator should detect when a nodeSelector change does not result in pod updates and still mark the NVIDIADriver CR as Ready. It should handle scenarios where the DaemonSet spec changes but no pods need to be updated.

**Environment (please provide the following information):**
 - GPU Operator Version: v25.3.2
 - OS: Ubuntu 24.04
 - Kernel Version: 6.8.0-71-generic
 - Container Runtime Version: containerd 1.7.27
 - Kubernetes Distro and Version: Kubernetes v1.32.4

**Logs / References**
Excerpt from operator logs (looping indefinitely):
```
{"level":"info","ts":1756998659.7785275,"logger":"state.state-driver","msg":"Object is not ready","controller":"nvidia-driver-controller","object":{"name":"default"},"namespace":"","name":"default","reconcileID":"6b8b24a7-6f09-468e-b73c-e4757e847bb6","Kind:":"DaemonSet","Name":"nvidia-gpu-driver-ubuntu24.04-8df5fc75"}
{"level":"info","ts":1756998659.7785566,"msg":"Sync not Done for custom resource","controller":"nvidia-driver-controller","object":{"name":"default"},"namespace":"","name":"default","reconcileID":"6b8b24a7-6f09-468e-b73c-e4757e847bb6"}
[...]

PR #1868, opened 2025-11-06
fix: resolve NVIDIADriver stuck in NotReady on nodeSelector changes with OnDelete strategy

karthikvetrivel:
Solves #1661. 

## Problem
When editing the `nodeSelector` field of an NVIDIADriver CR, the resource enters a permanent NotReady state if the change doesn't result in pod updates (e.g., replacing equivalent labels). This causes infinite reconciliation loops.

## Root Cause
The readiness check required `UpdatedNumberScheduled == NumberAvailable`, but with OnDelete update strategy, pods are never auto-updated even when already on correct nodes.

## New Logic Flow Diagram

```
isDaemonSetReady()
  │
  ├─ Check: DesiredNumberScheduled != 0?
  ├─ Check: NumberUnavailable == 0?
  ├─ Check: DesiredNumberScheduled == NumberAvailable?
  │
  ├─ If OnDelete strategy:
  │   │
  │   └─> isDaemonSetReadyOnDelete()
  │        │
  │        ├─ getOwnedPods() → Get pods for DaemonSet
  │        │
  │        ├─ arePodsHealthy() → All Running + Ready?
  │        │   └─ NO → Return NotReady ❌
  │        │
  │        ├─ getLatestRevisionHash() → Get DaemonSet revision
  │        │
  │        ├─ Check pod controller-revision-hash labels
  │        │   ├─ All match? → Return Ready ✅
  │        │   │
  │        │   └─ Some outdated? → verifyNodePlacement()
  │        │        │
  │        │        ├─ For each pod:
  │        │        │   └─ Get node, check labels match nodeSelector
  │        │        │
  │        │        ├─ All on correct nodes? → Return Ready ✅
  │        │        └─ Some on wrong nodes? → Return NotReady ❌
  │
  └─ If RollingUpdate strategy:
      │
      └─> Check: UpdatedNumberScheduled == NumberAvailable?
          ├─ YES → Return Ready ✅
          └─ NO  → Return NotReady ❌
```

### Before (Buggy Behavior)
```
User changes: nodeSelector: {zone: us-east} → {region: us-east}
  ↓
DaemonSet spec updated (creates new ControllerRevision: def456)
  ↓
Kubernetes checks nodes:
  - node1 has {zone: us-east, region: us-east} ✓
  - node2 has {zone: us-east, region: us-east} ✓
  - Pods already on these nodes
  ↓
OnDelete strategy: Don't touch existing pods
  ↓
DaemonSet Status:
  - DesiredNumberScheduled: 2 ✓
  - NumberAvailable: 2 ✓
  - UpdatedNumberScheduled: 0 ✗ (pods have old revision abc123)
  ↓
Readiness Check:
  if (2 != 0 && 2 == 2 && 0 == 2) → FALSE
  ↓
Status: NotReady
  ↓
Reconcile loop: Wait 5s → Check again → Still NotReady
  ↓
♾️ INFINITE LOOP - NEVER BECOMES READY
```

### After (Fixed Behavior)
```
User changes: nodeSelector: {zone: us-east} → {region: us-east}
  ↓
DaemonSet spec updated (creates new ControllerRevision: def456)
  ↓
Kubernetes checks nodes:
  - node1 has {zone: us-east, region: us-east} ✓
  - node2 has {zone: us-east, region: us-east} ✓
  - Pods already on these nodes
  ↓
OnDelete strategy: Don't touch existing pods
  ↓
DaemonSet Status:
  - DesiredNumberScheduled: 2 ✓
  - NumberAvailable: 2 ✓
  - UpdatedNumberScheduled: 0 ✗ (pods have old revision abc123)
  ↓
NEW Readiness Check:
  ├─ OnDelete strategy detected
  ├─ isDaemonSetReadyOnDelete() called
  │   ├─ getOwnedPods() → [pod1, pod2]
  │   ├─ arePodsHealthy() → TRUE ✓
  │   ├─ getLatestRevisionHash() → "def456"
  │   ├─ Check revisions:
  │   │   - pod1: abc123 ≠ def456 (outdated)
  │   │   - pod2: abc123 ≠ def456 (outdated)
  │   ├─ Pods outdated → verifyNodePlacement() called
  │   │   ├─ pod1 on node1 → node1 has region=us-east? YES ✓
  │   │   └─ pod2 on node2 → node2 has region=us-east? YES ✓
  │   └─ All pods on correct nodes → Return TRUE
  ↓
Status: Ready ✅
  ↓
No more reconciliation needed!
```

The fix recognizes that for OnDelete strategy, outdated pod revisions are acceptable **if** pods are already on nodes matching the current nodeSelector. This indicates the change was nodeSelector-only and doesn't require pod recreation.

### DIFF
--- a/internal/state/state_skel.go
+++ b/internal/state/state_skel.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -402,7 +404,7 @@ func (s *stateSkel) getSyncState(ctx context.Context, objs []*unstructured.Unstr
 
 		// Object exists, check for Kind specific readiness
 		if found.GetKind() == "DaemonSet" {
-			if ready, err := s.isDaemonSetReady(found, reqLogger); err != nil || !ready {
+			if ready, err := s.isDaemonSetReady(ctx, found, reqLogger); err != nil || !ready {
 				reqLogger.V(consts.LogLevelInfo).Info("Object is not ready", "Kind:", obj.GetKind(), "Name", obj.GetName())
 				return SyncStateNotReady, err
 			}
@@ -413,7 +415,7 @@ func (s *stateSkel) getSyncState(ctx context.Context, objs []*unstructured.Unstr
 }
 
 // isDaemonSetReady checks if daemonset is ready
-func (s *stateSkel) isDaemonSetReady(uds *unstructured.Unstructured, reqLogger logr.Logger) (bool, error) {
+func (s *stateSkel) isDaemonSetReady(ctx context.Context, uds *unstructured.Unstructured, reqLogger logr.Logger) (bool, error) {
 	buf, err := uds.MarshalJSON()
 	if err != nil {
 		return false, fmt.Errorf("failed to marshall unstructured daemonset object: %w", err)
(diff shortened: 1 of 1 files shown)


internal/state/state_skel.go
```diff
@@ -454,3 +475,125 @@ func (s *stateSkel) checkAttributesExist(attrs nodeinfo.NodeAttributes, attrType
...
+func (s *stateSkel) verifyNodePlacement(ctx context.Context, ds *appsv1.DaemonSet, pods []corev1.Pod, reqLogger logr.Logger) (bool, error) {
+	for _, pod := range pods {
+		node := &corev1.Node{}
+		if err := s.client.Get(ctx, types.NamespacedName{Name: pod.Spec.NodeName}, node); err != nil {
+			return false, err
+		}
+
+		// Check if node matches nodeSelector
+		for key, value := range ds.Spec.Template.Spec.NodeSelector {
+			if node.Labels[key] != value {
```
rahulait, line 589: ~As a safety check, should we also check that key exists in node.Labels and then match the values?~

internal/state/state_skel.go
```diff
@@ -454,3 +475,125 @@ func (s *stateSkel) checkAttributesExist(attrs nodeinfo.NodeAttributes, attrType
...
+				reqLogger.V(consts.LogLevelInfo).Info("Pod on non-matching node",
+					"pod", pod.Name, "selector", key, "expected", value, "actual", node.Labels[key])
```
rahulait: Should we log node name as well in this log to see where the pod is running?

karthikvetrivel: Updated!

internal/state/state_skel.go
```diff
@@ -454,3 +475,125 @@ func (s *stateSkel) checkAttributesExist(attrs nodeinfo.NodeAttributes, attrType
...
+		reqLogger.V(consts.LogLevelWarning).Info("Could not get revision hash, using health-only check", "error", err)
+		return true, nil
+	}
+
+	// Check if all pods have the latest revision
+	for _, pod := range ownedPods {
+		if hash, ok := pod.Labels["controller-revision-hash"]; !ok || hash != dsRevisionHash {
+			// Pods have outdated revision - verify they're on nodes matching current nodeSelector
+			reqLogger.V(consts.LogLevelInfo).Info("Pods have outdated revision, verifying node placement")
+			return s.verifyNodePlacement(ctx, ds, ownedPods, reqLogger)
```
rajathagasthya, line 507: Is node placement is the actual thing we care about for "ready" status? If so, can we just not check for node placement regardless of revision hash on the pods?

I don't think we necessarily need to know if a pod was updated in a level-triggered reconciliation. We just need to periodically check if the final condition is true.

karthikvetrivel: This is a great point. I was originally looking at ways we can see if the pod is updated/not but that's not strictly required. I will look into updating this.

rajathagasthya: I'm wondering if this also applies to the `UpdatedNumberScheduled` check. I'm thinking the entire status check can be reduced to:

1. `DesiredNumberScheduled == NumberAvailable`, AND
2. Pods are placed on the correct nodes (or, this might be more precise: `each node selected by the nodeSelector has a pod scheduled on it`)

cc @tariq1890 @cdesiniotis to validate if my assumptions are correct.

karthikvetrivel:
One issue with this change is that if the driver image spec is updated but the pods are still on the correct nodes, the DaemonSet is marked as Ready. Prior, the DaemonSet was marked as not ready.  

The only way I think we can get around this is distinguishing between placement (i.e. node selectory, taint tolerations) changes & workload changes (image, command/args, env variables). If a placement change was made but the pod placement is analogous to the prior policy, then we keep the DaemonSet as marked for ready. For all workload changes, we would mark the DaemonSet as not ready.

karthikvetrivel:
I'm closing this PR. After discussion, it seems like this requires a larger change to nodeSelector / NVIDIADriver CR.

PR #1873, opened 2025-11-07
use PATCH to update node labels instead of UPDATE

tariq1890:
This fixes an issue seen in the ClusterPolicy reconciler when it attempts to update node labels on a cluster. 

```
Status:
  Conditions:
    Last Transition Time:  2025-08-08T01:45:52Z
    Message:               
    Reason:                Error
    Status:                False
    Type:                  Ready
    Last Transition Time:  2025-08-08T01:45:52Z
    Message:               failed to initialize ClusterPolicy controller: unable to label node 4u2g-0506 for the GPU Operator deployment, err Operation cannot be fulfilled on nodes "4u2g-0506": the object has been modified; please apply your changes to the latest version and try again
    Reason:                ReconcileFailed
    Status:                True
 ```
 
 Moving from UPDATE to PATCH should fix this issue, as PATCH is the method used under the hood with the `kubectl label node...` command and PATCH is better at handling concurrent updates in general

### DIFF
--- a/controllers/state_manager.go
+++ b/controllers/state_manager.go
@@ -492,6 +492,8 @@ func (n *ClusterPolicyController) labelGPUNodes() (bool, int, error) {
 	gpuNodesTotal := 0
 	for _, node := range list.Items {
 		node := node
+
+		nodeOriginal := node.DeepCopy()
 		// get node labels
 		labels := node.GetLabels()
 		if !clusterHasNFDLabels {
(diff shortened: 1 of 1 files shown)


tariq1890:
`k8s.io/kubernetes` [code](https://github.com/kubernetes/kubernetes/blob/5300466a5c8988b479a151ceb77f49dd00065c83/staging/src/k8s.io/kubectl/pkg/cmd/label/label.go#L351) where the Merge Patch strategy is used by the `kubectl label` command

cdesiniotis (approved):

PR #1877, opened 2025-11-10
Add VFIO Validation for vgpu-device-manager

JunAr7112:


cmd/nvidia-validator/main.go
```diff
@@ -1734,6 +1734,22 @@ func (v *VGPUDevices) validate() error {
 }
 
 func (v *VGPUDevices) runValidation() error {
+	nvpci := nvpci.New()
+	GPUDevices, err := nvpci.GetGPUs()
+	if err != nil {
+		return fmt.Errorf("error checking for GPU devices on the host: %w", err)
+	}
+	for _, device := range GPUDevices {
+		if device.SriovInfo.PhysicalFunction == nil {
```
tariq1890:
```suggestion
		if device.SriovInfo == nil || device.SriovInfo.PhysicalFunction == nil {
```

cmd/nvidia-validator/main.go
```diff
@@ -1734,6 +1734,22 @@ func (v *VGPUDevices) validate() error {
...
+			continue
+		}
+		totalVF := int(device.SriovInfo.PhysicalFunction.TotalVFs)
+		if totalVF > 0 {
+			log.Infof("Found GPU device with SR-IOV VFs: %s (TotalVFs: %d)", device.Address, totalVF)
+			return nil
```
tariq1890: Why are we returning early here?

cmd/nvidia-validator/main.go
JunAr7112: I'm not sure if this would work as device.SriovInfo is a struct value (with type SriovInfo) and not a pointer, so it cannot be nil.

cmd/nvidia-validator/main.go
JunAr7112: If VFIO is enabled, than we wouldn't need to execute the rest of the function (and we would error by default when looking for mdev devices). That's why we would return early if VF exist. if VFIO is not enabled, there would be devices under mdevBusPath := "/sys/class/mdev_bus" and we would not enter into this for loop. Instead we would follow the earlier verification code.

cdesiniotis:
Closing as we are no longer working on this PR.

PR #1884, opened 2025-11-10
Add weekly forward compatibility testing

ArangoGutierrez:
Implement automated forward compatibility tests that validate GPU Operator
against the latest published images from NVIDIA component repositories.

## Changes

**Workflow Refactoring:**
- Refactor monolithic `ci.yaml` into reusable workflow modules (`variables.yaml`, `golang-checks.yaml`, `config-checks.yaml`, `image-builds.yaml`, `e2e-tests.yaml`, `release.yaml`)

**Forward Compatibility Testing:**
- Add `forward-compatibility.yaml` workflow (weekly + manual trigger)
- Create `get-latest-images.sh` to fetch latest commit-based images with retry/backoff
- Create `generate-values-overrides.sh` to produce Helm values override file
- Use artifact-based values override approach (per @cdesiniotis review)

**E2E Test Enhancements:**
- Extend `e2e-tests.yaml` with optional `use_values_override` input
- Create `env-to-values.sh` to convert env vars to Helm values file
- Update `install-operator.sh` with proper YAML merging via `yq`
- Handle vGPU options correctly with values file approach
- Pass `VALUES_FILE` to remote test instances (per @rajathagasthya review)

**Notifications:**
- Add Slack notifications for test failures (JSON payload, configurable mentions)
- Guard notification step against missing secrets

## Components Tested
- `ghcr.io/nvidia/container-toolkit`
- `ghcr.io/nvidia/k8s-device-plugin`
- `ghcr.io/nvidia/k8s-mig-manager`

We could add other operands later, but I wanted to start with the core ones.

**Schedule:** Every Monday at 2 AM UTC

## Review Feedback Addressed
- [x] Use values override file instead of individual input vars (@cdesiniotis)
- [x] Move `operator_image`/`operator_version` to variables workflow (@cdesiniotis)
- [x] Remove unnecessary workflow inputs from forward-compatibility (@cdesiniotis)
- [x] Remove year from copyright headers (@rajathagasthya)
- [x] Pass `VALUES_FILE` to remote test instance (@rajathagasthya)
- [x] Fix Slack JSON payload format
- [x] Use `yq` for proper YAML merging
- [x] Handle vGPU with values file approach
- [x] Add retry with exponential backoff for image verification
- [x] Remove push trigger from e2e-tests (secrets not inherited)
- [x] Restore golangci-lint-action to v9
- [x] Write SSH key before SCP step
- [x] Restore Holodeck to v0.2.18
- [x] Remove unused `operator_version` input from release workflow
- [x] Guard Slack notification against missing secrets

### DIFF
--- a/.github/scripts/get-latest-images.sh
+++ b/.github/scripts/get-latest-images.sh
@@ -32,7 +32,7 @@ fi
 case "${COMPONENT}" in
     toolkit)
         IMAGE_REPO="ghcr.io/nvidia/container-toolkit"
-        GITHUB_REPO="NVIDIA/container-toolkit"
+        GITHUB_REPO="NVIDIA/nvidia-container-toolkit"
         ;;
     device-plugin)
         IMAGE_REPO="ghcr.io/nvidia/k8s-device-plugin"
@@ -40,7 +40,7 @@ case "${COMPONENT}" in
         ;;
     mig-manager)
         IMAGE_REPO="ghcr.io/nvidia/k8s-mig-manager"
-        GITHUB_REPO="NVIDIA/k8s-mig-manager"
+        GITHUB_REPO="NVIDIA/mig-parted"
         ;;
     *)
         echo "Error: Unknown component '${COMPONENT}'" >&2
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,7 +80,10 @@ jobs:
     with:
       operator_image: ${{ needs.variables.outputs.operator_image }}
       operator_version: ${{ needs.variables.outputs.operator_version }}
-    secrets: inherit
+    secrets:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      AWS_SSH_KEY: ${{ secrets.AWS_SSH_KEY }}
 
   release:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
@@ -92,4 +95,3 @@ jobs:
     with:
       commit_short_sha: ${{ needs.variables.outputs.commit_short_sha }}
       operator_image_base: ${{ needs.variables.outputs.operator_image_base }}
-    secrets: inherit
--- a/.github/workflows/forward-compatibility.yaml
+++ b/.github/workflows/forward-compatibility.yaml
@@ -69,12 +69,20 @@ jobs:
 
   run-e2e-tests:
     needs: [fetch-latest-images]
+    permissions:
+      contents: read
+      id-token: write
     uses: ./.github/workflows/e2e-tests.yaml
     with:
       operator_image: ghcr.io/nvidia/gpu-operator
       operator_version: main-latest
       use_values_override: true
-    secrets: inherit
+    secrets:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      AWS_SSH_KEY: ${{ secrets.AWS_SSH_KEY }}
+      SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+      SLACK_CHANNEL_ID: ${{ secrets.SLACK_CHANNEL_ID }}
 
   notify-failure:
     runs-on: ubuntu-latest
--- a/tests/scripts/install-operator.sh
+++ b/tests/scripts/install-operator.sh
@@ -34,30 +34,26 @@ if [[ "${USE_VALUES_FILE}" == "true" ]]; then
 	# Generate a temporary values file from environment variables
 	# and merge it with the provided VALUES_FILE
 	TEMP_ENV_VALUES=$(mktemp)
-	trap 'rm -f "${TEMP_ENV_VALUES:-}" "${COMBINED_VALUES:-}"' EXIT
+	trap 'rm -f "${TEMP_ENV_VALUES:-}"' EXIT
 	${SCRIPT_DIR}/env-to-values.sh "${TEMP_ENV_VALUES}"
 
-	# If VALUES_FILE exists, merge it with env-generated values using yq
-	# Otherwise just use the env-generated values
+	# If VALUES_FILE exists, use both files with Helm's native multi-file
+	# support (-f). Helm merges values in order, with later files taking
+	# precedence — so env-generated values override the provided file.
 	if [[ -f "${VALUES_FILE}" ]]; then
 		echo ""
 		echo "Using provided values file: ${VALUES_FILE}"
 		cat "${VALUES_FILE}"
 		echo ""
-		echo "Merged with environment-based values:"
+		echo "Environment-based values (takes precedence):"
 		cat "${TEMP_ENV_VALUES}"
 
-		# Create a combined values file using yq for proper YAML merging
-		COMBINED_VALUES=$(mktemp)
-		if ! command -v yq >/dev/null 2>&1; then
-			echo "Error: yq is required to merge YAML values files but was not found in PATH." >&2
-			echo "Install yq: https://github.com/mikefarah/yq" >&2
-			exit 1
-		fi
-		# yq merges YAML properly, with later files taking precedence
-		yq ea '. as $item ireduce ({}; . * $item )' "${VALUES_FILE}" "${TEMP_ENV_VALUES}" > "${COMBINED_VALUES}"
-		VALUES_FILE="${COMBINED_VALUES}"
+		# Pass both files to helm; the env values file comes second
+		# so its values take precedence over the override file.
+		EXTRA_VALUES_FILES="-f ${VALUES_FILE}"
+		VALUES_FILE="${TEMP_ENV_VALUES}"
 	else
+		EXTRA_VALUES_FILES=""
 		VALUES_FILE="${TEMP_ENV_VALUES}"
 	fi
 	
@@ -113,7 +109,9 @@ if [[ "${USE_VALUES_FILE}" == "true" ]]; then
 	echo "Using values file approach: ${VALUES_FILE}"
 	${HELM} install ${PROJECT_DIR}/deployments/gpu-operator --generate-name \
 		-n "${TEST_NAMESPACE}" \
+		${EXTRA_VALUES_FILES} \
 		-f "${VALUES_FILE}" \
+		${OPERATOR_OPTIONS} \
 		--wait
 else
 	echo "Using --set flags approach"


ArangoGutierrez:
PING @rahulait @rajathagasthya @tariq1890

.github/workflows/forward-compatibility.yaml
```diff
@@ -0,0 +1,131 @@
...
+          curl -sSLo bin/regctl https://github.com/regclient/regclient/releases/download/${REGCTL_VERSION}/regctl-linux-amd64
+          chmod +x bin/regctl
+          echo "$(pwd)/bin" >> $GITHUB_PATH
+      
+      - name: Get latest component images
+        id: images
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Use workflow_dispatch inputs if provided, otherwise fetch latest
```
cdesiniotis: Question: when would someone set these variables as inputs? If the scope of this workflow is to **always** test the top-of-tree images for all operands, then maybe we can remove the input variables and move the entire `run:` block into a shell script? Thoughts?

.github/workflows/forward-compatibility.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+  run-e2e-tests:
+    needs: [fetch-latest-images]
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      operator_image: ghcr.io/nvidia/gpu-operator
+      operator_version: main-latest
+      toolkit_image: ${{ needs.fetch-latest-images.outputs.toolkit_image }}
+      device_plugin_image: ${{ needs.fetch-latest-images.outputs.device_plugin_image }}
+      mig_manager_image: ${{ needs.fetch-latest-images.outputs.mig_manager_image }}
```
cdesiniotis: I understand the current `tests/scripts/install-operator.sh` has separate input variables for the various operand images, but would it simplify our CI definition if we passed a values override file instead? Just an idea -- the previous step `fetch-latest-images` can craft the values override file that points to all the latest operand images. This file is passed as input to the `run-e2e-tests` step which launches our e2e test script with this overrides file. Thoughts? The reason I am suggesting this is because I see a lot of boilerplate in `forward-compatibility.yaml` and `e2e-tests.yaml` for declaring all of these input variables.

.github/workflows/forward-compatibility.yaml
```diff
@@ -0,0 +1,131 @@
...
+        required: false
+        type: string
+      device_plugin_image:
+        description: 'Override device-plugin image'
+        required: false
+        type: string
+      mig_manager_image:
+        description: 'Override mig-manager image'
+        required: false
+        type: string
```
cdesiniotis: Why do these variables need to be inputs? Aren't we always resolving the images in this workflow itself?

.github/workflows/ci.yaml
```diff
@@ -20,429 +20,48 @@ on:
...
+      label_image_source: ${{ needs.variables.outputs.label_image_source }}
+      push_on_build: ${{ needs.variables.outputs.push_on_build }}
+      operator_image_base: ${{ needs.variables.outputs.operator_image_base }}
+
+  e2e-tests:
+    needs: [variables, image-builds]
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      operator_image: ${{ needs.variables.outputs.operator_image_base }}
+      operator_version: ${{ needs.variables.outputs.operator_version }}
```
cdesiniotis, line 82: Question -- could we potentially move this to the `variables` workflow itself? As in, make `operator_image` and `operator_version` as outputs of the `variables` worflow?

.github/workflows/ci.yaml
```diff
@@ -20,429 +20,48 @@ on:
...
+    secrets: inherit
+
+  release:
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    needs: [variables, e2e-tests]
+    uses: ./.github/workflows/release.yaml
+    with:
+      commit_short_sha: ${{ needs.variables.outputs.commit_short_sha }}
+      operator_version: ${{ needs.variables.outputs.operator_version }}
+      operator_image_base: ${{ needs.variables.outputs.operator_image_base }}
```
cdesiniotis, line 97: Same comment as above -- is this really needed? Can we move these variables as outputs of the `variables` workflow itself?

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,183 @@
...
+
+jobs:
+  variables:
+    uses: ./.github/workflows/variables.yaml
+    with:
+      operator_image: ${{ inputs.operator_image }}
+      operator_version: ${{ inputs.operator_version }}
+      toolkit_image: ${{ inputs.toolkit_image }}
+      device_plugin_image: ${{ inputs.device_plugin_image }}
+      mig_manager_image: ${{ inputs.mig_manager_image }}
```
cdesiniotis: What is the rationale for passing these variables as input to the `variables` workflow?

.github/workflows/forward-compatibility.yaml
ArangoGutierrez: My intention is to allow manual triggering of the workflow in case we want to test a very specific image without having to wait for the "weekly" run. But I'll set it in a way that even during manual triggers, we will simply fetch all the latests images from operands.

.github/workflows/forward-compatibility.yaml
ArangoGutierrez: I took a stab at this comment, please let me know what you think

.github/workflows/forward-compatibility.yaml
ArangoGutierrez: removed

.github/workflows/ci.yaml
ArangoGutierrez: good catch, done!

.github/workflows/ci.yaml
ArangoGutierrez: same, done!

ArangoGutierrez:
@cdesiniotis @rahulait @rahulait @tariq1890  PTAL

.github/scripts/generate-values-overrides.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 NVIDIA CORPORATION
```
rajathagasthya: nit: remove year.

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,173 @@
...
+        id: get_public_dns_name
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq '.status.properties[] | select(.name == "public-dns-name") | .value' /github/workspace/.cache/holodeck.yaml
+      - name: Set test environment
+        run: |
+          echo "instance_hostname=ubuntu@${{ steps.get_public_dns_name.outputs.result }}" >> $GITHUB_ENV
+          echo "private_key=${{ github.workspace }}/key.pem" >> $GITHUB_ENV
+          if [[ "${{ inputs.use_values_override }}" == "true" ]]; then
+            echo "VALUES_FILE=${{ github.workspace }}/values-overrides.yaml" >> $GITHUB_ENV
```
rajathagasthya: I may be missing something, but how does the remote instance know about `VALUES_FILE`? Do we need to pass that environment variable here? https://github.com/NVIDIA/gpu-operator/blob/6531839893e387ac31569f55bde8c56841429313/tests/local.sh#L43

ArangoGutierrez:
@rajathagasthya @cdesiniotis PTAL

rajathagasthya:
@ArangoGutierrez Are you able to manually trigger a forward compatibility job in this PR or does this need to merge first?

ArangoGutierrez:
@rajathagasthya I did some last minute fixes before merging, but looks like some E2E are broken - Not related to this PR.

rajathagasthya:
Forward compatibility was tested in #2248 with a pull request trigger. Successful run: https://github.com/NVIDIA/gpu-operator/actions/runs/23505641426

.github/scripts/get-latest-images.sh
```diff
@@ -0,0 +1,105 @@
...
+if ! command -v regctl &> /dev/null; then
+    echo "Error: regctl not found. Please install regctl first." >&2
+    exit 1
+fi
+
+# Map component names to GHCR image repositories and GitHub source repositories
+case "${COMPONENT}" in
+    toolkit)
+        IMAGE_REPO="ghcr.io/nvidia/container-toolkit"
+        GITHUB_REPO="NVIDIA/container-toolkit"
```
rahulait: Even though it works, to keep it same as what we have upstream, we should fix the repo name here. Its `NVIDIA/nvidia-container-toolkit`

rajathagasthya: Updated.

rahulait:
Overall, LGTM

.github/workflows/ci.yaml
```diff
@@ -20,435 +20,76 @@ on:
...
-          CHANNELS: "stable"
-        run: |
-          make push-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ github.ref_name }}-latest
-          make push-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ env.COMMIT_SHORT_SHA }}
+      packages: write
+    uses: ./.github/workflows/release.yaml
+    with:
+      commit_short_sha: ${{ needs.variables.outputs.commit_short_sha }}
+      operator_image_base: ${{ needs.variables.outputs.operator_image_base }}
+    secrets: inherit
```
rahulait: Instead of passing all the repo/org secrets to the called workflow, we should restrict it to only needed ones.

.github/workflows/ci.yaml
```diff
@@ -20,435 +20,76 @@ on:
...
-          retention-days: 15
-
-  e2e-tests-nvidiadriver:
-    needs: [build-multi-arch-images]
-    runs-on: linux-amd64-cpu4
+    uses: ./.github/workflows/e2e-tests.yaml
+    with:
+      operator_image: ${{ needs.variables.outputs.operator_image }}
+      operator_version: ${{ needs.variables.outputs.operator_version }}
+    secrets: inherit
```
rahulait: Same here, lets use least-privilege and pass only secrets which are required

.github/workflows/ci.yaml
ArangoGutierrez: Good call. Removed `secrets: inherit` from the release job entirely — it only uses `secrets.GITHUB_TOKEN` which is automatic. For e2e-tests I enumerated the three required AWS secrets explicitly.

.github/workflows/ci.yaml
ArangoGutierrez: Done — e2e-tests now gets only `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SSH_KEY`. Also applied the same pattern to forward-compatibility.yaml (with the Slack secrets added there since that workflow uses them for failure notifications).

PR #1898, opened 2025-11-14
Tag bundle image with current commit sha

shivakunv:
Currently, the GPU Operator image configured in the bundle is ghcr.io/nvidia/gpu-operator:main-latest (see [here](https://github.com/NVIDIA/gpu-operator/blob/833f83b71dece80b00922e9d1df102df1fee4241/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L882)). This can be problematic for two reasons. First, it is difficult to deduce which commit SHA the operator image / OLM bundle was built from. Therefore, it is difficult to infer which commit SHA introduced a test failure. Second, the gpu-operator deployment runs with `imagePullPolicy=IfNotPresent` by default. This means if the gpu-operator image (with the main-latest) tag is already cached on the node, the latest image may not be pulled / tested. To address these shortcomings, we should generate an OLM bundle that is using SHA-tagged images of gpu-operator.
We should update the GitHub Actions job that builds + pushes an OLM bundle image as follows:

1) Before building the bundle image, it should update all gpu-operator image tags in the CSV file to refer to the commit SHA. E.g. ghcr.io/nvidia/gpu-operator:commit-short-sha
2) When pushing the OLM bundle image, we should push two tags – one tag will be main-latest and the other will be commit-short-sha

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -438,11 +438,20 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set environment variables
+        id: vars
+        run: |
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
+      - name: Update bundle CSV
+        run: |
+          sed -i '' -e 's|ghcr.io/nvidia/gpu-operator:[^ "]*|ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}|g' bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+          echo "Bundle CSV updated successfully"
       - name: Build bundle-image
         env:
-          BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
+          IMAGE_NAME: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle"
           VERSION: ""
           DEFAULT_CHANNEL: "stable"
           CHANNELS: "stable"
         run: |
-          make push-bundle-image
+          make push-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ github.ref_name }}-latest
+          make push-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ env.COMMIT_SHORT_SHA }}


shivakunv:
passed pipeline: 

https://github.com/NVIDIA/gpu-operator/actions/runs/19452460548

shivakunv:
PTAL @cdesiniotis @tariq1890

.github/workflows/ci.yaml
```diff
@@ -438,9 +438,15 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set environment variables
```
tariq1890, line 441: Why create a new step here? Can't we just refer to the `${GITHUB_SHA:0:8}` var directly?

.github/workflows/ci.yaml
```diff
@@ -438,9 +438,15 @@ jobs:
...
+        id: vars
+        run: |
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
       - name: Build bundle-image
         env:
           BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
+          BUNDLE_IMAGE_TAG: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ env.COMMIT_SHORT_SHA }}"
+          OPERATOR_IMAGE_TAG: "ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}"
```
tariq1890:
```suggestion
          OPERATOR_IMAGE_TAG: "ghcr.io/nvidia/gpu-operator:${GITHUB_SHA:0:8}"
```

Makefile
```diff
@@ -60,6 +60,9 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # BUNDLE_IMAGE defines the image:tag used for the bundle.
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMAGE=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMAGE ?= gpu-operator-bundle:$(VERSION)
+BUNDLE_IMAGE_TAG ?= gpu-operator-bundle:$(VERSION)
```
tariq1890: it's the same value as `BUNDLE_IMAGE` and I don't see the value of `BUNDLE_IMAGE_TAG` being changed anywhere else.

.github/workflows/ci.yaml
shivakunv: we cannot use GITHUB_SHA directly. I will give it a try with ${{ github.sha[0:8] }}

Makefile
shivakunv: steps: ci.yaml :-  
      - name: Build bundle-image
        env:
          BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
          BUNDLE_IMAGE_TAG: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ env.COMMIT_SHORT_SHA }}"
          OPERATOR_IMAGE_TAG: "ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}"
          VERSION: ""
          DEFAULT_CHANNEL: "stable"
          CHANNELS: "stable"
        run: |
          make push-bundle-image 
          
  is using it

.github/workflows/ci.yaml
shivakunv: verified.
${GITHUB_SHA:0:8} and ${{ github.sha[0:8] }} cannot be referenced inside the env: of github action.

.github/workflows/ci.yaml
shivakunv: ${GITHUB_SHA:0:8} and ${{ github.sha[0:8] }} cannot be referenced inside the env: of github action.

Makefile
```diff
@@ -126,16 +129,23 @@ bundle: manifests install-tools
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	operator-sdk bundle validate ./bundle
 
+# update the bundle image tag in csv file
+update-bundle-csv:
```
rajathagasthya: Can we just not set the operator image version during bundle generation here? https://github.com/NVIDIA/gpu-operator/blob/e0ba4ab484d0c1ca64427089471454b3b11b32f4/Makefile#L125

We could just do this instead of replacing contents of the CSV file:

```bash
cd config/manager && $(KUSTOMIZE) edit set image gpu-operator=$(IMAGE_WITH_COMMIT_SHA)
```

The bundle should reference an operator image tagged with commit SHA for immutability. It doesn't need to reference `gpu-operator:main-latest`. Then we tag and push the bundle itself with both `main-latest` and `COMMIT_SHA`.

Makefile
```diff
@@ -60,6 +60,9 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # BUNDLE_IMAGE defines the image:tag used for the bundle.
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMAGE=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMAGE ?= gpu-operator-bundle:$(VERSION)
+BUNDLE_IMAGE_TAG ?= gpu-operator-bundle:$(VERSION)
+BUNDLE_CSV_FILE ?= bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+OPERATOR_IMAGE_TAG ?= gpu-operator:$(VERSION)
```
rajathagasthya: The variable names are confusing. I suggest refactoring these variables like this to reuse image names and append versions as necessary:


```suggestion
BUNDLE_IMAGE_NAME ?= gpu-operator-bundle
BUNDLE_VERSION ?= $(VERSION)
BUNDLE_GIT_TAG ?= # Set in CI

BUNDLE_IMAGE := $(BUNDLE_IMAGE_NAME):$(BUNDLE_VERSION)
BUNDLE_IMAGE_GIT := $(if $(BUNDLE_GIT_TAG),$(BUNDLE_IMAGE_NAME):$(BUNDLE_GIT_TAG))

OPERATOR_IMAGE_NAME ?= gpu-operator
OPERATOR_VERSION ?= $(VERSION)

OPERATOR_IMAGE := $(OPERATOR_IMAGE_NAME):$(OPERATOR_VERSION)
```

Then set the appropriate variables in the CI job env:
```bash
BUNDLE_IMAGE_NAME: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle"
BUNDLE_VERSION: "${{ github.ref_name }}-latest"
BUNDLE_GIT_TAG: "${{ env.COMMIT_SHORT_SHA }}"
OPERATOR_IMAGE_NAME: "ghcr.io/nvidia/gpu-operator"
OPERATOR_VERSION: "${{ env.COMMIT_SHORT_SHA }}"
```

Makefile
```diff
@@ -126,16 +129,23 @@ bundle: manifests install-tools
...
+	@echo "Bundle CSV updated successfully"
+
 # Build the bundle image.
-build-bundle-image:
+build-bundle-image: update-bundle-csv
 	$(DOCKER) build \
 	--build-arg DEFAULT_CHANNEL=$(DEFAULT_CHANNEL) \
 	--build-arg GIT_COMMIT=$(GIT_COMMIT) \
-	-f docker/bundle.Dockerfile -t $(BUNDLE_IMAGE) .
+	-f docker/bundle.Dockerfile -t $(BUNDLE_IMAGE) -t $(BUNDLE_IMAGE_TAG) .
```
rajathagasthya:
```suggestion
		-f docker/bundle.Dockerfile \
	    -t $(BUNDLE_IMAGE) \
	    $(if $(BUNDLE_IMAGE_GIT),-t $(BUNDLE_IMAGE_GIT)) \
	    .
```

Makefile
```diff
@@ -126,16 +129,23 @@ bundle: manifests install-tools
...
 
 # Push the bundle image.
 push-bundle-image: build-bundle-image
 	$(DOCKER) push $(BUNDLE_IMAGE)
+	$(DOCKER) push $(BUNDLE_IMAGE_TAG)
```
rajathagasthya:
```suggestion
	$(if $(BUNDLE_IMAGE_GIT),$(DOCKER) push $(BUNDLE_IMAGE_GIT))
```

Makefile
shivakunv: Kustomize designed for (templating K8s resources).
 
 Issues:- 
 1. `$(KUSTOMIZE) edit set image gpu-operator=$(IMAGE_WITH_COMMIT_SHA)` Won't Solve the Problem:
  _It edits config/manager/kustomization.yaml  , not CSV files_ 


2. Multiple Locations in CSV That Need Updates
```
L190: containerImage: ghcr.io/nvidia/gpu-operator:main-latest
L202: image: ghcr.io/nvidia/gpu-operator:main-latest
L224: image: ghcr.io/nvidia/gpu-operator:main-latest
L882: image: ghcr.io/nvidia/gpu-operator:main-latest
L920: value: "ghcr.io/nvidia/gpu-operator:main-latest"
```

kustomize image transformation has different behavior for each:- 

```
Line 190: metadata.annotations.containerImage - an annotation string
Line 202: spec.relatedImages[0].image - array
Line 224: spec.relatedImages[12].image - array
Line 882: spec.install.spec.deployments[0].spec.template.spec.containers[0].image - actual container image
Line 920: spec.install.spec.deployments[0].spec.template.spec.containers[0].env[].value - environment variable 
```


3: Kustomize Image Transformer has Limited Scope
Kustomize's image transformer (images: field) ONLY replaces:
  `spec.template.spec.containers[].image fields and spec.template.spec.initContainers[].image fields`
  _not annotation strings like containerImage
  not arbitrary YAML fields likee relatedImages[]
  not environment variable values_

Sed approach:- 
It's a post-processing step that ensures the final image references are correct .
Simple & Clear: One line that does exactly what we need

Makefile
shivakunv: Thanks, fair enough. 
modified as per your comments

shivakunv:
@rajathagasthya : PTAL

shivakunv:
passed pipeline : https://github.com/NVIDIA/gpu-operator/actions/runs/20061702250

Makefile
rajathagasthya: Understood. My concern is this approach leaves a modified file in the repo.

Makefile
```diff
@@ -126,16 +134,26 @@ bundle: manifests install-tools
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	operator-sdk bundle validate ./bundle
 
+# Update the bundle CSV file with operator image tag
+update-bundle-csv:
```
rajathagasthya: @shivakunv Can we move these changes out of the Makefile and into the CI workflow before building the bundle?

Essentially, we want to keep these changes contained in CI scripts. We don't need to provide a way for anyone to update this file locally when building the bundle and having a dirty git tree.

shivakunv: done. now it is part of ci file

shivakunv:
passed pipeline: https://github.com/NVIDIA/gpu-operator/actions/runs/20156772775/job/57861546344

.github/workflows/ci.yaml
```diff
@@ -438,9 +438,21 @@ jobs:
...
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
+      - name: Update bundle CSV
+        run: |
+          sed -i'' -e 's|ghcr.io/nvidia/gpu-operator:[^ "]*|ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}|g' bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+          echo "Bundle CSV updated successfully"
       - name: Build bundle-image
         env:
-          BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
+          BUNDLE_IMAGE_NAME: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle"
+          BUNDLE_VERSION: "bundlecommitsha-latest"
```
rajathagasthya: This should be `${{ github.ref_name }}-latest`.

shivakunv: Thanks . forgot to remove feature branch

shivakunv:
Thanks @rajathagasthya 
PTAL @tariq1890 @cdesiniotis

.github/workflows/ci.yaml
```diff
@@ -438,9 +438,21 @@ jobs:
...
+          sed -i'' -e 's|ghcr.io/nvidia/gpu-operator:[^ "]*|ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}|g' bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+          echo "Bundle CSV updated successfully"
       - name: Build bundle-image
         env:
-          BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
+          BUNDLE_IMAGE_NAME: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle"
+          BUNDLE_VERSION: "${{ github.ref_name }}-latest"
+          BUNDLE_GIT_TAG: "${{ env.COMMIT_SHORT_SHA }}"
+          OPERATOR_IMAGE_NAME: "ghcr.io/nvidia/gpu-operator"
+          OPERATOR_VERSION: "${{ env.COMMIT_SHORT_SHA }}"
```
cdesiniotis: Adding all of these environment variables (and extra logic in the Makefile) seems unnecessary for the task at hand. Can we simplify this somehow?

What if we defined two steps in the `push-gpu-operator-bundle-image` job -- the first step would push the `main-latest` tag and the second step would push the `<sha>` tag? No changes would be required to the Makefile. 


Alternatively, we could update the `run` script to something like below:
```
make push-bundle-image BUNDLE_IMAGE=$IMAGE_NAME:${{ github.ref_name}}-latest
make push-bundle-image BUNDLE_IMAGE=$IMAGE_NAME:${{. env.COMMIT_SHORT_SHA }}
```

shivakunv: done
agree @cdesiniotis , its simple and clear. 

passed pipeline:- https://github.com/NVIDIA/gpu-operator/actions/runs/20256810896/job/58161057160

rajathagasthya (approved):

.github/workflows/ci.yaml
```diff
@@ -438,11 +438,20 @@ jobs:
...
       - name: Build bundle-image
         env:
-          BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
+          IMAGE_NAME: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle"
           VERSION: ""
           DEFAULT_CHANNEL: "stable"
           CHANNELS: "stable"
         run: |
-          make push-bundle-image
+          make build-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ github.ref_name}}-latest
```
rajathagasthya: This has to be `push-bundle-image`, not `build-bundle-image`.

shivakunv: ah. yes

.github/workflows/ci.yaml
```diff
@@ -438,11 +438,20 @@ jobs:
...
       - name: Build bundle-image
         env:
-          BUNDLE_IMAGE: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle:${{ github.ref_name }}-latest"
+          IMAGE_NAME: "ghcr.io/nvidia/gpu-operator/gpu-operator-bundle"
           VERSION: ""
           DEFAULT_CHANNEL: "stable"
           CHANNELS: "stable"
         run: |
-          make push-bundle-image
+          make push-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ github.ref_name}}-latest
```
rajathagasthya: Small nit:
```suggestion
          make push-bundle-image BUNDLE_IMAGE=${IMAGE_NAME}:${{ github.ref_name }}-latest
```

shivakunv: Thanks

.github/workflows/ci.yaml
```diff
@@ -438,11 +438,20 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set environment variables
+        id: vars
+        run: |
+          echo "COMMIT_SHORT_SHA=${GITHUB_SHA:0:8}" >> $GITHUB_ENV
+      - name: Update bundle CSV
+        run: |
+          sed -i'' -e 's|ghcr.io/nvidia/gpu-operator:[^ "]*|ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}|g' bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```
tariq1890: Shouldn't there be a space?
```suggestion
          sed -i '' -e 's|ghcr.io/nvidia/gpu-operator:[^ "]*|ghcr.io/nvidia/gpu-operator:${{ env.COMMIT_SHORT_SHA }}|g' bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```

rajathagasthya: `sed -i''` works as is. It's usually done for Mac OS compatibility. The `-i''` option works for both BSD sed (MacOS) and GNU sed (Linux), but it's not necessary in the CI script since we only use Linux runners.

shivakunv: It works on mac as well . only issue is with the -e option on macos, it creates a -e file locally, which is known behavior.

tariq1890: We have discussed internally; let's add the space between as suggested above. That change makes it portable across multiple `sed` implementations

shivakunv: done

tariq1890 (approved):

Issue #754, opened 2024-06-12
User "system:serviceaccount:default:gpu-operator" cannot list resource "daemonsets" in API group "apps" at the cluster scope with Helm Template


Li357:
### 1. Quick Debug Information
* OS/Version(e.g. RHEL8.6, Ubuntu22.04): Ubuntu 22.04
* Kernel Version: 5.4.0-177-generic
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker): containerd
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): k8s
* GPU Operator Version: 24.3.0

### 2. Issue or feature description
I'm using Kustomize's helmCharts to install `gpu-operator` (via `helm template` then apply) but for some reason getting

```
E0612 21:52:49.269498       1 reflector.go:147] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers.go:105: Failed to watch *v1.DaemonSet: failed to list *v1.DaemonSet: daemonsets.apps is forbidden: User "system:serviceaccount:default:gpu-operator" cannot list resource "daemonsets" in API group "apps" at the cluster scope
```
in my `gpu-operator ` pod. Also, the `gpu-operator` pod is not in the namespace specified by Helm, but for some reason put in the default namespace. This doesn't make sense as the cluster role is correctly configured [here](https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/templates/clusterrole.yaml#L68-L72)? I've verified this makes cluster role is correct in `helm template`.

### 3. Steps to reproduce the issue

```
helm template --generate-name --namespace test --create-namespace nvidia/gpu-operator | kubectl apply -f -
kubectl get pods # for some reason gpu-operator is now in the default namespace instead of the provided namespace?
kubectl logs <gpu-operator-podname>
```

I get that I should be installing through `helm install`, but I want to use this with Kustomize, which internally uses `helm template` to inflate charts. Why is the install behavior different than the template behavior (I didn't see any hooks?)

mike-ensor:
I have the same issue. I am using Kustomize to generate the configuration. The role and role binding both appear to be correct

rahulait:
This issue has been open for over 90 days without recent updates, and the context may now be outdated.

This has been fixed in the newer versions of gpu-operator. Given that gpu-operator 24.3.0 is EOL now, I would encourage you to try latest version and see if you still see this issue.

[...]

PR #1904, opened 2025-11-17
[fix] : add namespace when templating helm chart

rahulait:
Fixes https://github.com/NVIDIA/gpu-operator/issues/754

### DIFF
--- a/deployments/gpu-operator/templates/operator.yaml
+++ b/deployments/gpu-operator/templates/operator.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: gpu-operator
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "gpu-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: "gpu-operator"
--- a/deployments/gpu-operator/templates/role.yaml
+++ b/deployments/gpu-operator/templates/role.yaml
@@ -2,6 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: gpu-operator
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "gpu-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: "gpu-operator"
--- a/deployments/gpu-operator/templates/rolebinding.yaml
+++ b/deployments/gpu-operator/templates/rolebinding.yaml
@@ -2,6 +2,7 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: gpu-operator
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "gpu-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: "gpu-operator"
--- a/deployments/gpu-operator/templates/serviceaccount.yaml
+++ b/deployments/gpu-operator/templates/serviceaccount.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: gpu-operator
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "gpu-operator.labels" . | nindent 4 }}
     app.kubernetes.io/component: "gpu-operator"


deployments/gpu-operator/templates/_helpers.tpl
```diff
@@ -37,6 +41,7 @@ Common labels
 
 {{- define "gpu-operator.labels" -}}
 app.kubernetes.io/name: {{ include "gpu-operator.name" . }}
+app.kubernetes.io/namespace: {{ include "gpu-operator.namespace" . }}
```
tariq1890: let's remove this. The namespace showing up in the object metadata should suffice

deployments/gpu-operator/templates/_helpers.tpl
```diff
@@ -6,6 +6,10 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "gpu-operator.namespace" -}}
+{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
```
tariq1890: I believe this char limit only really applies to labels, since we don't want to apply it as a label, let's remove this macro

deployments/gpu-operator/templates/_helpers.tpl
cdesiniotis: I think I agree with Tariq, although not a strong opinion.

One observation, if we include the namespace label here, it will get added to cluster-scoped objects, like the ClusterRole and ClusterRoleBinding that our helm chart renders (which we probably don't want).

rahulait: Sure, updated.

deployments/gpu-operator/templates/_helpers.tpl
rahulait: I see error when creating namespace of more than 63 chars
```
$ kubectl create namespace y7it8p362pexwav96uzuh33u4m06m3wr5mueeic58mq46mg984vhi9urbt5cg9ycjchvu8
The Namespace "y7it8p362pexwav96uzuh33u4m06m3wr5mueeic58mq46mg984vhi9urbt5cg9ycjchvu8" is invalid:
* metadata.name: Invalid value: "y7it8p362pexwav96uzuh33u4m06m3wr5mueeic58mq46mg984vhi9urbt5cg9ycjchvu8": must be no more than 63 characters
* metadata.labels: Invalid value: "y7it8p362pexwav96uzuh33u4m06m3wr5mueeic58mq46mg984vhi9urbt5cg9ycjchvu8": must be no more than 63 characters
```

tariq1890: The namespace is a field we should treat as high fidelity. Truncation for labels is fine, but we should really just fail fast if a namespace is greater than 63 chars. Whatever a user sets for the namespace (a valid one) is what the user should get

rahulait: Ack, updated.

deployments/gpu-operator/templates/_helpers.tpl
```diff
@@ -6,6 +6,10 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "gpu-operator.namespace" -}}
+{{- default .Release.Namespace .Values.namespaceOverride -}}
```
tariq1890: Please note that we don't really use `.Value.namespaceOverride` anywhere in the main chart. It is only referenced in the NFD subchart. Let's just remove this macro and set the `{{ .Release.Namespace }}` inline

rahulait: ACK, removed it.

deployments/gpu-operator/templates/_helpers.tpl
```diff
@@ -6,6 +6,10 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{- define "gpu-operator.namespace" -}}
```
tariq1890: Let's remove this template macro please

deployments/gpu-operator/templates/operator.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: gpu-operator
+  namespace: {{ include "gpu-operator.namespace" . }}
```
tariq1890:
```suggestion
  namespace: {{ include .Release.namespace }}
```

deployments/gpu-operator/templates/_helpers.tpl
rahulait: If we keep the macro, it will enable us to just do changes in one place if we want to allow namespaceOverride or if we want to apply any custom logic for namespace than changing each file where its used. Currently its just using the default as .Release.namespace and that makes the macro look like its not useful.

Let me know your thoughts and I am ok to remove it as well.

tariq1890: I don't foresee us using `namespaceOverride`.  We always want to deploy the gpu-operator resources on one namespace instead of supporting deployments across multiple namespaces. The former is more secure as access-control and other namespace-scoped privileges are managed a lot easier when all the gpu-operator resources colocated in a single namespace

rahulait: updated.

deployments/gpu-operator/templates/operator.yaml
rahulait: done.

Issue #1622, opened 2025-08-22
Adding a new node with a different os-image and kernel-version ends up mixing up the nvidia-driver-daemonset(s)


ScottWatsonWork:
**Describe the bug**
I was running a node with os-image of 1592.11 and kernel-version 6.6.95-cloud-amd64 with driver version 570.172.08

Now I have added a new node to my cluster and it will be running with os-image 1877.2 and kernel version  6.12.40-cloud-amd64 and the same driver version 570.172.08 

There has been no change to the cluster policy as we are running the same driver version.

I see the gpu operator will create a new daemonset with what appears to be the correct name with the right os-image and kernel-version in the name.

```
k get ds -l app=nvidia-driver-daemonset
NAME                                                            DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                                                                                          AGE
nvidia-driver-daemonset-6.12.40-cloud-amd64-gardenlinux1877.2   1         1         0       0            0           feature.node.kubernetes.io/kernel-version.full=6.12.40-cloud-amd64,nvidia.com/gpu.deploy.driver=true   31m
nvidia-driver-daemonset-6.6.95-cloud-amd64-gardenlinux1592.11   1         1         1       0            1           feature.node.kubernetes.io/kernel-version.full=6.6.95-cloud-amd64,nvidia.com/gpu.deploy.driver=true    3h2m

```

The problem is that the image name that is associate with the daemonset is wrong.  The example below has the ds for os-image 1592.11 but the image that it is trying to pull is 1877.2

```
kubectl  get ds nvidia-driver-daemonset-6.6.95-cloud-amd64-gardenlinux1592.11 -o yaml | yq '.spec.template.spec.containers[0].image'
icx-data-ai-devops-docker.common.repositories.cloud.sap/nvidia/driver:570.172.08-6.6.95-cloud-amd64-gardenlinux1877.2
```

The daemonset will be redeployed by the operatore and the image value can be inverted again. 

Checking just now shows that it has now gone the other way

```
kubectl  get ds nvidia-driver-daemonset-6.6.95-cloud-amd64-gardenlinux**1592.11** -o yaml | yq '.spec.template.spec.containers[0].image'
icx-data-ai-devops-docker.common.repositories.cloud.sap/nvidia/driver:570.172.08-6.6.95-cloud-amd64-gardenlinux**1592.11**

[...]

PR #1917, opened 2025-11-20
Fix shared state bug causing incorrect driver images in multi-node-pool clusters

karthikvetrivel:
Solves #1622 

When a cluster has nodes with different OS/kernel versions (multiple node pools), the GPU operator creates separate DaemonSets for each pool. However, DaemonSets were being assigned incorrect driver container images - the image tags didn't match the kernel/OS versions in the DaemonSet names. 

### Root Cause 

`getDriverSpec()` was using `spec := &cr.Spec` instead of creating a deep copy. This caused all node pools to share the same `NVIDIADriverSpec` pointer in memory. When processing multiple node pools in a loop, mutations (like setting NodeSelector) from later iterations overwrote values from earlier iterations, leading to inconsistent DaemonSet configurations.

### DIFF
--- a/internal/state/driver.go
+++ b/internal/state/driver.go
@@ -545,7 +545,7 @@ func getDriverSpec(cr *nvidiav1alpha1.NVIDIADriver, nodePool nodePool) (*driverS
 	nvidiaDriverName := getDriverName(cr, nodePool.getOS())
 	nvidiaDriverAppName := getDriverAppName(cr, nodePool)
 
-	spec := &cr.Spec
+	spec := cr.Spec.DeepCopy()
 	imagePath, err := getDriverImagePath(spec, nodePool)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get driver image path: %v", err)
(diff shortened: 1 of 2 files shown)


karthikvetrivel:
> LGTM. Can you also share some test results after the change to see that its working fine now with the change.

Yes! I mocked the same setup where: 
- Node 1: Garden Linux 1592.11, kernel 6.6.95-cloud-amd64
- Node 2: Garden Linux 1877.2, kernel 6.12.40-cloud-amd64
- Driver version: 570.172.08

**Before fix:**
- DaemonSet for kernel 6.6.95 had wrong image tag (1877.2 instead of 1592.11)
- Both specs shared the same pointer (0xc00046ab90)

**After fix:**
- Each DaemonSet correctly matched its target kernel/OS version:
  - `nvidia-driver-daemonset-6.6.95-cloud-amd64-gardenlinux1592.11` uses `driver:570.172.08-6.6.95-cloud-amd64-gardenlinux1592.11`
  - `nvidia-driver-daemonset-6.12.40-cloud-amd64-gardenlinux1877.2` uses `driver:570.172.08-6.12.40-cloud-amd64-gardenlinux1877.2`
 - Each spec has a unique pointer (e.g., 0xc0004a2000 vs 0xc0004a2400)

I also pushed a new test to `driver_test.go` just now.

PR #1933, opened 2025-11-24
fix: Restart vgpu-device-manager pods when workload config changes

karthikvetrivel:
## Problem
`nvidia-vgpu-device-manager` pods were not restarting when node workload  config label changed from `vm-passthrough` to `vm-vgpu`, even with `sandboxWorkloads.enabled=true`.

## Solution
Add workload config hash annotation to the DaemonSet pod template. When workload configs change, the hash changes, causing Kubernetes to detect a spec change and trigger a rolling update.

This approach is inspired by `TransformKataManager`, which uses a similar hash-based mechanism to trigger pod restarts when Kata config changes.

## Changes
- Compute hash of node workload configs in `TransformVGPUDeviceManager`
- Set hash as pod template annotation (inspired by KataManager pattern)
- Add test verifying hash changes trigger restart

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -42,6 +42,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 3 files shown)


controllers/object_controls.go
```diff
@@ -2162,6 +2169,26 @@ func TransformVGPUDeviceManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPoli
...
+	if err := n.client.List(n.ctx, nodeList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			vgpuDeviceManagerLabelKey: "true",
+		}),
+	}); err != nil {
+		n.logger.Error(err, "Failed to list nodes for workload config hash")
+		return nil
+	}
+
+	workloadConfigs := make(map[string]string)
```
rahulait, line 2184:
```suggestion
	workloadConfigs := make(map[string]string, len(nodeList.Items))
```

karthikvetrivel:
Closing due to additional context provided on the problem. There is a race condition where the vgpu-device-manager is starting before the vgpu-manager finishes its install — in particular, before VFs are created on the GPUs. Another PR for this bug will come up soon.

PR #1939, opened 2025-11-26
Run nvidia-smi after modules are loaded in driver ds startup probe

cdesiniotis:
This commit eliminates the race condition where the startup probe in the driver daemonset runs after the kernel modules are built (and installed) but before the modules are loaded into the kernel. In this case, the invocation of nvidia-smi (by the startup probe) is what is actually loading the nvidia kernel module and not the modprobe we perform in our driver container scripts. As a result, the nvidia driver will be loaded with a default configuration -- none of the custom kernel module parameters provided by users (via a configmap) or set by our driver container will get applied.

### DIFF
--- a/assets/state-driver/0500_daemonset.yaml
+++ b/assets/state-driver/0500_daemonset.yaml
@@ -137,7 +137,7 @@ spec:
         startupProbe:
           exec:
             command:
-              [sh, -c, 'nvidia-smi && touch /run/nvidia/validations/.driver-ctr-ready']
+              [sh, -c, '[ -f /sys/module/nvidia/refcnt ] && nvidia-smi && touch /run/nvidia/validations/.driver-ctr-ready']
           initialDelaySeconds: 60
           failureThreshold: 120
           successThreshold: 1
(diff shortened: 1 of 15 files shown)


tariq1890:
I've created #1940 to resolve the build failure in the PR pipeline

karthikvetrivel (approved):
LGTM, Chris. I assume you've had the chance to test it under the same circumstances as the original bug?

shivamerla:
Good catch! do we need to add similar checks wherever we are running driver validation (i.e as in the driver-validation init-container of the toolkit)?

cdesiniotis:
> do we need to add similar checks wherever we are running driver validation (i.e as in the driver-validation init-container of the toolkit)?

@shivamerla Good question. The `driver-validation` currently waits on the presence of `/run/nvidia/validations/.driver-ctr-ready` before running nvidia-smi: https://github.com/NVIDIA/gpu-operator/blob/526cc24056a463996277a0e165195e363e13a0d9/cmd/nvidia-validator/main.go#L667-L678. Since the `.driver-ctr-ready` status file is created in the driver pod's startup probe, this race condition does not exist in this case and no changes are required.

tariq1890 (approved):

cdesiniotis:
> I assume you've had the chance to test it under the same circumstances as the original bug?

@karthikvetrivel Not yet, but based on my analysis this change is needed regardless. This is definitely a race condition that has always existed. Now whether this change actually resolves the reported issue won't be entirely confirmed until we get access to their environment.

PR #1942, opened 2025-11-26
Add cluster upgrade support by including MOFED dependency in RDMA scenarios

tginer:
This commit adds the proper GPU driver wait for the MOFED driver to be ready so RDMA APIs are available when driver is recompiled. This ensures the operator supports cluster upgrades, i.e., kernel version upgrades.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1109,7 +1109,9 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 	}
 
 	// add nodeSelector for MOFED wait label when GPUDirect RDMA is enabled
-	if config.Driver.GPUDirectRDMA != nil && config.Driver.GPUDirectRDMA.IsEnabled() {
+	if config.Driver.GPUDirectRDMA != nil &&
+		config.Driver.GPUDirectRDMA.IsEnabled() &&
+		!config.Driver.GPUDirectRDMA.IsHostMOFED() {
 		if obj.Spec.Template.Spec.NodeSelector == nil {
 			obj.Spec.Template.Spec.NodeSelector = make(map[string]string)
 		}
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1108,6 +1108,13 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 		setContainerEnv(driverToolkitContainer, "DRIVER_CONFIG_DIGEST", configDigest)
 	}
 
+	// add nodeSelector for MOFED wait label when GPUDirect RDMA is enabled
+	if config.Driver.GPUDirectRDMA != nil && config.Driver.GPUDirectRDMA.IsEnabled() {
+		if obj.Spec.Template.Spec.NodeSelector == nil {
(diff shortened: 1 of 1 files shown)


controllers/object_controls.go
```diff
@@ -1044,6 +1044,14 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 			return fmt.Errorf("ERROR: failed to transform the pre-compiled Driver Daemonset: %s", err)
 		}
 	}
+
+	// add nodeSelector for MOFED wait label when GPUDirect RDMA is enabled
+	if config.Driver.GPUDirectRDMA != nil && config.Driver.GPUDirectRDMA.IsEnabled() {
+		if obj.Spec.Template.Spec.NodeSelector == nil {
+			obj.Spec.Template.Spec.NodeSelector = make(map[string]string)
+		}
+		obj.Spec.Template.Spec.NodeSelector["network.nvidia.com/operator.mofed.wait"] = "false"
```
cdesiniotis, line 1118: This assumes the Network Operator is deployed (and MOFED is installed via their driver container). Will this always be the case? For example, on Ubuntu the MOFED driver can be preinstalled.

cdesiniotis, line 1118: Does this have any implications on what versions of Network Operator we support / work with? e.g. starting with what version of Network Operator do they add this label to nodes?

tariq1890: What if we skip applying this label if useHostMofed is set to true?

cdesiniotis: Yes, that should address my comment.

tariq1890: This was the [commit](https://github.com/Mellanox/network-operator/commit/e49c4d1a35880fc4d76fd57a4ed5ce9b8a06cb2d) that introduced this label. So it's been around for a long time

controllers/object_controls.go
```diff
@@ -1044,6 +1044,14 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 			return fmt.Errorf("ERROR: failed to transform the pre-compiled Driver Daemonset: %s", err)
 		}
 	}
+
+	// add nodeSelector for MOFED wait label when GPUDirect RDMA is enabled
+	if config.Driver.GPUDirectRDMA != nil && config.Driver.GPUDirectRDMA.IsEnabled() {
```
```suggestion
	if config.Driver.GPUDirectRDMA != nil &&
		config.Driver.GPUDirectRDMA.IsEnabled() &&
		!config.Driver.GPUDirectRDMA.IsHostMOFED() {
```

controllers/object_controls.go
```diff
@@ -1108,6 +1108,15 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 		setContainerEnv(driverToolkitContainer, "DRIVER_CONFIG_DIGEST", configDigest)
 	}
 
+	// add nodeSelector for MOFED wait label when GPUDirect RDMA is enabled
```
cdesiniotis, line 1111: Let's expand this comment to describe why this change is needed / what problem it solves.

controllers/object_controls.go
rajatchopra: I wish the label would come from the network operator library. And not magic coded.

PR #1950, opened 2025-12-01
[state-toolkit] add support for mounting runtime NRI sockets

tariq1890:
Add support for enabling NRI Plugin and leveraging the same when injecting GPU management CDI devices to all of the operands.

#### Summary of changes

- Adds a new field in the ClusterPolicy CRD `.spec.cdi.nriPluginEnabled | (true|false) default: false`
- NRI Plugin will only be enabled if CDI is enabled. 
- If NRI Plugin is enabled, then the operands are rendered with the NRI annotation instead of the `nvidia` runtime class in their pod spec after the Transform methods are run.
- This PR assumes that when NRI Plugin is enabled, the toolkit will no longer need to configure the runtime config TOML and restart the runtime process. Changes to enable both workflows will be addressed in another PR if this is required


controllers/object_controls.go
```diff
@@ -2559,13 +2579,13 @@ func getRuntimeClassName(config *gpuv1.ClusterPolicySpec) string {
...
-	runtimeClassName := getRuntimeClassName(config)
-	podSpec.RuntimeClassName = &runtimeClassName
-}
+// func setRuntimeClassName(podSpec *corev1.PodSpec, config *gpuv1.ClusterPolicySpec, runtime gpuv1.Runtime) {
+//	if !config.CDI.IsEnabled() && runtime == gpuv1.CRIO {
+//		return
+//	}
+//	runtimeClassName := getRuntimeClassName(config)
+//	podSpec.RuntimeClassName = &runtimeClassName
+//}
```
elezar: With this and the calls to this function commented out, this PR should still be in draft -- or is the plan to DEFINITELY replace this?

tariq1890: You're right, the PR should be in draft. The intention is not to replace this by default

tariq1890: Converted to draft, thanks for calling this out

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1722,6 +1722,14 @@ type CDIConfigSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Deprecated: This field is no longer used"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch,urn:alm:descriptor:com.tectonic.ui:hidden"
 	Default *bool `json:"default,omitempty"`
+
+	// NRIPluginEnabled indicates whether the NRI Plugin should be run as a means of injecting CDI devices to containers
```
cdesiniotis: Question -- do we want to specifically mention management applications? Or is that leaking unnecessary implementation details?
```suggestion
	// NRIPluginEnabled indicates whether an NRI Plugin should be run as a means of injecting CDI devices to management containers
```

cmd/nvidia-validator/main.go
```diff
@@ -1601,6 +1609,21 @@ func (c *CUDA) runWorkload() error {
...
+		cdiManagementDevice = "management.nvidia.com/gpu=all"
+	)
+
+	if nriMode {
+		annotations := pod.Annotations
+		if len(annotations) == 0 {
+			annotations = make(map[string]string)
+		}
+		annotationKey := fmt.Sprintf("%s/container.cuda-validation", nriAnnotationDomain)
+		annotations[annotationKey] = cdiManagementDevice
```
cdesiniotis: Question -- what about always including this annotation in the [pod template](https://github.com/NVIDIA/gpu-operator/blob/54ff621fa2919880eb4846a18dc002b66b8bbeb5/validator/manifests/cuda-workload-validation.yaml) itself? Is this an alternative we should consider?

controllers/object_controls.go
```diff
@@ -1308,6 +1336,10 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
 		setContainerEnv(toolkitMainContainer, CRIOConfigModeEnvName, "hook")
 	}
 
+	if config.CDI.IsNRIPluginEnabled() {
+		setContainerEnv(toolkitMainContainer, CDIEnableNRIPlugin, "true")
+	}
+
```
cdesiniotis: Question -- should we put this code snippet in the `transformToolkitCtrForCDI()` method?

cmd/nvidia-validator/main.go
tariq1890: Not a bad idea actually. So even if NRI is disabled, it would just be no-op. Is that what you had in mind?

controllers/object_controls.go
tariq1890: yeah that makes sense. Let me look into that

api/nvidia/v1/clusterpolicy_types.go
tariq1890: No, I think it's fine

cmd/nvidia-validator/main.go
cdesiniotis: Yes, exactly. Moving this to the pod template removes the need for an additional flag / envvar in the validator code.

tariq1890: Sounds good, I'll remove that code

api/nvidia/v1/clusterpolicy_types.go
tariq1890: Done

cmd/nvidia-validator/main.go
tariq1890: Done

controllers/object_controls.go
tariq1890: Done

PR #1967, opened 2025-12-04
Upgrade Controller: filter node watch to only upgrade state label changes

karthikvetrivel:
## Summary
Replace generic label change predicate with a specific predicate that only watches for `nvidia.com/gpu-driver-upgrade-state` label changes on nodes.

## Problem
The upgrade controller was using `TypedLabelChangedPredicate` which triggered reconciliation on ANY node label change, causing unnecessary reconciliations.

## Solution
Implement `upgradeStateLabelPredicate` that compares only the upgrade state label between old and new node objects, ignoring all other label changes.

## Testing
- Built and deployed to live cluster
- Verified changing non-upgrade labels (`test-label`, `nvidia.com/some-other-label`, etc.) does NOT trigger reconciliation
- Verified changing `nvidia.com/gpu-driver-upgrade-state` label DOES trigger immediate reconciliation
- Confirmed upgrade state machine still processes correctly through all states

### DIFF
--- a/controllers/state_manager_test.go
+++ b/controllers/state_manager_test.go
@@ -69,3 +69,120 @@ func TestGetRuntimeString(t *testing.T) {
 		})
 	}
 }
+
+func TestIsValidWorkloadConfig(t *testing.T) {
+	tests := []struct {
+		config string
+		want   bool
+	}{
+		{gpuWorkloadConfigContainer, true}, {gpuWorkloadConfigVMPassthrough, true}, {gpuWorkloadConfigVMVgpu, true},
+		{"invalid", false}, {"", false},
+	}
+	for _, tc := range tests {
+		if got := isValidWorkloadConfig(tc.config); got != tc.want {
+			t.Errorf("isValidWorkloadConfig(%q) = %v, want %v", tc.config, got, tc.want)
+		}
+	}
+}
+
+func TestHasOperandsDisabled(t *testing.T) {
+	tests := []struct {
+		labels map[string]string
+		want   bool
+	}{
+		{map[string]string{commonOperandsLabelKey: "false"}, true},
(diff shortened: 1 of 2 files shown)


rajathagasthya (approved):

controllers/state_manager_test.go
```diff
@@ -69,3 +69,120 @@ func TestGetRuntimeString(t *testing.T) {
...
+	}
+}
+
+func TestHasOperandsDisabled(t *testing.T) {
+	tests := []struct {
+		labels map[string]string
+		want   bool
+	}{
+		{map[string]string{"nvidia.com/gpu.deploy.operands": "false"}, true},
+		{map[string]string{"nvidia.com/gpu.deploy.operands": "true"}, false},
```
cdesiniotis: Do we want to use the constant for this label instead?

cdesiniotis: I have the same comment for the other test functions. Do we want to use the constants for the various label keys?

karthikvetrivel: Updated for labels that have existing constants! Thanks for mentioning this, I didn't see them.

controllers/state_manager_test.go
```diff
@@ -69,3 +69,120 @@ func TestGetRuntimeString(t *testing.T) {
...
+		}
+	}
+}
+
+func TestHasNFDLabels(t *testing.T) {
+	tests := []struct {
+		labels map[string]string
+		want   bool
+	}{
+		{map[string]string{"feature.node.kubernetes.io/cpu": "true"}, true},
```
cdesiniotis: Let's use `nfdLabelPrefix` here

karthikvetrivel: Updated this and a few other constants as well. Let me know if there are any others I've missed.

PR #1973, opened 2025-12-08
WIP: Updated NodeFeatureRules for newer chips

zvonkok:
All Hopper and Hopper+ architectures support CC.

### DIFF
--- a/deployments/gpu-operator/templates/nodefeaturerules.yaml
+++ b/deployments/gpu-operator/templates/nodefeaturerules.yaml
@@ -44,64 +44,37 @@ spec:
         matchExpressions:
           sev.enabled:
             op: Exists
-    - name: "NVIDIA H100"
+    - name: "NVIDIA Hopper GPU"
       labels:
-        "nvidia.com/gpu.H100": "true"
         "nvidia.com/gpu.family": "hopper"
       matchFeatures:
         - feature: pci.device
           matchExpressions:
             vendor: {op: In, value: ["10de"]}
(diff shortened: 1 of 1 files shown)


deployments/gpu-operator/templates/nodefeaturerules.yaml
msanft, line 1: We recently tweaked the NodeFeatureRules to work with a B200 (should be GB100, iirc) cluster we received, and the device ID is `0x2901`, which wouldn't be matched by the config in this PR. any chance this can be changed?

manuelh-dev:
We will likely want to adjust the nvidia-cc-manager too: https://github.com/NVIDIA/k8s-cc-manager/blob/45f968d36e9c4bc39e497bd87aa6f5265648d0dd/cmd/main.go#L113, see the `CC_CAPABLE_DEVICE_IDS` variable. The [draft PR](https://github.com/NVIDIA/gpu-operator/pull/1798/files) I have closed shows that this variable is also still used in this code base (see `assets/state-cc-manager/0500_daemonset.yaml` and `deployments/gpu-operator/values.yaml`

cdesiniotis:
Is there an alternative solution that does not involve maintaining an allowlist of PCI device ids? Is there way to detect whether a GPU is CC capable (when running on the node)?

zvonkok:
We need to rework the whole CC detection ... on my list.

zvonkok:
Closing this as we're using CC-manager to set the correct CC labels.

PR #1981, opened 2025-12-10
fix(controller): do not apply driver upgrade annotation when driver is disabled

### LINKED ISSUE #1277: nvidia.com/gpu-driver-upgrade-enabled: "true"  even when the driver.enabled=false
I installed the gpu-operator chart like this:

```
  helm upgrade --install --wait gpu-operator-1739580441 \
       --namespace gpu-operator --create-namespace \
       nvidia/gpu-operator \
       --version=v24.9.2 \
       --set driver.enabled=false,toolkit.enabled=false
```

After everything was settled, I still have the `nvidia.com/gpu-driver-upgrade-enabled: "true"` annotation on the node.

I hope it won't be effective.

friedrice99:
The applyDriverAutoUpgradeAnnotation() function was applying the nvidia.com/gpu-driver-upgrade-enabled annotation to GPU nodes even when driver.enabled=false. 
This occurred because the function only checked if driver.upgradePolicy.autoUpgrade was true, without verifying that the driver component itself was enabled.

This fix adds a check for Driver.IsEnabled() before applying the annotation, ensuring it is only set when:
1. Driver is enabled
2. Auto-upgrade policy exists and is enabled
3. Sandbox workloads are disabled

Added unit tests to validate the fix and prevent regression.

Fixes #1277

### DIFF
--- a/controllers/state_manager.go
+++ b/controllers/state_manager.go
@@ -438,7 +438,7 @@ func (n *ClusterPolicyController) applyDriverAutoUpgradeAnnotation() error {
 		updateRequired := false
 		value := "true"
 		annotationValue, annotationExists := node.Annotations[driverAutoUpgradeAnnotationKey]
-		if n.singleton.Spec.Driver.IsEnabled() &&
+		if (n.singleton.Spec.Driver.IsEnabled() || n.singleton.Spec.Driver.UseNvidiaDriverCRDType()) &&
 			n.singleton.Spec.Driver.UpgradePolicy != nil &&
 			n.singleton.Spec.Driver.UpgradePolicy.AutoUpgrade &&
 			!n.sandboxEnabled {


controllers/state_manager_test.go
```diff
@@ -186,3 +187,70 @@ func TestHasMIGCapableGPU(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			// Create a mock ClusterPolicy
+			clusterPolicy := &gpuv1.ClusterPolicy{}
+			clusterPolicy.Spec.Driver.Enabled = &tc.driverEnabled
+			clusterPolicy.Spec.Driver.UpgradePolicy = &upgrade_v1alpha1.DriverUpgradePolicySpec{
+				AutoUpgrade: tc.autoUpgradeEnabled,
+			}
+
+			// Simulate the logic from applyDriverAutoUpgradeAnnotation
+			shouldApply := clusterPolicy.Spec.Driver.IsEnabled() &&
```
rajathagasthya: I don't think the tests add much value if we're just duplicating logic here. You will need to either call `applyDriverAutoUpgradeAnnotation()` with appropriate fake client/objects set up (preferred) or remove the test entirely.

rajathagasthya: I removed the test so we can get this PR merged.

PR #1999, opened 2025-12-16
Fix chart's devicePlugin.config fields evaluated to objects instead of strings when empty

### LINKED ISSUE #1993: Helm render error with Agocd and devicePlugin when emtpy default profile
**Describe the bug**

> one or more synchronization tasks completed unsuccessfully, reason: ClusterPolicy.nvidia.com "cluster-policy" is invalid: spec.devicePlugin.config.default: Invalid value: "null": spec.devicePlugin.config.default in body must be of type string: "null" 

```yaml
devicePlugin:
  config:
    create: true
    name: deviceplugin-config
    default: "" # <- setting empty or "null" default profile for GPU slicing will render as null and break the validation ArgoCD applies
```

**To Reproduce**
in the Helm values file create any devicePlugin config, but set default profile to an empty string

**Expected behavior**

not setting default profile or setting it to empty string it should render as an empty string not `null` for the clusterpolicy to be valid

 - GPU Operator Version: [e.g. v25.10.1]

NoOverflow:
Since these fields are not quoted when templated, using any special value such as `null`, or an empty string may result in them being evaluated as objects. 

This causes an error when evaluating against the CRD definition while applying the object.
As there may be more fields with the same behavior, I would probably expand this PR if needed.

While I agree that these fields shouldn't be empty, not quoting those values anyway may lead to issues and misunderstandings.

## Testing process

- Create a test values file with an empty value for the config name, and a special name.
```YAML
devicePlugin:
  config:
    name: "null"
    create: true
    default: ""
```

### Pre-fix:
- Observe that the `devicePlugin.config` fields are not formatted correctly
```Bash
 nefast@sapphire $ helm template nvidia-gpu-operator . -f test.yaml | grep devicePlugin -A 10
  devicePlugin:
    enabled: true
    repository: nvcr.io/nvidia
    image: k8s-device-plugin
    version: "v0.18.1"
    imagePullPolicy: IfNotPresent
    config:
      name: null
      default: 
```

### Post-fix:
- Observe that the `devicePlugin.config` fields are templated properly and that it passes server validation
```Bash
nefast@sapphire $ helm template nvidia-gpu-operator . -f test.yaml | grep devicePlugin -A 8
  devicePlugin:
    enabled: true
    repository: nvcr.io/nvidia
    image: k8s-device-plugin
    version: "v0.18.1"
    imagePullPolicy: IfNotPresent
    config:
      name: "null"
      default: ""
```

Closes #1993

### DIFF
--- a/deployments/gpu-operator/templates/clusterpolicy.yaml
+++ b/deployments/gpu-operator/templates/clusterpolicy.yaml
@@ -471,7 +471,7 @@ spec:
     {{- end }}
     {{- if .Values.devicePlugin.config.name }}
     config:
-      name: {{ .Values.devicePlugin.config.name }}
+      name: {{ .Values.devicePlugin.config.name | quote }}
       default: {{ .Values.devicePlugin.config.default | quote }}
     {{- end }}
   dcgm:


rajathagasthya:
@NoOverflow Thanks for working on this! Could you add a section in the PR description to describe how this was tested?

NoOverflow:
Hi @rajathagasthya, no problem, sorry for the oversight.

NoOverflow:
@rajathagasthya Updated. Let me know if that's fine :smile:

rajathagasthya (approved):

PR #2002, opened 2025-12-17
Wait for parent devices (VFs) before applying vGPU config

karthikvetrivel:
### Problem

When switching workloads from `vm-passthrough` to `vm-vgpu`, the vgpu-device-manager can start before vgpu-manager finishes creating Virtual Functions (VFs), causing "no parent devices found" errors. This race condition is more pronounced on nodes with many GPUs (8+) where VF creation takes longer.

### Root Cause

The `vgpu-manager-ready` status file is created when `nvidia-smi` works, but VF creation via `sriov-manage -e` happens after that. The vgpu-device-manager needs VFs to exist before it can configure vGPU types on them.

The `sriov-manage` script enables VFs by writing `sriov_totalvfs` to `sriov_numvfs` for each GPU sequentially. On systems with 8 GPUs (160 total VFs), this can take 30+ seconds to complete.

### Solution

Add `waitForVFs()` to `VGPUManager.validate()` in the nvidia-validator. After the `nvidia-smi` check passes, poll SR-IOV sysfs via the `nvpci` library until all VFs are created:

- Find all NVIDIA PCI devices that are SR-IOV Physical Functions (`SriovInfo.IsPF()`)
- Poll until `sum(NumVFs) == sum(TotalVFs)` across all GPUs

### Testing

Verified on DGX A100 (8 GPUs, 160 VFs):

```
time="2025-12-18T00:32:43+02:00" level=info msg="Waiting for VFs: 0/160 enabled across 8 GPU(s)"
time="2025-12-18T00:32:45+02:00" level=info msg="Waiting for VFs: 0/160 enabled across 8 GPU(s)"
time="2025-12-18T00:32:47+02:00" level=info msg="Waiting for VFs: 0/160 enabled across 8 GPU(s)"
time="2025-12-18T00:32:52+02:00" level=info msg="All 160 VF(s) enabled on 8 NVIDIA GPU(s)"
```

Tested by building a standalone binary that calls waitForVFs() directly, then enabling VFs via `sriov_numvfs` while the test was running to simulate the vgpu-manager's `sriov-manage -e` behavior.

VFs were enabled by binding each GPU to `pci-pf-stub` and writing to `sriov_numvfs`.

### DIFF
--- a/cmd/nvidia-validator/main.go
+++ b/cmd/nvidia-validator/main.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -224,6 +225,8 @@ const (
 	wslNvidiaSMIPath = "/usr/lib/wsl/lib/nvidia-smi"
 	// shell indicates what shell to use when invoking commands in a subprocess
 	shell = "sh"
+	// defaultVFWaitTimeout is the default timeout for waiting for VFs to be created
+	defaultVFWaitTimeout = 5 * time.Minute
 )
 
 func main() {
@@ -1591,6 +1594,11 @@ func (v *VGPUManager) validate() error {
 		return err
 	}
 
+	log.Info("Waiting for VFs to be available...")
+	if err := waitForVFs(ctx, defaultVFWaitTimeout); err != nil {
+		return fmt.Errorf("vGPU Manager VFs not ready: %w", err)
+	}
+
 	statusFile := vGPUManagerStatusFile
 	if hostDriver {
 		statusFile = hostVGPUManagerStatusFile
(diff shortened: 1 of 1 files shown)


cmd/nvidia-validator/main.go
```diff
@@ -1622,6 +1630,44 @@ func (v *VGPUManager) runValidation(silent bool) (hostDriver bool, err error) {
...
+		gpus, err := nvpciLib.GetGPUs()
+		if err != nil {
+			log.Warnf("Error getting GPUs: %v", err)
+			return false, nil
+		}
+
+		var totalExpected, totalEnabled uint64
+		var sriovGPUCount int
+		for _, gpu := range gpus {
+			if gpu.SriovInfo.IsPF() {
```
tariq1890: let's move `gpu.SriovInfo` to a variable.

karthikvetrivel: Updated.

cmd/nvidia-validator/main.go
```diff
@@ -1622,6 +1630,45 @@ func (v *VGPUManager) runValidation(silent bool) (hostDriver bool, err error) {
...
+			log.Warnf("Error getting GPUs: %v", err)
+			return false, nil
+		}
+
+		var totalExpected, totalEnabled uint64
+		var sriovGPUCount int
+		for _, gpu := range gpus {
+			sriovInfo := gpu.SriovInfo
+			if sriovInfo.IsPF() {
+				sriovGPUCount++
```
cdesiniotis: nit: `sriovGPUCount` may not be the best name for this variable, since as currently implemented this count is just the total GPU count -- it does not distinguish between GPUs that support SRIOV and those that do not support SRIOV.

karthikvetrivel: Updated to be `pfCount` since we're counting Physical Functions.

cdesiniotis (approved):

PR #2003, opened 2025-12-18
[fix] : avoid setting empty value for migmanager config's name

rahulait:
## Description

This PR avoids rendering migManager config's name if it is set to empty. Default value for it based on CRD specification is [default-mig-parted-config](https://github.com/NVIDIA/gpu-operator/blob/v25.10.1/api/nvidia/v1/clusterpolicy_types.go#L1427)

Currently, on helm template, it gets rendered as:
```
$ helm template nvidiagpu  -n gpu-operator --create-namespace nvidia/gpu-operator --version v25.10.1 -f values.yaml | grep -A8 migManager
  migManager:
    enabled: true
    repository: nvcr.io/nvidia/cloud-native
    image: k8s-mig-manager
    version: "v0.13.1"
    imagePullPolicy: IfNotPresent
    config:
      name:                    <---- value is empty here
      default: "all-disabled"
```

If Argocd is used to install gpu-operator via helm, it sometimes causes sync diffs on subsequent sync runs as the applied value gets updated to default value and the new rendered value is empty/null.
Screenshot 2025-12-17 at 3 56 52 PM

This PR avoids setting name if its empty and hence argocd doesn't find the value changed on subsequent syncs as the key is not rendered if its empty.

## Checklist

## Testing

- [ ] Unit tests (`make coverage`)
- [x] Manual cluster testing (describe below)
- [ ] N/A or Other (docs, CI config, etc.)

**Test details:**
Manually tested the change by rendering the updated helm chart. It correctly skips the key. Example:

```
helm template nvidiagpu  -n gpu-operator --create-namespace deployments/gpu-operator/ -f values.yaml | grep -A8 migManager
  migManager:
    enabled: true
    repository: nvcr.io/nvidia/cloud-native
    image: k8s-mig-manager
    version: "v0.13.1"
    imagePullPolicy: IfNotPresent
    config:
      default: "all-disabled"
    gpuClientsConfig:
```

### DIFF
--- a/deployments/gpu-operator/templates/clusterpolicy.yaml
+++ b/deployments/gpu-operator/templates/clusterpolicy.yaml
@@ -596,7 +596,9 @@ spec:
     {{- end }}
     {{- if .Values.migManager.config }}
     config:
+      {{- if .Values.migManager.config.name }}
       name: {{ .Values.migManager.config.name }}
+      {{- end }}
       default: {{ .Values.migManager.config.default | quote }}
     {{- end }}
     {{- if .Values.migManager.gpuClientsConfig }}


tariq1890:
How about defaulting to `default-mig-parted-config` instead? This way, it's the same outcome as the CRD default is `default-mig-parted-config`. Falling back to the default value is more reader-friendly anyway since most users aren't going to read the CRD spec

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -596,7 +596,9 @@ spec:
     {{- end }}
     {{- if .Values.migManager.config }}
     config:
+      {{- if .Values.migManager.config.name }}
       name: {{ .Values.migManager.config.name }}
```
tariq1890, line 600:
```suggestion
      name: {{ .Values.migManager.config.name | default "default-mig-parted-config" }}
```

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -596,7 +596,9 @@ spec:
     {{- end }}
     {{- if .Values.migManager.config }}
```
tariq1890, line 597: @cdesiniotis Thoughts on this?
```suggestion
    {{- if .Values.migManager.config.create }}
```

deployments/gpu-operator/templates/clusterpolicy.yaml
tariq1890: Resolving this comment as setting this default can backfire when we may want to away from the static configmap model

tariq1890 (approved):

PR #2020, opened 2026-01-06
Consolidate e2e module by switching helm client to CLI and removing nested module

karthikvetrivel:
## Overview

This change eliminates the separate Go module at `tests/e2e/` by:

1. Replacing the `github.com/mittwald/go-helm-client` library with direct `helm` binary execution
2. Removing the nested module files to allow Dependabot to manage all dependencies uniformly

This prevents Helm-related CVEs from being introduced into the GPU Operator images while fixing dependency drift issues.

**Tested:**

1. E2E tests compile: `cd tests/e2e && go build ./...`
2. Unit tests still pass: `make unit-test`
3. Module validation succeeds: `make validate-modules`
5. E2E tests work (requires cluster): `cd tests/e2e && make e2e-test`


tests/e2e/helpers/operator.go
```diff
@@ -92,37 +41,81 @@ type ChartOptions struct {
...
+	cmd := exec.CommandContext(ctx, "helm", args...)
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("error installing operator: %w: %s", err, stderr.String())
+	}
+
+	releaseName, err := parseReleaseName(stdout.String())
```
rajathagasthya: We only need to parse release name from the output when we use `--generate-name` flag. Is there a need to use that option in tests? Can we just not generate a release name (with a random suffix or something) and use it everywhere?

karthikvetrivel: Good idea, implemented.

PR #2024, opened 2026-01-08
[GPU Driver Container Resiliency] Compute and propagate driver configuration digest via environment variables

karthikvetrivel:
Relevant PRs: 

- https://github.com/NVIDIA/gpu-driver-container/pull/534
- https://github.com/NVIDIA/k8s-driver-manager/pull/148

## Description

- Adds driver configuration digest computation and propagation to enable fast path optimization
- Computes digest in operator controller and propagates via `DRIVER_CONFIG_DIGEST` environment variable to driver containers and k8s-driver-manager.
- Enables downstream components to detect unchanged configuration and skip driver uninstall/reinstall (see relevant PRs)

## Checklist

### DIFF
--- a/cmd/nvidia-validator/main.go
+++ b/cmd/nvidia-validator/main.go
@@ -150,6 +150,8 @@ const (
 	defaultMetricsPort = 0
 	// hostDevCharPath indicates the path in the container where the host '/dev/char' directory is mounted to
 	hostDevCharPath = "/host-dev-char"
+	// nvidiaModuleRefcntPath is the path to check if the nvidia kernel module is loaded
+	nvidiaModuleRefcntPath = "/sys/module/nvidia/refcnt"
 	// defaultDriverInstallDir indicates the default path on the host where the driver container installation is made available
 	defaultDriverInstallDir = "/run/nvidia/driver"
 	// defaultDriverInstallDirCtrPath indicates the default path where the NVIDIA driver install dir is mounted in the container
@@ -916,6 +918,20 @@ func (d *Driver) createStatusFile(driverInfo driverInfo) error {
 	return createStatusFileWithContent(outputDirFlag+"/"+driverStatusFile, statusFileContent)
 }
 
+// isNvidiaModuleLoaded checks if NVIDIA kernel module is already loaded in kernel memory.
+func isNvidiaModuleLoaded() bool {
+	// Check if the nvidia module is loaded by checking if nvidiaModuleRefcntPath exists
+	if _, err := os.Stat(nvidiaModuleRefcntPath); err == nil {
+		refcntData, err := os.ReadFile(nvidiaModuleRefcntPath)
+		if err == nil {
+			refcnt := strings.TrimSpace(string(refcntData))
+			log.Infof("NVIDIA kernel module already loaded in kernel memory (refcnt=%s)", refcnt)
+			return true
+		}
+	}
+	return false
+}
+
 // createDevCharSymlinks creates symlinks in /host-dev-char that point to all possible NVIDIA devices nodes.
 func createDevCharSymlinks(driverInfo driverInfo, disableDevCharSymlinkCreation bool) error {
 	if disableDevCharSymlinkCreation {
@@ -926,8 +942,16 @@ func createDevCharSymlinks(driverInfo driverInfo, disableDevCharSymlinkCreation 
 
 	log.Info("creating symlinks under /dev/char that correspond to NVIDIA character devices")
 
-	// Only attempt to load NVIDIA kernel modules when we can chroot into driverRoot
-	loadKernelModules := driverInfo.isHostDriver || (driverInfo.devRoot == driverInfo.driverRoot)
+	// Check if NVIDIA module is already loaded in kernel memory.
+	// If it is, we don't need to run modprobe (which would fail if modules aren't in /lib/modules/).
+	// This handles the case where the driver container performed a userspace-only install
+	// after detecting that module was already loaded from a previous boot.
+	moduleAlreadyLoaded := isNvidiaModuleLoaded()
+
+	// Only attempt to load NVIDIA kernel modules when:
+	// 1. Module is not already loaded in kernel memory, AND
+	// 2. We can chroot into driverRoot to run modprobe
+	loadKernelModules := !moduleAlreadyLoaded && (driverInfo.isHostDriver || (driverInfo.devRoot == driverInfo.driverRoot))
 
 	// driverRootCtrPath is the path of the driver install dir in the container. This will either be
 	// driverInstallDirCtrPathFlag or '/host'.
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1042,6 +1042,32 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 			return fmt.Errorf("ERROR: failed to transform the pre-compiled Driver Daemonset: %s", err)
 		}
 	}
+
+	// Compute driver configuration digest after all transformations are complete.
+	// This digest enables fast-path driver installation by detecting when configuration
+	// hasn't changed, avoiding unnecessary driver reinstalls and pod evictions.
+	// Used by k8s-driver-manager to decide if driver cleanup is needed and by
+	// nvidia-driver container to skip full reinstall for matching configurations.
+	configDigest := utils.GetObjectHash(obj.Spec)
+
+	// Set the computed digest in driver-manager initContainer
+	driverManagerContainer := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
+	if driverManagerContainer != nil {
+		setContainerEnv(driverManagerContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+	}
+
+	// Set the computed digest in nvidia-driver container
+	driverContainer := findContainerByName(obj.Spec.Template.Spec.Containers, "nvidia-driver-ctr")
+	if driverContainer != nil {
+		setContainerEnv(driverContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+	}
+
+	// Used by dtk-build-driver to determine if fast path should be used (skip rebuild)
+	driverToolkitContainer := findContainerByName(obj.Spec.Template.Spec.Containers, "openshift-driver-toolkit-ctr")
+	if driverToolkitContainer != nil {
+		setContainerEnv(driverToolkitContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+	}
+
 	return nil
 }
 
--- a/controllers/transforms_test.go
+++ b/controllers/transforms_test.go
@@ -2067,6 +2067,23 @@ func newBoolPtr(b bool) *bool {
 	return boolPtr
 }
 
+// removeDigestFromDaemonSet removes DRIVER_CONFIG_DIGEST env var from a DaemonSet
+func removeDigestFromDaemonSet(ds *appsv1.DaemonSet) {
+	removeDigestFromContainers := func(containers []corev1.Container) {
+		for i := range containers {
+			var filtered []corev1.EnvVar
+			for _, env := range containers[i].Env {
+				if env.Name != "DRIVER_CONFIG_DIGEST" {
+					filtered = append(filtered, env)
+				}
+			}
+			containers[i].Env = filtered
+		}
+	}
+	removeDigestFromContainers(ds.Spec.Template.Spec.Containers)
+	removeDigestFromContainers(ds.Spec.Template.Spec.InitContainers)
+}
+
 func TestTransformDriverManagerInitContainer(t *testing.T) {
 	testCases := []struct {
 		description string
@@ -2783,6 +2800,9 @@ func TestTransformDriver(t *testing.T) {
 				return
 			}
 			require.NoError(t, err)
+
+			// Remove dynamically generated digest before comparison
+			removeDigestFromDaemonSet(tc.ds.DaemonSet)
 			require.EqualValues(t, tc.expectedDs, tc.ds)
 		})
 	}
@@ -3149,6 +3169,9 @@ func TestTransformDriverWithLicensingConfig(t *testing.T) {
 				return
 			}
 			require.NoError(t, err)
+
+			// Remove dynamically generated digest before comparison
+			removeDigestFromDaemonSet(tc.ds.DaemonSet)
 			require.EqualValues(t, tc.expectedDs, tc.ds)
 		})
 	}
@@ -3268,6 +3291,9 @@ func TestTransformDriverWithResources(t *testing.T) {
 				return
 			}
 			require.NoError(t, err)
+
+			// Remove dynamically generated digest before comparison
+			removeDigestFromDaemonSet(tc.ds.DaemonSet)
 			require.EqualValues(t, tc.expectedDs, tc.ds)
 		})
 	}
@@ -3348,6 +3374,9 @@ func TestTransformDriverRDMA(t *testing.T) {
 		ClusterPolicyController{client: mockClient, runtime: gpuv1.Containerd,
 			operatorNamespace: "test-ns", logger: ctrl.Log.WithName("test")})
 	require.NoError(t, err)
+
+	// Remove dynamically generated digest before comparison
+	removeDigestFromDaemonSet(ds.DaemonSet)
 	require.EqualValues(t, expectedDs, ds)
 }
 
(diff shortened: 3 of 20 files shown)


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -499,6 +499,14 @@ type DriverSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:select:auto,urn:alm:descriptor:com.tectonic.ui:select:open,urn:alm:descriptor:com.tectonic.ui:select:proprietary"
 	KernelModuleType string `json:"kernelModuleType,omitempty"`
 
+	// DriverType defines the type of NVIDIA driver to be deployed.
+	// +kubebuilder:validation:Enum=gpu;vgpu;vgpu-host-manager
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Driver Type"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.description="Driver Type"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:select:gpu,urn:alm:descriptor:com.tectonic.ui:select:vgpu,urn:alm:descriptor:com.tectonic.ui:select:vgpu-host-manager"
+	DriverType string `json:"driverType,omitempty"`
```
cdesiniotis: Let's revert this change as it should not be required. `DriverType` is only relevant in the NVIDIADriver CR where we use a single config struct to represent multiple types of drivers. For ClusterPolicy, we leverage the `driver` section for both datacenter & vGPU guest drivers, and the `vgpuManager` section for the vGPU host driver.

controllers/object_controls.go
```diff
@@ -2057,6 +2077,13 @@ func TransformVFIOManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec
 		}
 	}
 
+	// Compute configuration digest after all transformations are complete
+	digest := utils.GetObjectHash(obj.Spec)
+
+	// Set the computed digest in driver-manager initContainer
+	driverManagerContainer := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
+	setContainerEnv(driverManagerContainer, "DRIVER_CONFIG_DIGEST", digest)
+
```
cdesiniotis: I am questioning whether we really need this logic here. For the VFIOManager, the fast-track install path is not a relevant feature. What if we omitted this code and made sure the `k8s-driver-manager` code handled the case where `DRIVER_CONFIG_DIGEST` was not defined?

controllers/object_controls.go
```diff
@@ -1067,6 +1081,12 @@ func TransformVGPUManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec
 		return fmt.Errorf("failed to transform the Driver Toolkit container: %s", err)
 	}
 
+	// Compute configuration digest after all transformations are complete
+	digest := utils.GetObjectHash(obj.Spec)
+
+	driverManagerContainer := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
+	setContainerEnv(driverManagerContainer, "DRIVER_CONFIG_DIGEST", digest)
+
```
cdesiniotis: The same comment I left below in `TransformVFIOManager` is relevant here as well:

> I am questioning whether we really need this logic here. For the VFIOManager, the fast-track install path is not a relevant feature. What if we omitted this code and made sure the k8s-driver-manager code handled the case where DRIVER_CONFIG_DIGEST was not defined?

controllers/transforms_test.go
```diff
@@ -1861,6 +1863,10 @@ func TestTransformVFIOManager(t *testing.T) {
 		t.Run(tc.description, func(t *testing.T) {
 			err := TransformVFIOManager(tc.daemonset.DaemonSet, tc.clusterPolicySpec, ClusterPolicyController{logger: ctrl.Log.WithName("test")})
 			require.NoError(t, err)
+
+			actualDigest := getDigestFromDaemonSet(tc.daemonset.DaemonSet)
+			updateExpectedDigest(&tc.expectedDaemonset, actualDigest)
```
cdesiniotis: To simplify this, what if we just removed the `DRIVER_CONFIG_DIGEST` envvar from the rendered daemonset before comparing it to the expected daemonset?

manifests/state-driver/0500_daemonset.yaml
```diff
@@ -226,6 +226,10 @@ spec:
           value: "true"
         {{- end }}
       {{- end }}
+      {{- if .Driver.Spec.DriverType }}
+        - name: DRIVER_TYPE
+          value: {{ .Driver.Spec.DriverType | quote }}
+      {{- end }}
```
cdesiniotis: This is not needed anymore now that we introduced `DRIVER_CONFIG_DIGEST`, right?

internal/state/testdata/golden/driver-rdma.yaml
```diff
@@ -296,7 +300,6 @@ spec:
         - name: DRAIN_USE_FORCE
           value: "false"
         - name: DRAIN_POD_SELECTOR_LABEL
-          value: ""
```
cdesiniotis, line 343: Was this changed intended? (quite a few of the test files have this change)

internal/state/driver_test.go
```diff
@@ -80,13 +119,12 @@ func TestDriverRenderMinimal(t *testing.T) {
 	require.Nil(t, err)
 	require.NotEmpty(t, objs)
 
-	actual, err := getYAMLString(objs)
-	require.Nil(t, err)
+	setDigestInTestObjects(t, stateDriver, objs)
```
cdesiniotis: Question -- similar to my comment above in `transforms_test.go` would it be simpler to just remove the `DRIVER_CONFIG_DIGEST` from the rendered daemonset?

controllers/object_controls.go
```diff
@@ -1044,6 +1046,18 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 			return fmt.Errorf("ERROR: failed to transform the pre-compiled Driver Daemonset: %s", err)
 		}
 	}
+
+	// Compute driver configuration digest after all transformations are complete
```
cdesiniotis: Once this PR is reviewed more, let's update this comment to be a bit more descriptive on what this digest is used for. Let's add a similar comment in the NVIDIADriver code that adds this digest.

cmd/nvidia-validator/main.go
```diff
@@ -846,6 +846,20 @@ func (d *Driver) createStatusFile(driverInfo driverInfo) error {
 	return createStatusFileWithContent(outputDirFlag+"/"+driverStatusFile, statusFileContent)
 }
 
+// areNvidiaModulesLoaded checks if NVIDIA kernel modules are already loaded in kernel memory.
+func areNvidiaModulesLoaded() bool {
+	// Check if the nvidia module is loaded by checking if /sys/module/nvidia/refcnt exists
+	if _, err := os.Stat("/sys/module/nvidia/refcnt"); err == nil {
```
tariq1890: let's factor this out into a string constant

cmd/nvidia-validator/main.go
```diff
@@ -846,6 +846,20 @@ func (d *Driver) createStatusFile(driverInfo driverInfo) error {
 	return createStatusFileWithContent(outputDirFlag+"/"+driverStatusFile, statusFileContent)
 }
 
+// areNvidiaModulesLoaded checks if NVIDIA kernel modules are already loaded in kernel memory.
+func areNvidiaModulesLoaded() bool {
```
tariq1890: nit: We are just checking if a single module is loaded, so let's use singular here instead

internal/state/testdata/golden/driver-rdma.yaml
karthikvetrivel: I have a change to `driver_test.go` that cleans up artifacts in the golden files (introduced from going from unstructured object --> DaemonSet). However, this is unintended and I'll remove.

internal/state/driver_test.go
```diff
@@ -49,6 +51,29 @@ func getYAMLString(objs []*unstructured.Unstructured) (string, error) {
 		scheme.Scheme, json.SerializerOptions{Yaml: true, Pretty: false, Strict: false})
 	var sb strings.Builder
 	for _, obj := range objs {
+		// Clean up artifacts from runtime.DefaultUnstructuredConverter that shouldn't be in golden files:
+		// 1. status: zero-valued status fields are auto-added during struct conversion
+		// 2. resources: {}: empty ResourceRequirements structs are created even when omitted
+		delete(obj.Object, "status")
+
+		if spec, ok := obj.Object["spec"].(map[string]interface{}); ok {
```
tariq1890: Is there a setting in the UnstructuredConverter to prevent the auto-add?

Also, why can't cast the Object into a Daemonset and then modify the object?

internal/state/driver_test.go
```diff
@@ -60,6 +85,20 @@ func getYAMLString(objs []*unstructured.Unstructured) (string, error) {
...
 }
 
+func setDigestInTestObjects(t *testing.T, stateDriver *stateDriver, objs []*unstructured.Unstructured) {
+	ctx := context.Background()
+	logger := log.Log.WithName("test")
+	err := stateDriver.setDigestInDaemonSet(ctx, objs, logger)
+	require.Nil(t, err)
+}
+
+func compareOrUpdateGolden(t *testing.T, testName string, actual string) {
```
tariq1890: I don't understand why this is named as `compareOrUpdate`; I don't see any update operation being performed here

internal/state/driver_test.go
karthikvetrivel: From what I saw, there's no setting in UnstructuredConverter to prevent auto-adding these fields. When converting typed objects to unstructured, Go's zero values are included by default. We could cast to a DaemonSet struct and modify it, but that would require importing all the specific types and handling each kind separately. Open to your thoughts.

api/nvidia/v1/clusterpolicy_types.go
karthikvetrivel: Updated.

controllers/object_controls.go
karthikvetrivel: Thanks for calling this out. Updated.

controllers/object_controls.go
karthikvetrivel: Updated.

internal/state/driver_test.go
karthikvetrivel: Updated.

controllers/object_controls.go
karthikvetrivel: I added an initial comment here, feel free to take a look.

cmd/nvidia-validator/main.go
karthikvetrivel: Updated.

cmd/nvidia-validator/main.go
karthikvetrivel: Updated.

internal/state/driver_test.go
karthikvetrivel: Updated.

controllers/transforms_test.go
karthikvetrivel: Added in this instance!

manifests/state-driver/0500_daemonset.yaml
karthikvetrivel: Removed.

internal/state/driver_test.go
cdesiniotis: @karthikvetrivel is this change still needed? From the latest diff, I am not following why a change here is required (the golden files have not been updated).

internal/state/driver.go
```diff
@@ -24,6 +24,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/go-logr/logr"
```
cdesiniotis: Let's not introduce a new logging package. We already are using `sigs.k8s.io/controller-runtime/pkg/log` can we continue to use that?

internal/state/driver.go
```diff
@@ -692,6 +701,68 @@ func (s *stateDriver) createConfigMapVolumeMounts(ctx context.Context, namespace
...
+	}
+	return nil
+}
+
+// setDigestInDaemonSet computes a hash of the DaemonSet configuration and sets it
+// as DRIVER_CONFIG_DIGEST env var. This enables fast-path driver installation by
+// detecting when configuration hasn't changed, avoiding unnecessary reinstalls.
+// Used by k8s-driver-manager to decide if driver cleanup is needed and by
+// nvidia-driver container to skip full reinstall for matching configurations.
+func (s *stateDriver) setDigestInDaemonSet(ctx context.Context, objs []*unstructured.Unstructured, logger logr.Logger) error {
```
cdesiniotis: I don't see `logger` being used, can we remove it?

internal/state/driver_test.go
```diff
@@ -60,6 +88,16 @@ func getYAMLString(objs []*unstructured.Unstructured) (string, error) {
 	return sb.String(), nil
 }
 
+// compareWithGolden compares the actual output with the golden file.
+// It reads the expected content from the golden file and asserts equality.
+// Despite the old name suggesting "update", this function only performs comparison.
```
cdesiniotis: Not following this comment. What is the "old name" you are referring to?

karthikvetrivel: File removed.

internal/state/driver.go
karthikvetrivel: Removed.

internal/state/driver.go
karthikvetrivel: Removed.

internal/state/driver_test.go
karthikvetrivel: File is removed!

internal/state/driver.go
```diff
@@ -257,6 +258,9 @@ func (s *stateDriver) getManifestObjects(ctx context.Context, cr *nvidiav1alpha1
 		HostRoot:      clusterPolicy.Spec.HostPaths.RootFS,
 	}
 
+	// Compute digest from base config before nodepool-specific fields are added
+	renderData.ConfigDigest = utils.GetObjectHash(renderData)
```
cdesiniotis: Is this the right place to compute the digest? At this point `renderData` is pretty minimal, most notably, `renderData.Driver` is empty. Should we not compute the digest right before rendering the daemonset template?

karthikvetrivel: Woops, sorry--this was a brain fart. Moved it right before we render manifests for the node pool.

controllers/object_controls.go
```diff
@@ -1046,6 +1046,26 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
...
+	// nvidia-driver container to skip full reinstall for matching configurations.
+	configDigest := utils.GetObjectHash(obj.Spec)
+
+	// Set the computed digest in driver-manager initContainer
+	driverManagerContainer := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
+	if driverManagerContainer != nil {
+		setContainerEnv(driverManagerContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+	}
+
+	// Set the computed digest in nvidia-driver container
```
cdesiniotis, line 1059: Question -- does the `openshift-driver-toolkit-ctr` also need this digest (when running on Openshift)?

karthikvetrivel: Yes. Updated to add here (and in NVIDIADRIVER CR). Corresponding golden tests updated.

internal/state/driver.go
```diff
@@ -306,6 +307,9 @@ func (s *stateDriver) getManifestObjects(ctx context.Context, cr *nvidiav1alpha1
 			logger.Error(err, "error rendering addition driver volume", "NodePool", nodePool.name)
 		}
 
+		// Compute digest from complete config after all nodepool-specific fields are added
+		renderData.ConfigDigest = utils.GetObjectHash(renderData)
```
tariq1890: Instead of calling the function here, we can invoke the GetObjectHash method from within the template

tariq1890: 1. You can add a new method to the template functions map [here](https://github.com/NVIDIA/gpu-operator/blob/main/internal/render/render.go#L101). 
2. After that you can invoke the function by adding a new env var in the template [here](https://github.com/NVIDIA/gpu-operator/blob/main/manifests/state-driver/0500_daemonset.yaml#L145)
3. Say the function in the template function map is called `getObjectHash`, then the new env var you add to the template would look something like:

```
          - name: DRIVER_CONFIG_DIGEST
            value: {{ getObjectHash . | quote }}
```

karthikvetrivel: Updated. Let me know if that's what you had in mind.

cdesiniotis (approved):

cdesiniotis:
🚀

PR #2027, opened 2026-01-09
refactor driverRenderData to only include fields needed for template render

tariq1890:
This PR removes fields from the `driverRenderData` struct that aren't used to render the templated YAMLs. This PR also removes unused code

### DIFF
--- a/controllers/clusterinfo/clusterinfo.go
+++ b/controllers/clusterinfo/clusterinfo.go
@@ -19,7 +19,6 @@ package clusterinfo
 import (
 	"context"
 	"fmt"
-	"maps"
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -29,7 +28,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/client-go/discovery"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
@@ -41,30 +39,18 @@ import (
 // Interface to the clusterinfo package
 type Interface interface {
 	GetContainerRuntime() (string, error)
-	GetKubernetesVersion() (string, error)
 	GetOpenshiftVersion() (string, error)
-	GetRHCOSVersions(map[string]string) ([]string, error)
 	GetOpenshiftDriverToolkitImages() map[string]string
 	GetOpenshiftProxySpec() (*configv1.ProxySpec, error)
-	GetKernelVersions(map[string]string) ([]string, error)
 }
-
-const (
-	nfdOSTreeVersionLabelKey = "feature.node.kubernetes.io/system-os_release.OSTREE_VERSION"
-	nfdKernelLabelKey        = "feature.node.kubernetes.io/kernel-version.full"
-)
-
 type clusterInfo struct {
 	ctx     context.Context
 	config  *rest.Config
 	oneshot bool
 
 	containerRuntime             string
-	kubernetesVersion            string
 	openshiftVersion             string
-	rhcosVersions                []string
 	openshiftDriverToolkitImages map[string]string
-	kernelVersions               []string
 	proxySpec                    *configv1.ProxySpec
 }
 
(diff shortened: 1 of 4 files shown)


internal/state/driver.go
```diff
@@ -70,9 +70,7 @@ type driverRuntimeSpec struct {
 	KubernetesVersion             string
```
cdesiniotis: Question -- is `KubernetesVersion` used anywhere in our templates? If not, could we remove it from this struct?

tariq1890: Yes, good catch. I've removed it

controllers/clusterinfo/clusterinfo.go
```diff
@@ -151,15 +150,6 @@ func (l *clusterInfo) GetContainerRuntime() (string, error) {
 	return getContainerRuntime(l.ctx, l.config)
 }
 
-// GetKubernetesVersion returns the k8s version detected in the cluster
-func (l *clusterInfo) GetKubernetesVersion() (string, error) {
-	if l.oneshot {
-		return l.kubernetesVersion, nil
-	}
-
-	return getKubernetesVersion(l.config)
```
cdesiniotis, line 160: Can we remove this helper function too?

internal/state/driver.go
```diff
@@ -241,11 +238,22 @@ func (s *stateDriver) getManifestObjects(ctx context.Context, cr *nvidiav1alpha1
 	}
 	clusterInfo := info.(clusterinfo.Interface)
 
-	runtimeSpec, err := getRuntimeSpec(ctx, s.client, s.namespace, clusterInfo, &cr.Spec)
+	openshiftVersion, err := clusterInfo.GetOpenshiftVersion()
```
cdesiniotis: I don't see us setting `driverRenderData.RuntimeSpec.OpenShiftVersion` after getting the `openShiftVersion`

Also, I think I would slightly prefer if we moved this code snippet (which gets the openshift version) back to `getRuntimeSpec` since there is a field for it in `RuntimeSpec`

controllers/clusterinfo/clusterinfo.go
cdesiniotis: Resolving as you have done this in a follow-up commit.

internal/state/driver.go
```diff
@@ -241,11 +238,22 @@ func (s *stateDriver) getManifestObjects(ctx context.Context, cr *nvidiav1alpha1
...
+		return nil, fmt.Errorf("failed to get openshift version: %v", err)
+	}
+	openshift := openshiftVersion != ""
+
+	runtimeSpec, err := getRuntimeSpec(openshiftVersion, s.namespace, clusterInfo, &cr.Spec)
 	if err != nil {
 		return nil, fmt.Errorf("failed to construct cluster runtime spec: %w", err)
 	}
 
+	nodePools, err := getNodePools(ctx, s.client, cr.Spec.NodeSelector, cr.Spec.UsePrecompiledDrivers(), openshift)
```
tariq1890: @cdesiniotis This is the reason I moved the openshift version getter code snippet out of `getRuntimeSpec`. The `openshift` bool value needs the openshift version

internal/state/driver.go
tariq1890: Responded to this in another thread

internal/state/driver.go
cdesiniotis: Right, but couldn't you get the openshift version via `runtimeSpec.OpenShiftVersion`? I don't want to harp on this too much, just thought we could simplify the call-site a little bit.

internal/state/driver.go
cdesiniotis: `.Runtime.OpenShiftVersion` is used when rendering the daemonset template here: https://github.com/NVIDIA/gpu-operator/blob/90e757a5699e917bf9aaac0b29e1b46affa5eb30/manifests/state-driver/0500_daemonset.yaml#L231

Just making sure this field is getting set appropriately (and is not empty).

internal/state/driver.go
tariq1890: ahh yes, you're right. Thanks for pointing that out :)

internal/state/driver.go
cdesiniotis: Okay disregard my comment, I must have been looking at a stale version of this PR. I see us passing `openShiftVersion` to `getRuntimeSpec()`

internal/state/driver.go
tariq1890: Please check now

internal/state/types.go
```diff
@@ -29,13 +29,12 @@ type SyncingSource source.SyncingSource
 // driverSpec is a wrapper of NVIDIADriverSpec with an additional ImagePath field
 // which is to be populated with the fully-qualified image path.
 type driverSpec struct {
-	Spec              *nvidiav1alpha1.NVIDIADriverSpec
-	AppName           string
-	Name              string
-	ImagePath         string
-	ManagerImagePath  string
-	OCPToolkitEnabled bool
```
cdesiniotis, line 37: Quick question -- I am assuming the `OCPToolkitEnabled` field was not referenced anywhere which is why it was removed?

tariq1890: Correct

internal/state/driver.go
cdesiniotis: Looks good now. Thanks!

PR #2032, opened 2026-01-12
add nfd rules for additional nvidia drivers on nodes

rahulait:
## Description

This PR adds additional nodefeaturerules which will label the node about the additional nvidia drivers installed on the node.
With nvidiaDriver CR, we would like to have the ability to target workloads to certain nodes which has the additional driver installed. This PR adds nodefeaturerules which will add labels to nodes based on the drivers status.
## Checklist

## Testing

Deployed a cluster with the following nfd rules and having gdrcopy and gds true. Here is the output of node labels after install:
```
k get node 2u2g-gen-0409 -o json | jq .metadata.labels | grep "gpu.driver."
  "nvidia.com/gpu-driver-upgrade-state": "upgrade-done",
  "nvidia.com/gdrcopy.capable": "true",
  "nvidia.com/gds.capable": "true",
```

### DIFF
--- a/deployments/gpu-operator/templates/nodefeaturerules.yaml
+++ b/deployments/gpu-operator/templates/nodefeaturerules.yaml
@@ -103,5 +103,36 @@ spec:
             matchExpressions:
               nvidia.com/gpu.family: {op: In, value: ["hopper"]}
               tdx.enabled: {op: IsTrue}
+---
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: nvidia-kernel-modules
+spec:
+  rules:
+    - name: kernel-module-gdrdrv
+      labels:
+        nvidia.com/gdrcopy.capable: "true"
+      matchFeatures:
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            gdrdrv:
+              op: Exists
+    - name: kernel-module-nvidia_fs
+      labels:
+        nvidia.com/gds.capable: "true"
+      matchFeatures:
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            nvidia_fs:
+              op: Exists
+    - name: kernel-module-nvidia_peermem
+      labels:
+        nvidia.com/peermem.capable: "true"
+      matchFeatures:
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            nvidia_peermem:
+              op: Exists
 {{- end }}
 


deployments/gpu-operator/templates/nodefeaturerules.yaml
```diff
@@ -103,5 +103,36 @@ spec:
...
+      labels:
+        nvidia.com/gpu.driver.gds: "true"
+      matchFeatures:
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            nvidia_fs:
+              op: Exists
+    - name: kernel-module-nvidia_peermem
+      labels:
+        nvidia.com/gpu.driver.rdma: "true"
```
tariq1890:
```suggestion
        nvidia.com/peermem-enabled: "true"
```

deployments/gpu-operator/templates/nodefeaturerules.yaml
```diff
@@ -103,5 +103,36 @@ spec:
...
+      labels:
+        nvidia.com/gpu.driver.gdrcopy: "true"
+      matchFeatures:
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            gdrdrv:
+              op: Exists
+    - name: kernel-module-nvidia_fs
+      labels:
+        nvidia.com/gpu.driver.gds: "true"
```
tariq1890:
```suggestion
        nvidia.com/gds.capable: "true"
```

deployments/gpu-operator/templates/nodefeaturerules.yaml
```diff
@@ -103,5 +103,36 @@ spec:
...
+---
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: nvidia-kernel-modules
+spec:
+  rules:
+    - name: kernel-module-gdrdrv
+      labels:
+        nvidia.com/gpu.driver.gdrcopy: "true"
```
tariq1890: Given that we currently use labels such as `nvidia.com/mig.capable`, `nvidia.com/cc.capable`, `nvidia.com/mps.capable` etc. How about the label below?
```suggestion
        nvidia.com/gdrcopy.capable: "true"
```

deployments/gpu-operator/templates/nodefeaturerules.yaml
tariq1890: To be specific, when we enable RDMA in our driver container, we are essentially enabling the `nvidia-peermem` module. It is also possible to get RDMA without this module, so it may be useful to clarify what exactly is enabled

rahulait: I was trying to keep it consistent with our values.yaml fields as users know about them. I am not sure if they know it uses peermem under the hood and hence went with rdma instead of mofed or peermem.

```
driver:
  rdma:
    enabled: true
    useHostMofed: false
```

Do you think its better to use `nvidia.com/peermem.capable` or I should use `nvidia.com/rdma.capable`?

tariq1890: Yes, it is better to be specific here as we are only really enabling `nvidia_peermem` which is just one way of getting an RDMA capability

rahulait: Ack, updated.

tariq1890:
Thanks @rahulait . @cdesiniotis  Requesting your review on this

cdesiniotis (approved):

PR #2041, opened 2026-01-13
Add support for NO_RUNTIME_CONFIG to disable runtime config in the nvidia-container-toolkit

elezar:
## Description

If a user sets NO_RUNTIME_CONFIG=true in the toolkit.Env this signals that no runtime shouls be configured for use with the NVIDIA Container Toolkit. When this is the case, we disable the injection of runtime config files and sockets.

In conjunction with https://github.com/NVIDIA/nvidia-container-toolkit/pull/1521 this allows users to use the GPU Operator to install the toolkit, but to not update the runtime config.

## Checklist

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1366,14 +1366,29 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n 
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rahulait:
@elezar is this still relevant? If yes, can you please rebase it so that new CI can run?

PR #2044, opened 2026-01-14
add support for RockyLinux driver containers

tariq1890:


internal/state/nodepool.go
```diff
@@ -132,5 +133,9 @@ func getNodePools(ctx context.Context, k8sClient client.Client, selector map[str
 }
 
 func (n nodePool) getOS() string {
+	if n.osRelease == "rocky" {
+		// If the OS is RockyLinux, we will omit the RockyLinux minor version when constructing the os image tag
+		n.osVersion = strings.Split(n.osRelease, ".")[0]
```
cdesiniotis: Should this be

```suggestion
		n.osVersion = strings.Split(n.osVersion, ".")[0]
```
?

tariq1890: Yes, great catch! Thanks

PR #2045, opened 2026-01-15
Add fabric manager configuration support

mresvanis:
## Description

This PR enables [Fabric Manager (FM) configuration](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#virtualization-models) for `vm-passthrough` workloads using the [Shared NVSwitch virtualization model](https://docs.nvidia.com/datacenter/tesla/fabric-manager-user-guide/index.html#shared-nvswitch-virtualization-model).

It enables users to configure the Fabric Manager mode (i.e. `FABRIC_MODE=[0,1,2]`, `0` - full-passthrough, `1` - shared NVSwitch, `2` - vGPU) through the `ClusterPolicy` CRD, providing better support for NVIDIA multi-GPU systems in virtualized environments.

In the FM shared NVSwitch virtualization model the NVIDIA driver on the host is used for the NVSwitch devices, while the GPU devices are bound to the `vfio-pci` driver. The goal is for the GPU devices to be passed-through to kubevirt VMs, while the respective fabric is managed on the host. 

Depends on / relates to: https://github.com/NVIDIA/gpu-driver-container/pull/538

#### ClusetrPolicy API
- add `FabricManagerSpec` to the `ClusterPolicy` CRD with support for two modes:
  - `full-passthrough` (`FABRIC_MODE=0`) - default mode.
  - `shared-nvswitch` (`FABRIC_MODE=1`) - shared NVSwitch virtualization mode.
- update all CRD manifests across bundle, config, and deployment directories to include the new Fabric Manager configuration fields.

#### Controller logic
- enable driver installation when using `vm-passthrough` with FM shared NVSwitch mode and pass an env var to the driver container to indicate the selected fabric mode (the driver container is the one configuring and starting the FM).
- integrate FM configuration checks into the state manager workflow.

#### Driver state management
- add logic to detect and handle Fabric Manager shared NVSwitch mode.
- update driver startup probe behavior for `vm-passthrough` and FM shared NVSwitch mode case.
- adjust the driver startup probe to accommodate Fabric Manager requirements in `vm-passthrough` with shared NVSwitch mode.

#### Sandbox validator
- add driver validation as the first init container when FM shared NVSwitch mode.
- add wait flow to the vfio-pci validation.

#### VFIO manager
- wait for driver to be ready when FM shared NVSwitch mode - this step is required because we need a mapping of GPU physical module ID to the respective PCIe address, as FM identifies GPUs by their physical module ID. The latter can be found via `nvidia-smi`, which requires the driver to be loaded and bound to the GPU devices. Once that's done we can bind the GPU devices to vfio-pci.
- replace init container with `vfio-manage unbind --all` when FM shared NVSwitch mode.

## Checklist

## Testing

TBD


LandonTClipp:
How coincidental that I resolved to implement something like this and 2 hours ago you submitted this draft!

I want to ask what the plan is for the CDI-side. The ideal scenario is that the fabricmanager can be spawned as a Kata container, which means we need to inject the NVSwitch VFIO cdevs just like how we do for passthrough GPUs. When I tried to use GPU operator a few months ago, this was simply not possible at the time so I used libvirt instead. Does the GPU Operator CDI already expose the NVswitches to k8s now? I apologize if my knowledge is a little out of date.

assets/state-driver/0400_configmap.yaml
```diff
@@ -22,8 +22,14 @@ data:
     fi
 
     if ! nvidia-smi; then
-      echo "nvidia-smi failed"
-      exit 1
+      # For vm-passthrough with shared-nvswitch mode, nvidia-smi may fail due to unbound devices
+      # Fall back to checking if nvidia module is loaded when FABRIC_MANAGER_FABRIC_MODE=1
```
cdesiniotis, line 26: Question (for my understanding) -- GPUs may not be bound to the `nvidia` driver since there is a chance that the `vfio-manager` ran already and unbound the devices? Am I understanding this correct?

assets/state-driver/0400_configmap.yaml
```diff
@@ -22,8 +22,14 @@ data:
...
+      if [ "${FABRIC_MANAGER_FABRIC_MODE:-}" = "1" ] && lsmod | grep -q "^nvidia "; then
```
cdesiniotis: Question -- isn't the right-hand-side of this if statement redundant? Don't we already know the nvidia module is loaded prior to this (see L19-22)?

assets/state-driver/0500_daemonset.yaml
cdesiniotis, line 1: This file is hard for me to review since the diff is large and it appears to mostly be a change in indentation. Did anything meaningful change here besides the indentation? (if not, I'd prefer if we reverted this to minimize the diff)

controllers/object_controls.go
```diff
@@ -2091,10 +2012,70 @@ func TransformKataManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec
...
 func TransformVFIOManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) error {
-	// update k8s-driver-manager initContainer
-	err := transformDriverManagerInitContainer(obj, &config.VFIOManager.DriverManager, nil)
-	if err != nil {
-		return fmt.Errorf("failed to transform k8s-driver-manager initContainer for VFIO Manager: %v", err)
+	// Check if we're in shared-nvswitch mode
+	if config.FabricManager.IsSharedNVSwitchMode() {
+		// In shared-nvswitch mode, replace driver uninstall with device unbind
+		// Find the k8s-driver-manager init container and replace it with vfio-manage unbind
+		for i := range obj.Spec.Template.Spec.InitContainers {
```
cdesiniotis: We have a helper `findContainerByName()` that we should use here:
```
container := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
container.Name = "vfio-device-unbind"
// ... all other transformations ...
```

controllers/object_controls.go
```diff
@@ -2091,10 +2012,70 @@ func TransformKataManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec
...
+  echo "waiting for the driver validations to be ready..."
+  sleep 5
+done
+
+set -o allexport
+cat /run/nvidia/validations/driver-ready
+. /run/nvidia/validations/driver-ready
+
+echo "Driver is ready, proceeding with device unbind"
+exec vfio-manage unbind --all`}
```
cdesiniotis: Instead of adding this in code, what about encapsulating this logic in a custom entrypoint script that is stored in a ConfigMap? The fabric manager mode can be indicated via an envvar and the entrypoint script can check the envvar to determine what actions to take (and what command to run -- `k8s-driver-manager uninstall_driver` vs `vfio-manage unbind --all`).

cdesiniotis:
Question -- do we need to add an extra host path volume to the `sandbox-device-plugin` so that it can access the fabric manager UNIX socket?

assets/state-driver/0500_daemonset.yaml
mresvanis: Thank you for pointing this out! I apologize for that unnecessary diff, nothing changed in the driver DaemonSet (it was just my inconsiderate vim configuration, I'll take care of this for the rest of the diffs). I have reverted this.

assets/state-driver/0400_configmap.yaml
mresvanis: You're absolutely right, I have removed this unnecessary condition.

controllers/object_controls.go
mresvanis: Thank you for pointing this out, I completely missed that helper! I've updated this change to use the helper.

mresvanis:
@cdesiniotis thank you very much for your review!
 
> Question -- do we need to add an extra host path volume to the `sandbox-device-plugin` so that it can access the fabric manager UNIX socket?

That's correct, I hadn't pushed all commits by the time you reviewed this change. It should be there now: 116e9bf29515a886432e6c72be4ed277299adf38

controllers/object_controls.go
mresvanis: Great suggestion! It didn't feel right putting this script inline here in the first place. I have updated this part accordingly. PTAL

assets/state-driver/0400_configmap.yaml
mresvanis: Yes, that's correct. In `shared-nvswitch` mode, the `vfio-manager` runs `vfio-manage unbind --all` to unbind GPUs from the nvidia driver and rebind them to `vfio-pci` for VM passthrough. The nvidia kernel module remains loaded (needed for Fabric Manager / NVSwitch management), but since the GPU devices are no longer bound to the nvidia driver `nvidia-smi` fails. So we fall back to just verifying the kernel module is loaded, which is sufficient for this mode.

cdesiniotis:
@mresvanis thanks for the work on this. A decision has been made to enable FM-based partitioning using DRA (not the device-plugin) so most of the changes in this PR are not needed anymore. Closing this PR. 

https://github.com/NVIDIA/gpu-driver-container/pull/538 is still needed however, I will take another look at that one soon.

PR #2054, opened 2026-01-20
operand: Updates to cc-manager and sandbox-device-plugin

zvonkok:
sandbox-device-plugin: Add /var/run/cdi mount since the DP is now handling CDI spec creation
cc-manager: Remove obsolete CC_CAPABLE_DEVICE_IDS

### DIFF
--- a/assets/state-sandbox-device-plugin/0500_daemonset.yaml
+++ b/assets/state-sandbox-device-plugin/0500_daemonset.yaml
@@ -69,6 +69,8 @@ spec:
               mountPath: /var/lib/kubelet/device-plugins
             - name: vfio
               mountPath: /dev/vfio
+            - name: cdi-root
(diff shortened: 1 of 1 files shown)


assets/state-sandbox-device-plugin/0500_daemonset.yaml
```diff
@@ -80,3 +82,7 @@ spec:
         - name: vfio
           hostPath:
             path: /dev/vfio
+        - name: cdi-root
+          hostPath:
+            path: /var/run/cdi
+            type: DirectoryOrCreate
```
cdesiniotis: nit: let's add the newline back

zvonkok: @cdesiniotis Fixed

PR #2059, opened 2026-01-22
[state-toolkit] expose the namespace as an envar for the main toolkit container

tariq1890:
This change allows the namespace to be retrieved by the toolkit container at runtime when instantiating the NRI plugin server

### DIFF
--- a/assets/state-container-toolkit/0500_daemonset.yaml
+++ b/assets/state-container-toolkit/0500_daemonset.yaml
@@ -71,6 +71,10 @@ spec:
           value: "void"
         - name: TOOLKIT_PID_FILE
           value: "/run/nvidia/toolkit/toolkit.pid"
+        - name: NRI_NAMESPACE
+          valueFrom:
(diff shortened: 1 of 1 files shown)


assets/state-container-toolkit/0500_daemonset.yaml
```diff
@@ -71,6 +71,10 @@ spec:
           value: "void"
         - name: TOOLKIT_PID_FILE
           value: "/run/nvidia/toolkit/toolkit.pid"
+        - name: KUBERNETES_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
elezar: Using `NRI_NAMESPACE` here will make the intent clearer.

tariq1890:
Thanks for the review @karthikvetrivel ! This PR should be good to merge now that https://github.com/NVIDIA/nvidia-container-toolkit/pull/1601 has been merged

PR #2062, opened 2026-01-23
Fix state reconciliation early return on unchanged DaemonSet

Mirza-Samad-Ahmed-Baig:
## Description

Fixed a critical reconciliation bug where stateSkel.createOrUpdateObjs would return nil (exiting early) when a DaemonSet’s hash was unchanged, potentially skipping reconciliation of later objects; now it continues.

## Checklist
- [x ] No secrets, sensitive information, or unrelated changes
- [ x] Test cases are added for new code paths

## Testing
Added a regression unit test that fails on the old behavior and passes with the fix .

### DIFF
--- a/internal/state/state_skel_test.go
+++ b/internal/state/state_skel_test.go
@@ -1,0 +1,72 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package state
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


internal/state/state_skel_test.go
```diff
@@ -0,0 +1,72 @@
...
+		consts.NvidiaAnnotationHashKey: dsHash,
+	})
+
+	k8sClient := fake.NewClientBuilder().WithScheme(scheme.Scheme).WithRuntimeObjects(existingDs).Build()
+	stateSkel.client = k8sClient
+
+	err := stateSkel.createOrUpdateObjs(ctx, func(_ *unstructured.Unstructured) error { return nil }, []*unstructured.Unstructured{ds, cm})
+	require.NoError(t, err)
+
+	gotCm := &unstructured.Unstructured{}
```
rajathagasthya, line 68: Can we also simulate a case where a ConfigMap exists and an update happens to it successfully (while the DaemonSet remains unchanged)?

rahulait:
This is kind of ok as its not really a bug. When nvidiaDriver is changed, hash for every daemonset owned by it changes. There isn't any case as of today where nvidiadriver manages multiple daemonsets and hash of only one daemonset changes. So returning early is kind of ok. But yeah, if others think we can just evaluate all than returning early, then that is fine as well.

rajatchopra:
> This is kind of ok as its not really a bug. When nvidiaDriver is changed, hash for every daemonset owned by it changes. There isn't any case as of today where nvidiadriver manages multiple daemonsets and hash of only one daemonset changes. So returning early is kind of ok. But yeah, if others think we can just evaluate all than returning early, then that is fine as well.

How about the case that @rajathagasthya suggested. Configmap changes but daemonset does not.
However unlikely that is.

rajathagasthya:
Since this PR hasn't been updated and it doesn't affect functionality, I'm going to close this to keep the PR backlog current. Please re-open if necessary.

PR #2066, opened 2026-01-24
[gpu-feature-discovery] disable nodefeature API for openshift clusters

tariq1890:
This change is needed to unblock GFD in OpenShift clusters. 

With the recent change in GFD where the NodeFeature API was [enabled](https://github.com/NVIDIA/k8s-device-plugin/pull/1504/files) by default, we observed that the latest GFD images were not writing the gpu feature labels to the node on OpenShift clusters. After further analysis, we found that the NFD that comes in OpenShift clusters still leverage gRPC as opposed to the `NodeFeature` Custom Resource


cdesiniotis:
cc @rajathagasthya

PR #2075, opened 2026-01-27
fix: eliminate hardcoded command names(move to distroless)

jojimt:
## Description

Some of the operands viz. cc-manager and sandbox device plugin have their commands hardcoded. This causes problems, e.g. when we move to python or make a command name change for any reason. Instead, we want to let the Dockerfile specify the entrypoint. This is needed for moving to distroless. Also, updated the preStop hook not to use shell, consistent with distroless.

## Checklist

## Testing

Manual testing on a cluster. Verified preStop lifecycle hook as well.

### DIFF
--- a/assets/state-cc-manager/0500_daemonset.yaml
+++ b/assets/state-cc-manager/0500_daemonset.yaml
@@ -27,7 +27,6 @@ spec:
         - name: nvidia-cc-manager
           image: "FILLED BY THE OPERATOR"
           imagePullPolicy: IfNotPresent
-          command: ["k8s-cc-manager"]
           env:
           - name: NODE_NAME
             valueFrom:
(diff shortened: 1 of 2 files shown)


jojimt:
@rajatchopra @cdesiniotis @zvonkok fyi

rajatchopra:
LGTM. Thanks Joji.

assets/state-cc-manager/0500_daemonset.yaml
```diff
@@ -54,7 +55,7 @@ spec:
           lifecycle:
             preStop:
               exec:
-                command: ["/bin/sh", "-c", "rm -f /run/nvidia/validations/.cc-manager-ctr-ready"]
+                command: ["/bin/rm", "-f", "/run/nvidia/validations/.cc-manager-ctr-ready"]
```
elezar, line 58: Was this change intended to be part of this PR? At the very least I would expect this to be a separate commit since it is not related to removing the hard-coded command names.

jojimt: Yes, it is intended to be a part of this PR. The goal is to move to distroless which does not include the shell. This is also why we need the hardcoded names to be removed. I have updated the description and heading to mention this.

cdesiniotis (approved):

PR #2078, opened 2026-01-28
new privileges for sandbox-device-plugin

rajatchopra:
## Description

sandbox-device-plugin will launch the GFD job within the sandbox env, so we need new privileges for it to do so
Changes:
- add CRUD privileges for jobs to Role
- add NODE_NAME/POD_NAMESPACE info so that the device plugin can launch job on its own node

GFD launch PR in sandbox-device-plugin: https://github.com/NVIDIA/sandbox-device-plugin/pull/30

## Checklist

## Testing

Manual testing on cluster

### DIFF
--- a/assets/state-sandbox-device-plugin/0200_role.yaml
+++ b/assets/state-sandbox-device-plugin/0200_role.yaml
@@ -12,3 +12,22 @@ rules:
   - use
   resourceNames:
   - privileged
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - delete
+- apiGroups:
+  - nfd.k8s-sigs.io
+  resources:
+  - nodefeatures
+  verbs:
+  - create
+  - get
+  - list
+  - watch
--- a/assets/state-sandbox-device-plugin/0500_daemonset.yaml
+++ b/assets/state-sandbox-device-plugin/0500_daemonset.yaml
@@ -61,6 +61,17 @@ spec:
         - image: "FILLED BY THE OPERATOR"
           imagePullPolicy: IfNotPresent
           name: nvidia-sandbox-device-plugin-ctr
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: spec.nodeName
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: metadata.namespace
           securityContext:
             privileged: true
           volumeMounts:


assets/state-sandbox-device-plugin/0200_role.yaml
```diff
@@ -12,3 +12,14 @@ rules:
   - use
   resourceNames:
   - privileged
+- apiGroups:
+  - batch
+  resources:
+  - jobs
+  verbs:
+  - create
```
tariq1890, line 20: Do you need all of these permissions? I prefer starting with a minimal list. It seems that `create`, `get`, `watch` and `delete` would suffice here, yes?

rajatchopra: We can skip `update` I think. I always assumed watch requires list/get.
Fixed.

deployments/gpu-operator/values.yaml
```diff
@@ -489,12 +489,14 @@ kataManager:
...
-  version: v1.4.0
+  image: sandbox-device-plugin
+  version: v1.4.1
   imagePullPolicy: IfNotPresent
   imagePullSecrets: []
   args: []
-  env: []
+  env:
+    - name: GFD_IMAGE
+      value: "nvcr.io/nvidia/k8s-device-plugin:v0.17.0"
```
elezar: Why is this not v0.18.2? What about the other envvars for GFD?

rajatchopra: It can be. The only drawback we have is that the image has to have /bin/sh in it because we run GFD binary followed by a sleep. 18.2 does not have a /bin/sh in it.

Other env vars for GFD are planted by sandbox-device-plugin which actually launches the pod: https://github.com/NVIDIA/sandbox-device-plugin/pull/30/changes/55f204bfdfc81374fe209dda5d61f69f87b8327c#diff-1585cc3519154c5be76021e2a775a4c53dba9724b7f52abd34b6a67f3c9d1e4aR226

rajatchopra: This portion is now removed. Not needed anymore since sandbox-device-plugin carries the GFD binary itself.

assets/state-sandbox-device-plugin/0210_clusterrole.yaml
```diff
@@ -17,3 +17,13 @@ rules:
...
+- apiGroups:
+  - nfd.k8s-sigs.io
+  resources:
+  - nodefeatures
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - delete
```
cdesiniotis: This needs to move to the Role object (not ClusterRole) since NodeFeatures are namespaced objects.

rajatchopra: fixed.

assets/state-sandbox-device-plugin/0200_role.yaml
```diff
@@ -12,3 +12,23 @@ rules:
...
+- apiGroups:
+  - nfd.k8s-sigs.io
+  resources:
+  - nodefeatures
+  verbs:
+  - create
+  - get
+  - list
+  - watch
+  - delete
```
cdesiniotis: We should remove this line. 

The `delete` verb is not present in the gpu-operator role object (or GFD's role object for that matter). Also, it is not needed by the sandbox-device-plugin since GFD is running as oneshot (it creates the CR but never deletes).

```suggestion
```

rajatchopra: fixed.

PR #2086, opened 2026-01-30
Enable cc manager by default on all kata nodes

jojimt:
## Description

- Enable cc manager by default on all kata-nodes (indicated by `nvidia.com/gpu.workload.config=vm-passthrough`)
- Eliminate `nvidia.com/cc.capable` label and set `ccManager.enabled=true` to accomplish the above.
- Eliminate CC specific NFD rule as it's handled by kata-deploy now.

Related PR: https://github.com/NVIDIA/k8s-cc-manager/pull/40
Kata PR to add cc selector to gpu runtimeclasses to follow.

## Checklist

## Testing

Manual testing in cluster.


jojimt:
@cdesiniotis @tariq1890 PTAL

deployments/gpu-operator/values.yaml
```diff
@@ -503,8 +503,8 @@ sandboxDevicePlugin:
   resources: {}
 
 ccManager:
-  enabled: false
-  defaultMode: "off"
+  enabled: true
+  defaultMode: "on"
```
cdesiniotis, line 507: Question -- does this mean that by default when cc-manager runs on a CC-capable system it will enable CC mode on the GPUs?

jojimt: Yes, default is on for cc-capable host and off for non-cc host.

PR #2088, opened 2026-01-30
add support rhel10 driver containers

tariq1890:


internal/state/nodepool.go
```diff
@@ -132,10 +139,19 @@ func getNodePools(ctx context.Context, k8sClient client.Client, selector map[str
 	return nodePools, nil
 }
 
-func (n nodePool) getOS() string {
-	if n.osRelease == "rocky" {
-		// If the OS is RockyLinux, we will omit the RockyLinux minor version when constructing the os image tag
-		n.osVersion = strings.Split(n.osVersion, ".")[0]
+func getOSTag(osRelease, osVersion string) (string, error) {
```
cdesiniotis, line 143: Can we add unit tests for this method?

internal/state/nodepool.go
```diff
@@ -94,7 +96,12 @@ func getNodePools(ctx context.Context, k8sClient client.Client, selector map[str
 		nodePool.nodeSelector[nfdOSVersionIDLabelKey] = osVersion
 		nodePool.osRelease = osID
 		nodePool.osVersion = osVersion
-		nodePool.name = nodePool.getOS()
```
cdesiniotis, line 97: Do we no longer need to set `nodePool.name`?

internal/state/driver_test.go
```diff
@@ -507,6 +507,9 @@ func TestGetDriverAppName(t *testing.T) {
 		osRelease: "ubuntu",
 		osVersion: "20.04",
 	}
+	var err error
```
cdesiniotis, line 510: Should we add additional test cases for rhel and rocky?

internal/state/nodepool.go
tariq1890: Sure

internal/state/nodepool.go
tariq1890: The `nodePool.name` has been replaced with `nodePool.osTag`

internal/state/driver_test.go
tariq1890: Do you mean YAML template render tests?

internal/state/nodepool.go
tariq1890: Ok, I think you're right. This should not have been removed

tariq1890: I've added this back

internal/state/driver_test.go
cdesiniotis: This comment is addressed by the unit tests you added in `nodepool_test.go`. Resolving this comment...

cdesiniotis (approved):

PR #2089, opened 2026-02-01
do not render/delete runtimeclasses when NRI Plugin is enabled

tariq1890:


controllers/object_controls.go
```diff
@@ -5174,6 +5183,24 @@ func RuntimeClasses(n ClusterPolicyController) (gpuv1.State, error) {
 		return transformKataRuntimeClasses(n)
 	}
 
+	if n.stateNames[state] == "pre-requisites" && !n.isStateEnabled(n.stateNames[state]) {
```
cdesiniotis, line 5178: Question -- could we not simplify this to below?
```suggestion
	if !n.isStateEnabled(n.stateNames[state]) {
```
Regardless of what state we are processing, don't we want to cleanup any runtime class objects when the state is disabled?

controllers/object_controls.go
```diff
@@ -5174,6 +5183,24 @@ func RuntimeClasses(n ClusterPolicyController) (gpuv1.State, error) {
...
+			if exists, err = k8sObjectExists(n, obj.Name, obj.Kind); err != nil {
+				return gpuv1.NotReady, fmt.Errorf("error checking if RuntimeClass %s exists: %w", obj.Name, err)
+			}
+			if exists {
+				logger.Info("Found, deleting...")
+				if err = n.client.Delete(n.ctx, &obj); err != nil {
+					return gpuv1.NotReady, fmt.Errorf("error deleting RuntimeClass %s: %w", obj.Name, err)
+				}
+			}
+		}
```
cdesiniotis: Thoughts on simplifying this to the below? 
```suggestion
			err := n.client.Delete(n.ctx, &obj)
			if err != nil && !apierrors.IsNotFound(err) {
				n.logger.Info("Couldn't delete", "RuntimeClass", obj.Name, "Error", err)
				return gpuv1.NotReady, err
			}
```

tariq1890: Yeah, this is a good suggestion

tariq1890: This is done, thanks!

controllers/object_controls.go
tariq1890: I prefer to keep this level of specificity. If we add runtimeclasses in the future as part of another state, then we wouldn't accidentally delete those

cdesiniotis (approved):

PR #2097, opened 2026-02-04
[improvement] : Capture nvidiadriver CRs as well

rahulait:
Changes to must-gather script include:
1. Additional step to capture nvidiadriver CRs as well
2. Capture pod logs with timestamps. Without this, for ex for driver pods, its difficult to tell when the pod was restarted and when those log lines were written.
3. Add labels to filter driver pods in case driver daemonsets are provisioned using nvidiadriver. This will now run nvidia-bug-report script for nodes where driver pod is provisioned by nvidiadriver CR
4. Add logic to capture which processes are using GPU if any. This will help in identifying cases when we don't expect anyone to be using GPU

## Checklist

### DIFF
--- a/hack/must-gather.sh
+++ b/hack/must-gather.sh
@@ -103,6 +103,21 @@ fi
 
 echo
 echo "#"
+echo "# NVIDIADriver"
+echo "#"
+echo
+
+NVIDIA_DRIVERS=$($K get nvidiadrivers.nvidia.com -A -oname)
+
+if [[ "${NVIDIA_DRIVERS}" ]]; then
+    echo "Get NVIDIADriver resources"
+    $K get nvidiadrivers.nvidia.com -A -oyaml > "${ARTIFACT_DIR}/nvidiadrivers.yaml"
+else
+    echo "NVIDIADriver resource(s) not found in the cluster."
+fi
+
+echo
+echo "#"
 echo "# Nodes and machines"
 echo "#"
 echo
@@ -166,10 +181,12 @@ $K get "${OPERATOR_POD_NAME}" \
 echo "Get the GPU Operator Pod logs"
 $K logs "${OPERATOR_POD_NAME}" \
     -n "${OPERATOR_NAMESPACE}" \
+    --timestamps \
     > "${ARTIFACT_DIR}/gpu_operator_pod.log"
 
 $K logs "${OPERATOR_POD_NAME}" \
     -n "${OPERATOR_NAMESPACE}" \
+    --timestamps \
     --previous \
     > "${ARTIFACT_DIR}/gpu_operator_pod.previous.log"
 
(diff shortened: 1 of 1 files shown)


rajathagasthya (approved):

cdesiniotis (approved):

rahulait:
@cdesiniotis @rajathagasthya re-requesting review as I added one more commit which now adds labelselector for nvidiadriver pods so that we capture nvidia-bug-report output from nodes as well when nvidiadriver cr is used. I also added some additional logic to capture GPU usage on nodes for easier troubleshooting of scenarios where someone outside of gpu-operator took access to GPU and causes issues with driver install/upgrades.

It can be tested using:
```
curl -fsSL https://raw.githubusercontent.com/rahulait/gpu-operator/refs/heads/update-must-gather/hack/must-gather.sh | bash
```

hack/must-gather.sh
```diff
@@ -262,7 +281,17 @@ echo "# nvidia-bug-report.sh"
 echo "#"
 echo ""
 
-for pod in $($K get pods -lopenshift.driver-toolkit -oname -n "${OPERATOR_NAMESPACE}"; $K get pods -lapp=nvidia-driver-daemonset -oname -n "${OPERATOR_NAMESPACE}"; $K get pods -lapp=nvidia-vgpu-manager-daemonset -oname -n "${OPERATOR_NAMESPACE}");
+# Find driver pods using multiple label selectors to support different deployment methods
+driver_pods=""
+driver_pods="$driver_pods $($K get pods -lopenshift.driver-toolkit -oname -n "${OPERATOR_NAMESPACE}" 2>/dev/null || true)"
+driver_pods="$driver_pods $($K get pods -lapp=nvidia-driver-daemonset -oname -n "${OPERATOR_NAMESPACE}" 2>/dev/null || true)"
+driver_pods="$driver_pods $($K get pods -lapp.kubernetes.io/component=nvidia-driver -oname -n "${OPERATOR_NAMESPACE}" 2>/dev/null || true)"
```
cdesiniotis: This is not critical as we do de-duplicate the list of pods -- but should we consider consolidating this to the below since the ClusterPolicy deployed driver also has the component label in its [spec](https://github.com/NVIDIA/gpu-operator/blob/90e8bf449be087703a4878d3df436a4963248bd2/assets/state-driver/0500_daemonset.yaml#L25)?
```suggestion
driver_pods="$driver_pods $($K get pods -lapp.kubernetes.io/component=nvidia-driver -oname -n "${OPERATOR_NAMESPACE}" 2>/dev/null || true)"
```

rahulait: Yeah, makes sense. That label was added to clusterpolicy since v23.9.0 and we shouldn't have anyone using versions older than that. I'll update the code.

tariq1890:
Can you amend the commit message to "[improvement] capture NVIDIADriver CRs in must-gather" ?

hack/must-gather.sh
rahulait: Updated.

rahulait:
> Can you amend the commit message to "[improvement] capture NVIDIADriver CRs in must-gather" ?

Yup, updated.

PR #2101, opened 2026-02-05
Driver custom repo configmap fix 

### LINKED ISSUE #562: Heterogenous cluster with airgap failing to detect customrepo configmap
### 1. Quick Debug Information
* OS/Version: master node and other worker node Rocky 8.8 , Gpu worker node Rhel 8.8
* Kernel Version: 4.18.0-477.15.1.el8_8.x86_64 
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker): Containerd
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): k8s (v1.24.12)
* GPU Operator Version: 23.3.2

### 2. Issue or feature description
Kubernetes cluster with master node and non gpu worker node running on Rocky 8.8 OS and gpu worker node running on Rhel 8.8 OS in an airgapped environment.
The custom repo configmap that is injected to the driver daemon set says not supported. Following is an image from gpu-operator pod logs.
Because of this only **gpu-operator** pod and **gpu-node-feature-discovery** pods are coming up, rest of the pods like driver, container-toolkit, dcgm-exporter etc are missing ( their daemonsets are also not present)

gpu-operator-logs

### 3. Steps to reproduce the issue

1. create a kubernetes cluster with 1 master (Rocky OS 8.8) and 2 worker nodes (1 node with Rocky OS 8.8 and other gpu node with Rhel OS 8.8
2. Install gpu operator through helm with 23.3.2 version, also pass custom configmap for driver in values.yaml
```
driver: 
    repoConfig:
        configMapName: "repo-config"
```

### 4. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)

 - [ ] kubernetes all resource status: `kubectl get all -n gpu-operator`
 
Screenshot 2023-07-29 at 1 19 45 PM

### 5. Our debug analysis
1.  We found that gpu operator pod is always scheduled on the master node because of nodeAffinity
2. We edited the gpu operator deployment and provided nodeName for it to be scheduled on the gpu node. Once the gpu operator pod started running on the gpu node all the pods (driver daemonset , toolkit, dcgm exporter etc) came up 
3. If it is not supported in heterogenous, when we dont pass custom configmap ( non airgap scenario), there is no error in gpu operator pod logs and all the pods (driver, containertoolkit, dcgm-exporter etc) are up
4.  We want to understand why there is a custom configmap check added, even if it is added, we want to understand why the distribution is not supported for **heterogenous cluster** in an **airgap** environment

Shivkumar13:
## Description
Fixes #562 

Changed the getRepoConfigPath() working to use the OS version labels from GPU worker nodes (added by NFD) when determining what OS-specific paths to use for repository configuration files. As per current implementation, for the GPU stack we only check for OS version labels on GPU nodes but for injecting custom ConfigMap we are using /etc/os-release on the current operator node.

## Checklist

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -2918,7 +2918,7 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 		if config.Driver.RepoConfig != nil && config.Driver.RepoConfig.ConfigMapName != "" {
 			// note: transformDriverContainer() will have already created a Volume backed by the ConfigMap.
 			// Only add a VolumeMount for nvidia-fs-ctr.
-			destinationDir, err := getRepoConfigPath()
+			destinationDir, err := n.getRepoConfigPath()
 			if err != nil {
 				return fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %w", err)
 			}
@@ -2931,7 +2931,7 @@ func transformGDSContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 
 		// set any custom ssl key/certificate configuration provided
 		if config.Driver.CertConfig != nil && config.Driver.CertConfig.Name != "" {
-			destinationDir, err := getCertConfigPath()
+			destinationDir, err := n.getCertConfigPath()
 			if err != nil {
 				return fmt.Errorf("ERROR: failed to get destination directory for ssl key/cert config: %w", err)
 			}
@@ -3011,7 +3011,7 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 		if config.Driver.RepoConfig != nil && config.Driver.RepoConfig.ConfigMapName != "" {
 			// note: transformDriverContainer() will have already created a Volume backed by the ConfigMap.
 			// Only add a VolumeMount for nvidia-gdrcopy-ctr.
-			destinationDir, err := getRepoConfigPath()
+			destinationDir, err := n.getRepoConfigPath()
 			if err != nil {
 				return fmt.Errorf("ERROR: failed to get destination directory for custom repo config: %w", err)
 			}
@@ -3024,7 +3024,7 @@ func transformGDRCopyContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolic
 
 		// set any custom ssl key/certificate configuration provided
 		if config.Driver.CertConfig != nil && config.Driver.CertConfig.Name != "" {
-			destinationDir, err := getCertConfigPath()
+			destinationDir, err := n.getCertConfigPath()
 			if err != nil {
 				return fmt.Errorf("ERROR: failed to get destination directory for ssl key/cert config: %w", err)
 			}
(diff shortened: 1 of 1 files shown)


rahulait:
@Shivkumar13 Can you please sign your commit so that CI can be triggered?

controllers/object_controls.go
```diff
@@ -3285,18 +3285,55 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
+		if !ok || id == "" {
+			n.logger.Info("WARNING: GPU node missing NFD OS release label, is NFD installed?",
+				"NodeName", node.Name, "Label", nfdOSReleaseIDLabelKey)
+			continue
+		}
+		if osID == "" {
+			osID = id
+		} else if osID != id {
+			n.logger.Info("WARNING: GPU nodes have different OS distributions",
+				"expected", osID, "found", id, "NodeName", node.Name)
```
rahulait: We can't predict here what is the expected osID as it will always consider first detected one as expected. Can we re-use the `kernelFullVersion` function here as it does the same things we are doing in this new function `getGPUNodeOSID`.

controllers/object_controls.go
rahulait: We should also fix this. `getCertConfigPath` has the same bug as `getRepoConfigPath`

Shivkumar13: Hi Rahul, I was thinking of having another PR for that, should we do this in the same PR?

Shivkumar13: Added this change in the same PR.

controllers/object_controls.go
Shivkumar13: Done

Shivkumar13:
PTAL @tariq1890 @rahulait, Thanks!

controllers/object_controls.go
```diff
@@ -3285,32 +3285,37 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
+func (n ClusterPolicyController) getCertConfigPath() (string, error) {
+	_, osTag, _ := kernelFullVersion(n)
+	if osTag == "" {
+		return "", fmt.Errorf("unable to determine GPU node OS from NFD labels, is NFD installed?")
 	}
-
-	os := release["ID"]
-	if path, ok := CertConfigPathMap[os]; ok {
-		return path, nil
+	for osID, path := range CertConfigPathMap {
```
rahulait: Instead of iterating over each key of map, we could also do something like:
```
	// Extract base OS ID by stripping version suffix from osTag
	// Examples: "rhel10" -> "rhel", "ubuntu22.04" -> "ubuntu", "rocky9" -> "rocky"
	osID := strings.TrimRight(osTag, "0123456789.")
    if path, ok := CertConfigPathMap[osID]; ok {
        return path, nil
    }
    return "", fmt.Errorf("distribution %s not supported", osID)
```

controllers/object_controls.go
```diff
@@ -3285,32 +3285,37 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
-	if path, ok := RepoConfigPathMap[os]; ok {
-		return path, nil
+	// osTag is e.g. "rhel8.8", "ubuntu22.04", "rocky9" — match against RepoConfigPathMap keys
+	for osID, path := range RepoConfigPathMap {
+		if strings.HasPrefix(osTag, osID) {
+			return path, nil
+		}
 	}
-	return "", fmt.Errorf("distribution not supported")
+	return "", fmt.Errorf("distribution %s not supported", osTag)
```
rahulait: Making comment more specific.
```suggestion
	return "", fmt.Errorf("repository configuration not supported for distribution %s", osTag)
```

controllers/object_controls.go
```diff
@@ -3285,32 +3285,37 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
-	os := release["ID"]
-	if path, ok := CertConfigPathMap[os]; ok {
-		return path, nil
+	for osID, path := range CertConfigPathMap {
+		if strings.HasPrefix(osTag, osID) {
+			return path, nil
+		}
 	}
-	return "", fmt.Errorf("distribution not supported")
+	return "", fmt.Errorf("distribution %s not supported", osTag)
```
rahulait:
```suggestion
	return "", fmt.Errorf("certificate configuration not supported for distribution %s", osTag)
```

Shivkumar13: Done Rahul

controllers/object_controls.go
Shivkumar13: Done

controllers/object_controls.go
Shivkumar13: Done, used strings.TrimRight()

controllers/object_controls.go
```diff
@@ -3299,32 +3285,38 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
-		return "", err
-	}
-
-	os := release["ID"]
-	if path, ok := RepoConfigPathMap[os]; ok {
+// getRepoConfigPath returns the standard OS specific path for repository configuration files.
+// It re-uses kernelFullVersion to obtain the GPU node OS from NFD labels,
+// ensuring the OS is read from GPU worker nodes rather than the operator pod's node.
+func (n ClusterPolicyController) getRepoConfigPath() (string, error) {
+	_, osTag, _ := kernelFullVersion(n)
```
tariq1890, line 3291: L3292-L3298 & L3309-3315 are duplicated. Can we factor them out into a private method?

Shivkumar13: done Tariq

tariq1890 (approved):

PR #2108, opened 2026-02-09
fix: device plugin needs update rights to nodefeatures

rajatchopra:
## Description

If the device plugin restarts, it will attempt to update the existing nodefeatures object
Also need to update the nodefeatures object for the 'gpu.count' label

## Checklist

## Testing

Manual testing

### DIFF
--- a/assets/state-sandbox-device-plugin/0200_role.yaml
+++ b/assets/state-sandbox-device-plugin/0200_role.yaml
@@ -31,3 +31,4 @@ rules:
   - get
   - list
   - watch
+  - update


assets/state-sandbox-device-plugin/0200_role.yaml
```diff
@@ -31,3 +31,4 @@ rules:
   - get
   - list
   - watch
+  - update
```
tariq1890, line 34: Are you able to `patch` instead to update your NFD CRs ? `update` in general is a riskier operation as it is more sensitive to race conditions.

rajatchopra: No. The GFD seems to use the Update call: 
```
E0210 20:21:14.746771       1 main.go:127] failed to update NodeFeature object "nvidia-features-for-dellr760tdx": nodefeatures.nfd.k8s-sigs.io "nvidia-features-for-dellr760tdx" is forbidden: User "system:serviceaccount:gpu-operator:nvidia-sandbox-device-plugin" cannot update resource "nodefeatures" in API group "nfd.k8s-sigs.io" in the namespace "gpu-operator"
```

rajatchopra: This line: https://github.com/NVIDIA/k8s-device-plugin/blob/main/internal/lm/output.go#L166

tariq1890: Got it. I meant updating the code to use patch instead of update

tariq1890 (approved):

PR #2113, opened 2026-02-10
[Perf] Update labelGPUNodes() to reset updateLabels inside loop

smalleni:
As a golden rule of performance, we should never do work that is not necessary. In the current implementation where updateLabels is declared outside the for loop, if updateLabels becomes true for the first node in a large GPU cluster, it triggers a PATCH request to the API for every subsequent node even when not necessary. This maybe a cheap call but we should never do work that is not necessary. Moving updateLabels inside of the for loop, reduces APi calls and load, along with improving reconcilliation time.

## Checklist

### DIFF
--- a/controllers/state_manager.go
+++ b/controllers/state_manager.go
@@ -488,11 +488,10 @@ func (n *ClusterPolicyController) labelGPUNodes() (bool, int, error) {
 	}
 
 	clusterHasNFDLabels := false
-	updateLabels := false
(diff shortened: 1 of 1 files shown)


tariq1890 (approved):
Thanks for catching this @smalleni ! Agreed with the premise and this optimisation is fair!

rahulait (approved):
LGTM

Issue #1030, opened 2024-10-10
Allow adding custom labels and securityContext to the components deployed by ClusterPolicy


inesshz:
Hello,

* We have requirement in our company where we need to have a custom labels for all Daemonsets and Jobs created by the ClusterPolicy as well as node feature discovery DSs and deployments. 
- For all of the DSs created by ClusterPolicy, we are using: 
                              daemonsets:
                                labels: {}
- However, this label is not applied to the kubernetes job created by the validator DS (cuda-validator) and we were not able to find a way to set labels for this job. The CRD also doesnt allow adding a label for this specific job.  (https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml#L1766)

- Regarding the node feature discovery, there isn't a way to set custom labels as the only labels are being filled through _helpers.tpl:

                            
                            {{- define "node-feature-discovery.labels" -}}
                            helm.sh/chart: {{ include "node-feature-discovery.chart" . }}
                            {{ include "node-feature-discovery.selectorLabels" . }}
                            {{- if .Chart.AppVersion }}
                            app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
                            {{- end }}
                            app.kubernetes.io/managed-by: {{ .Release.Service }}
                            {{- end -}}
                            

                            {{- define "node-feature-discovery.selectorLabels" -}}
                            app.kubernetes.io/name: {{ include "node-feature-discovery.name" . }}
                            app.kubernetes.io/instance: {{ .Release.Name }}
                            {{- end -}}

* Another issue we are facing is that we are not able to set our custom securityContext for the DSs created by the ClusterPolicy. (CRD doesnt allow setting this as well)

Can we get help with templating and customizing the above ? 

Thanks!

justinthelaw:
Definitely have a need for this too. When using hardened containers, the validator pod's containers, like `driver-validator` fail to run due to permissions issues caused by the `securityContext` in the container and intiContainers. We have to apply the following patches:

```bash
kubectl patch daemonset nvidia-operator-validator -n gpu-operator --type='json' -p='[
                {"op": "add", "path": "/spec/template/spec/containers/0/securityContext/runAsUser", "value": 0}
              ]'
              kubectl patch daemonset nvidia-operator-validator -n gpu-operator --type='json' -p='[
[...]

PR #2120, opened 2026-02-11
Add PodSecurityContext support for DaemonSets

faganihajizada:
## Description

Partially fixes [#1030 ](https://github.com/NVIDIA/gpu-operator/issues/1030)

Adds `PodSecurityContext` support to both ClusterPolicy and NVIDIADriver CRs:

- ClusterPolicy — `spec.daemonsets.podSecurityContext` sets pod-level security context for all DaemonSet pods managed by ClusterPolicy
- NVIDIADriver — `spec.podSecurityContext` sets pod-level security context for the nvidia-driver DaemonSet pod

### Design notes

- Pod-level SecurityContext is full replace, not merge. This matches the existing pattern for tolerations and priorityClassName in `applyCommonDaemonsetConfig`. No embedded DaemonSet manifest currently sets a pod-level SecurityContext, so there is nothing to lose on replace.
- For NVIDIADriver, the PodSecurityContext is templated directly into the driver DaemonSet manifest, following the same Go template pattern used for other optional spec fields (tolerations, labels, annotations).

## Checklist

## Testing

- Unit tests for `mergeSecurityContext` (5 cases: nil target, nil defaults, empty target, merge with existing, no-overwrite)
- Unit tests for `applyCommonDaemonsetConfig` (4 new cases: podSecurityContext, nil container SecurityContext, merge into existing, initContainers)
- Unit tests for `applyDaemonsetMetadataToPod` (4 cases: no-op, label skip, annotations, nil labels)
- Unit tests for `applyCommonDaemonsetMetadata` (existing tests still pass)


deployments/gpu-operator/charts/node-feature-discovery/values.yaml
```diff
@@ -10,6 +10,9 @@ nameOverride: ""
 fullnameOverride: ""
 namespaceOverride: ""
 
+# commonLabels are added to all resources created by this chart (DaemonSets, Deployments, etc.)
+commonLabels: {}
```
tariq1890: We don't really edit the NFD subchart directly. We import it upstream

tariq1890: Also, this change may be out-of-scope for this PR. Let's limit this PR to just SecurityContext-related changes

controllers/object_controls.go
```diff
@@ -818,6 +810,49 @@ func applyCommonDaemonsetMetadata(obj *appsv1.DaemonSet, dsSpec *gpuv1.Daemonset
 	}
 }
 
+// mergeSecurityContext copies fields from defaults into target only when the target's field is nil.
+func mergeSecurityContext(target, defaults *corev1.SecurityContext) {
```
tariq1890: Can we use a library like `mergo` instead to perform struct merges? 

https://github.com/darccio/mergo

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -245,9 +245,15 @@ type DaemonsetsSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Rolling update configuration for all DaemonSet pods"
 	RollingUpdate *RollingUpdateSpec `json:"rollingUpdate,omitempty"`
+
+	// Optional: Set pod-level security context for all DaemonSet pods
+	PodSecurityContext *corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`
+
+	// Optional: Set default container-level security context for all DaemonSet pods
+	SecurityContext *corev1.SecurityContext `json:"securityContext,omitempty"`
```
tariq1890: I see that we are applying `SecurityContext` to all of the containers and init-containers of a DaemonSet. Won't setting `PodSecurityContext` make this redundant?

If yes, we should probably just stick with `PodSecurityContext`

deployments/gpu-operator/charts/node-feature-discovery/values.yaml
faganihajizada: I see. this makes sense, thanks!

faganihajizada: Reverted the NFD subchart changes so we don’t edit the upstream NFD chart.

controllers/object_controls.go
faganihajizada: I simplified the implementation to only add `PodSecurityContext` and removed the container-level `SecurityContext` field and the merge logic. There’s no struct merge anymore, we just assign the pod `SecurityContext` when set. I believe `mergo` isn’t needed for this PR.

api/nvidia/v1/clusterpolicy_types.go
faganihajizada: Agreed. I’ve removed the container-level `SecurityContext` field and all related logic; the PR now only adds `PodSecurityContext`. Thanks!

rahulait:
We need to add this to nvidiadriver CR as well if we are adding it for clusterpolicy. Basically here: https://github.com/NVIDIA/gpu-operator/blob/main/api/nvidia/v1alpha1/nvidiadriver_types.go#L199

After that, we need to update the nvidia-driver specific daemonset manifest to also include securityContext if set here: https://github.com/NVIDIA/gpu-operator/blob/main/manifests/state-driver/0500_daemonset.yaml#L86

tariq1890:
@faganihajizada Thanks again for the contribution! Can you amend the commit message and PR title so that it reflects the changes after review?

Please commit with the `--signoff` option so that the DCO checks are satisfied

faganihajizada:
> @faganihajizada Thanks again for the contribution! Can you amend the commit message and PR title so that it reflects the changes after review?
> 
> Please commit with the `--signoff` option so that the DCO checks are satisfied

Thanks @tariq1890 I updated commit message and also addressed comment from @rahulait

faganihajizada:
@rahulait seems like there's no mechanism to regenerate the golden test files in `internal/state/testdata/golden/` when the API types change. Is there a preferred way to handle this, or do you typically update them manually? 

I can add a small `UPDATE_GOLDEN=1` flag in the tests in a follow-up PR if it helps.

rahulait:
@faganihajizada unit-tests will need updates so that CI can pass. I think they are generated manually for now as I don't see any script doing that. You can just update the failing test for now to match the new fields added.

internal/state/testdata/golden/driver-additional-configs.yaml
```diff
@@ -185,7 +185,7 @@ spec:
             fieldRef:
               fieldPath: status.hostIP
         - name: DRIVER_CONFIG_DIGEST
-          value: "3041669332"
+          value: "4192128261"
```
tariq1890, line 188: I am a bit concerned that the `DRIVER_CONFIG_DIGEST` is changed here even though we don't set the `.Driver.Spec.PodSecurityContext` field. Is this because the struct schema has changed? 

We need to keep in mind that this change will cancel fast path re-deployments where is was previously possible.

cc @cdesiniotis @karthikvetrivel

faganihajizada: can I help on this?

karthikvetrivel: Taking a look at this. No action required on your side right now @faganihajizada, thanks!

PR #2121, opened 2026-02-11
Restore static MIG configs

rajathagasthya:
This change restores static MIG configs due to issues with R535 drivers not being able to detect MIG profiles. The configmap is volume mounted onto the MIG manager container as config-default.yaml, so that MIG manager can fall back to this config whenever generation fails.

## Checklist

## Testing

Tested as part of https://github.com/NVIDIA/mig-parted/pull/312

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -101,6 +101,8 @@ const (
 	ValidatorRuntimeClassEnvName = "VALIDATOR_RUNTIME_CLASS"
 	// MigStrategyEnvName indicates env name for passing MIG strategy
 	MigStrategyEnvName = "MIG_STRATEGY"
+	// MigPartedDefaultConfigMapName indicates name of ConfigMap containing default mig-parted config
+	MigPartedDefaultConfigMapName = "default-mig-parted-config"
 	// MigDefaultGPUClientsConfigMapName indicates name of ConfigMap containing default gpu-clients
 	MigDefaultGPUClientsConfigMapName = "default-gpu-clients"
 	// DCGMRemoteEngineEnvName indicates env name to specify remote DCGM host engine ip:port
@@ -539,6 +541,14 @@ func createConfigMap(n ClusterPolicyController, configMapIdx int) (gpuv1.State, 
 			return gpuv1.NotReady, err
 		}
 		return gpuv1.Disabled, nil
+	}
+
+	// avoid creating default 'mig-parted-config' ConfigMap if custom one is provided
+	if obj.Name == MigPartedDefaultConfigMapName {
+		if name, isCustom := gpuv1.GetConfigMapName(config.MIGManager.Config, MigPartedDefaultConfigMapName); isCustom {
+			logger.Info("Not creating resource, custom ConfigMap provided", "Name", name)
+			return gpuv1.Ready, nil
+		}
 	}
 
 	// avoid creating default 'gpu-clients' ConfigMap if custom one is provided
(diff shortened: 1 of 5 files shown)


controllers/object_controls.go
```diff
@@ -1941,25 +1951,27 @@ func TransformMIGManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+			}
+		}
+		break
+	}
 
-		// NOTE: assumes ConfigMap has key "config.yaml"
+	if hasCustomConfig {
 		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), "CONFIG_FILE", "/mig-parted-config/config.yaml")
+	} else {
+		setContainerEnv(&(obj.Spec.Template.Spec.Containers[0]), "FALLBACK_CONFIG_FILE", "/mig-parted-config/config-fallback.yaml")
```
tariq1890: Can we name it as `DEFAULT_CONFIG_FILE` instead?

rajathagasthya: Renamed.

controllers/object_controls.go
```diff
@@ -1941,25 +1951,27 @@ func TransformMIGManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
-			migConfigMount,
-		)
+	for i, vol := range obj.Spec.Template.Spec.Volumes {
+		if !strings.Contains(vol.Name, "mig-parted-config") {
+			continue
+		}
+		name, _ := gpuv1.GetConfigMapName(config.MIGManager.Config, MigPartedDefaultConfigMapName)
+		obj.Spec.Template.Spec.Volumes[i].ConfigMap.Name = name
+		if !hasCustomConfig {
+			// Remap key to config-default.yaml for the default config path
```
cdesiniotis: Let's expand the comment here so the reader knows why we want the default config file to be named `config-default.yaml` when it gets mounted into the mig-manager container.

Or add a comment below when setting the `CONFIG_FILE` / `DEFAULT_CONFIG_FILE` envvars.

controllers/object_controls.go
```diff
@@ -1941,25 +1951,27 @@ func TransformMIGManager(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 	setRuntimeClassName(&obj.Spec.Template.Spec, config, n.runtime)
 	setNRIPluginAnnotation(&obj.Spec.Template.ObjectMeta, &config.CDI, obj.Spec.Template.Spec.Containers[0].Name)
 
-	// mount custom mig-parted config if provided
+	// Set ConfigMap name for mig-parted-config volume
 	hasCustomConfig := config.MIGManager.Config != nil && config.MIGManager.Config.Name != ""
-	if hasCustomConfig {
-		migConfigVolume := createConfigMapVolume(config.MIGManager.Config.Name, nil)
```
tariq1890, line 1947: So we no longer need to create the configMap volume if the user passes in a custom mig configmap?

controllers/transforms_test.go
```diff
@@ -1629,20 +1647,21 @@ func TestTransformMigManager(t *testing.T) {
 					{Name: NvidiaCDIHookPathEnvName, Value: "/path/to/install/toolkit/nvidia-cdi-hook"},
 				},
 				VolumeMounts: []corev1.VolumeMount{
-					{Name: "mig-parted-config", MountPath: "/mig-parted-config", ReadOnly: true},
```
cdesiniotis, line 1632: I see this volume mount was read only before. Do we want to retain that?

controllers/object_controls.go
cdesiniotis: The ConfigMap volume is defined in the daemonset template.

controllers/object_controls.go
```diff
@@ -541,6 +543,14 @@ func createConfigMap(n ClusterPolicyController, configMapIdx int) (gpuv1.State,
 		return gpuv1.Disabled, nil
 	}
 
+	// avoid creating default 'mig-parted-config' ConfigMap if custom one is provided
```
tariq1890, line 546: Why is this showing up in the diff? I already see this code block in the main branch. Do we need to rebase this PR?

rajathagasthya: I don't see this on the main branch. Are you maybe thinking of lines below this? https://github.com/NVIDIA/gpu-operator/blob/dc552ddc0e8a35fe05134ab6f149a9efc20b1695/controllers/object_controls.go#L544-L550

FYI, this change was restored from before the previous dynamic MIG config changes. So if your main is not current, you will see this code block in an older version.

tariq1890: Ah yes, you're right. These code blocks look way too similar :)

tariq1890:
Thanks @rajathagasthya !
Let's get this in after https://github.com/NVIDIA/mig-parted/pull/312 is merged

PR #2123, opened 2026-02-13
[improvement] : remove driver upgrade label from nodes

rahulait:
this commit removes driver upgrade label from nodes which don't have any driver pod running on them

## Description

This PR adds automatic cleanup of stale upgrade labels from nodes where GPU driver pods are no longer scheduled, addressing the edge case where `nodeSelector` changes cause pods to terminate but leave `nvidia.com/gpu-driver-upgrade-state` labels behind.

## Problem

When a `NVIDIADriver` CR's `nodeSelector` is updated:
1. The DaemonSet updates its pod template with the new selector
2. Pods on nodes no longer matching the selector are terminated
3. The upgrade labels (`nvidia.com/gpu-driver-upgrade-state`) remain on those nodes indefinitely
4. This creates confusion and incorrect upgrade state tracking

Example scenario:
```yaml
# Initial: nodeSelector targets GPU nodes with specific label
spec:
  nodeSelector:
    nvidia.com/gpu: "true"

# Updated: nodeSelector narrows to specific GPU type
spec:
  nodeSelector:
    nvidia.com/gpu: "true"
    nvidia.com/gpu-type: "A100"
```

Nodes without `nvidia.com/gpu-type: "A100"` lose their driver pods but keep upgrade labels.

## Solution
Added `clearUpgradeLabelsWhereDriverNotRunning()` function to the upgrade controller that:

* Runs automatically every 2 minutes during normal upgrade controller reconciliation
* Identifies stale labels by finding nodes with upgrade labels but no driver pods
* Protects active upgrades by skipping nodes in ClusterUpgradeState.NodeStates (actively managed by upgrade process)
* Removes labels safely using Patch() for efficient, conflict-resistant updates

## Checklist

### DIFF
--- a/controllers/upgrade_controller.go
+++ b/controllers/upgrade_controller.go
@@ -153,6 +153,10 @@ func (r *UpgradeReconciler) reconcileClusterPolicyDriverUpgrades(ctx context.Con
 		return ctrl.Result{}, err
 	}
 
+	if err := r.clearStaleUpgradeLabels(ctx, state, driverLabel, clusterPolicyCtrl.operatorNamespace); err != nil {
+		r.Log.Error(err, "Failed to clear stale upgrade labels")
+	}
+
 	reqLogger.Info("Propagate state to state manager")
 	reqLogger.V(consts.LogLevelDebug).Info("Current cluster upgrade state", "state", state)
 
@@ -240,6 +244,10 @@ func (r *UpgradeReconciler) reconcileNVIDIADriverUpgrades(ctx context.Context, r
 		return ctrl.Result{}, err
 	}
 
+	if err := r.clearStaleUpgradeLabels(ctx, clusterState, map[string]string{AppComponentLabelKey: DriverAppComponentLabelValue}, clusterPolicyCtrl.operatorNamespace); err != nil {
+		r.Log.Error(err, "Failed to clear stale upgrade labels")
+	}
+
 	// Partition the cluster upgrade state into per-NVIDIADriver buckets by reading the
 	// nvidia.com/gpu-operator.driver.owner label from each node.
 	statesByNVD := make(map[string]*upgrade.ClusterUpgradeState)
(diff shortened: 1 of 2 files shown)


controllers/upgrade_controller.go
```diff
@@ -198,6 +205,74 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			if err := r.Patch(ctx, node, client.MergeFrom(nodeCopy)); err != nil {
+				r.Log.Error(err, "Failed to clear upgrade label from node", "node", node.Name)
+				// Continue with other nodes even if one fails
+				continue
+			}
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
rahulait: added

controllers/upgrade_controller_test.go
```diff
@@ -0,0 +1,440 @@
...
+)
+
+const (
+	testNamespace = "gpu-operator"
+	upgradeLabel  = "nvidia.com/gpu-driver-upgrade-state"
+)
+
+func TestClearUpgradeLabelsWhereDriverNotRunning(t *testing.T) {
+	// Set the driver name so the upgrade label key matches
+	upgrade.SetDriverName("gpu")
```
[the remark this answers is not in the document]
rahulait: updated

controllers/upgrade_controller_test.go
```diff
@@ -0,0 +1,440 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+const (
+	testNamespace = "gpu-operator"
+	upgradeLabel  = "nvidia.com/gpu-driver-upgrade-state"
+)
+
+func TestClearUpgradeLabelsWhereDriverNotRunning(t *testing.T) {
```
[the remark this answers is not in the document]
rahulait: done

controllers/upgrade_controller.go
```diff
@@ -152,6 +152,14 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
+	// Clear stale upgrade labels from nodes that no longer have driver pods
+	// Use the built state to avoid duplicate API calls and to skip nodes actively being upgraded
+	// This is best-effort cleanup that runs every reconciliation
+	if err := r.clearUpgradeLabelsWhereDriverNotRunning(ctx, state, driverLabel, clusterPolicyCtrl.operatorNamespace); err != nil {
```
tariq1890: Let's go with this name as it is more concise. The comment already explains what it considers as a stale upgrade label
```suggestion
	if err := r.clearStaleUpgradeLabels(ctx, state, driverLabel, clusterPolicyCtrl.operatorNamespace); err != nil {
```

controllers/upgrade_controller.go
```diff
@@ -152,6 +152,14 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// Log the error but continue with the upgrade process, as this is a best-effort cleanup and should not block upgrades
+		r.Log.Error(err, "Failed to clear stale upgrade labels")
```
tariq1890: Should this be Warning instead?

controllers/upgrade_controller.go
```diff
@@ -198,6 +206,102 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	// List driver DaemonSets to check which nodes should have pods
+	// This protects against removing labels during pod recreation (e.g., during rolling updates)
+	dsList := &appsv1.DaemonSetList{}
+	if err := r.List(ctx, dsList, client.InNamespace(namespace), client.MatchingLabels(driverLabel)); err != nil {
+		return fmt.Errorf("failed to list driver DaemonSets: %w", err)
+	}
+
+	// Build a set of nodes that DaemonSets would schedule to
+	// This protects nodes during pod recreation (e.g., rolling updates where pod is temporarily missing)
+	nodesShouldHavePods := make(map[string]bool)
```
tariq1890: How about this name?
```suggestion
	nodesSelected := make(map[string]bool)
```

I was initially thinking of this, but chose the above for brevity. Let me know what you think of the two

```suggestion
	nodesSelectedByDriverDS := make(map[string]bool)
```

controllers/upgrade_controller.go
rahulait: Sure, updated.

controllers/upgrade_controller.go
rahulait: Yeah, warning makes more sense. Updated.

controllers/upgrade_controller.go
rahulait: Updated with `nodesSelected`.

controllers/upgrade_controller_test.go
```diff
@@ -0,0 +1,435 @@
...
+				objs = append(objs, pod)
+			}
+
+			scheme := runtime.NewScheme()
+			_ = corev1.AddToScheme(scheme)
+			_ = appsv1.AddToScheme(scheme)
+
+			baseClient := fake.NewClientBuilder().
+				WithScheme(scheme).
+				WithObjects(objs...).
```
rajathagasthya: Consider adding an interceptor here to track patched nodes instead of a new `patchTrackingClient` struct.

```go
patchedNodes := map[string]*corev1.Node{}
baseClient := fake.NewClientBuilder().
    WithScheme(scheme).
    WithObjects(objs...).
    WithInterceptorFuncs(interceptor.Funcs{
        Patch: func(ctx context.Context, client client.WithWatch, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
            if node, ok := obj.(*corev1.Node); ok {
                patchedNodes[node.Name] = node.DeepCopy()
            }
            return client.Patch(ctx, obj, patch, opts...)
        }}).
    Build()
```

controllers/upgrade_controller_test.go
```diff
@@ -0,0 +1,435 @@
...
+		upgrade.SetDriverName(previousDriverName)
+	})
+	upgrade.SetDriverName("gpu")
+
+	t.Run("handles list nodes error gracefully", func(t *testing.T) {
+		scheme := runtime.NewScheme()
+		_ = corev1.AddToScheme(scheme)
+
+		// Create a client that will fail on List operations
+		fakeClient := &errorClient{
```
rajathagasthya: Same here. Build a regular fake client with an interceptor that returns an error on the operations you want. But my preference is to remove this test case. This is just testing the function returns an error when the client can't list objects; doesn't seem valuable to me.

PR #2136, opened 2026-02-18
Delete driver pod in update-clusterpolicy

JunAr7112:
## Description

Fix flaking failures in the update-clusterpolicy test. The issue here is that when the NVIDIADriver CR is patched with new labels, the pod template in the DaemonSet is updated with the new labels, but the existing pod isn't automatically restarted.

## Checklist

### DIFF
--- a/tests/scripts/update-clusterpolicy.sh
+++ b/tests/scripts/update-clusterpolicy.sh
@@ -232,6 +232,14 @@ test_custom_labels_override() {
 
   operands="nvidia-driver-daemonset nvidia-container-toolkit-daemonset nvidia-operator-validator gpu-feature-discovery nvidia-dcgm-exporter nvidia-device-plugin-daemonset"
 
+  # Wait for the operator to update the DaemonSet pod templates with new labels (driver daemonset as representative).
+  echo "Waiting for DaemonSet pod template to be updated with new labels..."
+  kubectl wait daemonset -n "$TEST_NAMESPACE" nvidia-driver-daemonset --for=jsonpath='{.spec.template.metadata.labels.cloudprovider}'=aws --timeout=120s
+
+  # Delete driver pod to force recreation with updated labels. Existing pods are not automatically restarted due to the DaemonSet OnDelete updateStrategy.
+  echo "Deleting driver pod to trigger recreation with updated labels..."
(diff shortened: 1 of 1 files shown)


rahulait:
@JunAr7112 is the description correct? I see you had mentioned `The issue here is that when the NVIDIADriver CR is patched`, but it looks like its patching clusterpolicy and there is no nvidiadriver CR here.

tests/scripts/update-clusterpolicy.sh
```diff
@@ -232,6 +232,14 @@ test_custom_labels_override() {
 
   operands="nvidia-driver-daemonset nvidia-container-toolkit-daemonset nvidia-operator-validator gpu-feature-discovery nvidia-dcgm-exporter nvidia-device-plugin-daemonset"
 
+  # Wait for the operator to update the DaemonSet pod templates with new labels (driver daemonset as representative).
+  echo "Waiting for DaemonSet pod template to be updated with new labels..."
+  kubectl wait daemonset -n "$TEST_NAMESPACE" nvidia-driver-daemonset --for=jsonpath='{.spec.template.metadata.labels.cloudprovider}'=aws --timeout=120s
+
+  # Delete driver pod to force recreation with updated labels. Existing pods are not automatically restarted due to the DaemonSet OnDelete updateStrategy.
+  echo "Deleting driver pod to trigger recreation with updated labels..."
+  kubectl delete pod -n "$TEST_NAMESPACE" -l app=nvidia-driver-daemonset --ignore-not-found
```
rahulait, line 241: Upgrade-controller is responsible for deleting the pod. We shouldn't be deleting manually in the tests. When the labels for daemonset changes, daemonset hash changes. Upgrade controller then detects that hash of daemonset has changed and then it labels the node as "upgrade-required". Then upgrade-controller gracefully terminates the driver pod and new one gets created after that.

If its failing to delete sometimes, we need to gather more logs as to why and maybe it takes some extra time to catch up. So I would try to understand and see if increasing time for the test(like timing out after 10 mins to timeout after 15 mins) helps or not.

JunAr7112: Ok I see. I'll try looking at the logs here and verifying why it isn't deleting.

PR #2138, opened 2026-02-19
remove dependency on /etc/os-release host mount

tariq1890:
This PR builds on top of #2101 - using the NFD labels to get the `/etc/os-release` information instead of directly querying the node's `/etc/os-release` file via a host mount.

With the existing approach, the gpu-operator assumes that the node on which the gpu-operator pod is scheduled has the same OS as the GPU node. There are cases where users may choose to deploy the gpu-operator pod on a CPU node having an OS different from that of the GPU node. These changes enable the gpu-operator to function in these scenarios.

### DIFF
--- a/deployments/gpu-operator/templates/operator.yaml
+++ b/deployments/gpu-operator/templates/operator.yaml
@@ -59,10 +59,6 @@ spec:
               fieldPath: metadata.namespace
         - name: "DRIVER_MANAGER_IMAGE"
           value: "{{ include "driver-manager.fullimage" . }}"
-        volumeMounts:
-          - name: host-os-release
-            mountPath: "/host-etc/os-release"
-            readOnly: true
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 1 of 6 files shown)


deployments/gpu-operator/templates/operator.yaml
```diff
@@ -59,10 +59,6 @@ spec:
               fieldPath: metadata.namespace
         - name: "DRIVER_MANAGER_IMAGE"
           value: "{{ include "driver-manager.fullimage" . }}"
-        volumeMounts:
-          - name: host-os-release
-            mountPath: "/host-etc/os-release"
-            readOnly: true
```
rahulait, line 65: Do we need to do this change in https://github.com/NVIDIA/gpu-operator/blob/main/config/manager/manager.yaml? Who uses those files? Openshift?

tariq1890: Thanks for catching this, will make the change

tariq1890: Updated, please check now

PR #2142, opened 2026-02-20
Changes to support kata device plugin and mode

jojimt:
## Description

Kata/Coco uses a new sandbox device plugin. We need to update the operator to natively support it. To this end,
the following has been added to the cluster policy:
 - a mode field to sandboxWorkloads, with ["kubevirt", "kata"] as valid values
 - a kataDevicePlugin section

## Checklist

### DIFF
--- a/controllers/state_manager_test.go
+++ b/controllers/state_manager_test.go
@@ -229,3 +229,145 @@ func TestValidateClusterPolicySpec(t *testing.T) {
 		})
 	}
 }
+
+func TestGetEffectiveStateLabels(t *testing.T) {
+	// getEffectiveStateLabels returns labels for workload config and sandbox mode.
+	// For container and vm-vgpu, mode has no effect. For vm-passthrough, mode selects
+	// sandbox-device-plugin (kubevirt) vs kata-device-plugin (kata).
+	t.Run("container", func(t *testing.T) {
+		got := getEffectiveStateLabels(gpuWorkloadConfigContainer, "kubevirt")
+		require.NotNil(t, got)
+		require.Contains(t, got, "nvidia.com/gpu.deploy.device-plugin")
+		require.Equal(t, "true", got["nvidia.com/gpu.deploy.device-plugin"])
+	})
+	t.Run("vm-vgpu", func(t *testing.T) {
+		got := getEffectiveStateLabels(gpuWorkloadConfigVMVgpu, "kata")
+		require.NotNil(t, got)
+		require.Contains(t, got, "nvidia.com/gpu.deploy.sandbox-device-plugin")
+		require.Equal(t, "true", got["nvidia.com/gpu.deploy.sandbox-device-plugin"])
+	})
+	// vm-passthrough: test kubevirt first (map has sandbox-device-plugin), then kata.
+	t.Run("vm-passthrough-kubevirt", func(t *testing.T) {
+		got := getEffectiveStateLabels(gpuWorkloadConfigVMPassthrough, string(gpuv1.KubeVirt))
+		require.NotNil(t, got)
+		require.Contains(t, got, kubevirtDevicePluginDeployLabelKey)
+		require.Equal(t, "true", got[kubevirtDevicePluginDeployLabelKey])
+		require.NotContains(t, got, kataDevicePluginDeployLabelKey)
+	})
+	t.Run("vm-passthrough-kata", func(t *testing.T) {
+		got := getEffectiveStateLabels(gpuWorkloadConfigVMPassthrough, string(gpuv1.Kata))
+		require.NotNil(t, got)
+		require.Contains(t, got, kataDevicePluginDeployLabelKey)
+		require.Equal(t, "true", got[kataDevicePluginDeployLabelKey])
+		require.NotContains(t, got, kubevirtDevicePluginDeployLabelKey)
+	})
+	t.Run("invalid config", func(t *testing.T) {
+		got := getEffectiveStateLabels("invalid", "kubevirt")
+		require.Nil(t, got)
+	})
+}
+
+func TestRemoveAllGPUStateLabels(t *testing.T) {
+	// removeAllGPUStateLabels removes all gpuStateLabels keys plus kata-device-plugin and mig-manager.
+	t.Run("removes kata device plugin label", func(t *testing.T) {
+		labels := map[string]string{
+			kataDevicePluginDeployLabelKey: "true",
+			"other":                        "keep",
+		}
+		modified := removeAllGPUStateLabels(labels)
+		require.True(t, modified)
+		require.NotContains(t, labels, kataDevicePluginDeployLabelKey)
+		require.Equal(t, "keep", labels["other"])
+	})
+	t.Run("removes sandbox deploy label", func(t *testing.T) {
+		labels := map[string]string{
+			kubevirtDevicePluginDeployLabelKey: "true",
+		}
+		modified := removeAllGPUStateLabels(labels)
+		require.True(t, modified)
+		require.Empty(t, labels[kubevirtDevicePluginDeployLabelKey])
+	})
+}
+
+func TestIsStateEnabled_SandboxAndKataDevicePlugin(t *testing.T) {
+	boolTrue := ptr.To(true)
+	boolFalse := ptr.To(false)
+	tests := []struct {
+		name           string
+		sandboxEnabled bool
+		spec           gpuv1.ClusterPolicySpec
+		stateName      string
+		wantEnabled    bool
+	}{
+		{
+			name:           "state-sandbox-device-plugin enabled when sandbox+plugin+mode kubevirt",
+			sandboxEnabled: true,
+			spec: gpuv1.ClusterPolicySpec{
+				SandboxWorkloads:    gpuv1.SandboxWorkloadsSpec{Enabled: boolTrue, Mode: "kubevirt"},
+				SandboxDevicePlugin: gpuv1.SandboxDevicePluginSpec{Enabled: boolTrue},
+			},
+			stateName:   "state-sandbox-device-plugin",
+			wantEnabled: true,
+		},
+		{
+			name:           "state-sandbox-device-plugin disabled when mode kata",
+			sandboxEnabled: true,
+			spec: gpuv1.ClusterPolicySpec{
+				SandboxWorkloads:    gpuv1.SandboxWorkloadsSpec{Enabled: boolTrue, Mode: "kata"},
+				SandboxDevicePlugin: gpuv1.SandboxDevicePluginSpec{Enabled: boolTrue},
+			},
+			stateName:   "state-sandbox-device-plugin",
+			wantEnabled: false,
+		},
+		{
+			name:           "state-kata-device-plugin enabled when sandbox+kata plugin+mode kata",
+			sandboxEnabled: true,
+			spec: gpuv1.ClusterPolicySpec{
+				SandboxWorkloads:        gpuv1.SandboxWorkloadsSpec{Enabled: boolTrue, Mode: "kata"},
+				KataSandboxDevicePlugin: gpuv1.KataDevicePluginSpec{ComponentCommonSpec: gpuv1.ComponentCommonSpec{Enabled: boolTrue}},
+			},
+			stateName:   "state-kata-device-plugin",
+			wantEnabled: true,
+		},
+		{
+			name:           "state-kata-device-plugin disabled when mode kubevirt",
+			sandboxEnabled: true,
+			spec: gpuv1.ClusterPolicySpec{
+				SandboxWorkloads:        gpuv1.SandboxWorkloadsSpec{Enabled: boolTrue, Mode: "kubevirt"},
+				KataSandboxDevicePlugin: gpuv1.KataDevicePluginSpec{ComponentCommonSpec: gpuv1.ComponentCommonSpec{Enabled: boolTrue}},
+			},
+			stateName:   "state-kata-device-plugin",
+			wantEnabled: false,
+		},
+		{
+			name:           "state-kata-device-plugin disabled when KataSandboxDevicePlugin.Enabled false",
+			sandboxEnabled: true,
+			spec: gpuv1.ClusterPolicySpec{
+				SandboxWorkloads:        gpuv1.SandboxWorkloadsSpec{Enabled: boolTrue, Mode: "kata"},
+				KataSandboxDevicePlugin: gpuv1.KataDevicePluginSpec{ComponentCommonSpec: gpuv1.ComponentCommonSpec{Enabled: boolFalse}},
+			},
+			stateName:   "state-kata-device-plugin",
+			wantEnabled: false,
+		},
+		{
+			name:           "state-kata-device-plugin disabled when sandbox workloads disabled",
+			sandboxEnabled: false,
+			spec: gpuv1.ClusterPolicySpec{
+				SandboxWorkloads:        gpuv1.SandboxWorkloadsSpec{Enabled: boolTrue, Mode: "kata"},
+				KataSandboxDevicePlugin: gpuv1.KataDevicePluginSpec{ComponentCommonSpec: gpuv1.ComponentCommonSpec{Enabled: boolTrue}},
+			},
+			stateName:   "state-kata-device-plugin",
+			wantEnabled: false,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			n := ClusterPolicyController{
+				singleton:      &gpuv1.ClusterPolicy{Spec: tc.spec},
+				sandboxEnabled: tc.sandboxEnabled,
+			}
+			got := n.isStateEnabled(tc.stateName)
+			require.Equal(t, tc.wantEnabled, got, "isStateEnabled(%q)", tc.stateName)
+		})
+	}
+}
--- a/controllers/transforms_test.go
+++ b/controllers/transforms_test.go
@@ -2778,6 +2778,84 @@ func TestTransformSandboxValidator(t *testing.T) {
 	}
 }
 
+func TestTransformKataDevicePlugin(t *testing.T) {
+	resources := corev1.ResourceRequirements{
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("100m"),
+			corev1.ResourceMemory: resource.MustParse("128Mi"),
+		},
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("50m"),
+			corev1.ResourceMemory: resource.MustParse("64Mi"),
+		},
+	}
+	testCases := []struct {
+		description string
+		ds          Daemonset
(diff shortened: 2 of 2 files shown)


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -197,6 +199,11 @@ type SandboxWorkloadsSpec struct {
 	// +kubebuilder:validation:Enum=container;vm-passthrough;vm-vgpu
 	// +kubebuilder:default=container
 	DefaultWorkload string `json:"defaultWorkload,omitempty"`
+	// Mode indicates the sandbox mode (e.g. kube-virt vs kata)
+	// worker nodes in the cluster for
+	// +kubebuilder:validation:Enum=kube-virt;kata
+	// +kubebuilder:default=kube-virt
```
tariq1890: I see that the usage of `kubevirt` is far more common in the web.
```suggestion
	// Mode indicates the sandbox mode (e.g. kubevirt vs kata)
	// worker nodes in the cluster for
	// +kubebuilder:validation:Enum=kubevirt;kata
	// +kubebuilder:default=kubevirt
```

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1460,6 +1467,63 @@ type MIGGPUClientsConfigSpec struct {
 	Name string `json:"name,omitempty"`
 }
 
+// ImageSpec defines shared fields for component images
+type ImageSpec struct {
+	// NVIDIA Sandbox Device Plugin image repository
+	// +kubebuilder:validation:Optional
+	Repository string `json:"repository,omitempty"`
+
+	// NVIDIA Sandbox Device Plugin image name
```
tariq1890: Since this is a part of a potentially reusable object, we should remove references to sandbox device plugin here imo

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1460,6 +1467,63 @@ type MIGGPUClientsConfigSpec struct {
...
+	// +kubebuilder:validation:Pattern=[a-zA-Z0-9\-]+
+	Image string `json:"image,omitempty"`
+
+	// NVIDIA Sandbox Device Plugin image tag
```
tariq1890: remove sandbox device plugin reference here

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1460,6 +1467,63 @@ type MIGGPUClientsConfigSpec struct {
 	Name string `json:"name,omitempty"`
 }
 
+// ImageSpec defines shared fields for component images
+type ImageSpec struct {
+	// NVIDIA Sandbox Device Plugin image repository
```
tariq1890: remove sandbox device plugin reference here

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -94,6 +94,8 @@ type ClusterPolicySpec struct {
 	CCManager CCManagerSpec `json:"ccManager,omitempty"`
 	// HostPaths defines various paths on the host needed by GPU Operator components
 	HostPaths HostPathsSpec `json:"hostPaths,omitempty"`
+	// KataDevicePlugin component spec
+	KataDevicePlugin KataDevicePluginSpec `json:"kataDevicePlugin,omitempty"`
```
cdesiniotis: Thoughts on naming this `kataSandboxDevicePlugin`?

controllers/object_controls.go
```diff
@@ -1740,6 +1741,35 @@ func TransformSandboxDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPo
 	return nil
 }
 
+// TransformKataDevicePlugin transforms kata-device-plugin daemonset with required config as per ClusterPolicy
+func TransformKataDevicePlugin(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n ClusterPolicyController) error {
```
cdesiniotis, line 1669: Do we also need to transform the validation init container?

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -197,6 +199,11 @@ type SandboxWorkloadsSpec struct {
 	// +kubebuilder:validation:Enum=container;vm-passthrough;vm-vgpu
 	// +kubebuilder:default=container
 	DefaultWorkload string `json:"defaultWorkload,omitempty"`
+	// Mode indicates the sandbox mode (e.g. kubevirt vs kata)
+	// worker nodes in the cluster for
```
cdesiniotis: Was this line intended? It doesn't flow with the previous line.

controllers/object_controls.go
cdesiniotis: I see there is no validation init container in the daemonset manifest you added. Was that intentional?

deployments/gpu-operator/values.yaml
```diff
@@ -508,6 +510,17 @@ sandboxDevicePlugin:
   env: []
   resources: {}
 
+# Kata device plugin (used when sandboxWorkloads.mode is "kata"). No enabled flag — implied by mode.
+kataDevicePlugin:
+  repository: nvcr.io/nvidia
+  image: kata-gpu-device-plugin
```
rajatchopra: We do plan to change this right? Version as well for the release.

controllers/state_manager.go
```diff
@@ -338,6 +340,27 @@ func getWorkloadConfig(labels map[string]string, sandboxEnabled bool) (string, e
...
 }
 
+// getEffectiveStateLabels returns the state labels to apply for the given workload config and sandbox mode.
+// When config is vm-passthrough and mode is "kata", returns labels with kata-device-plugin instead of sandbox-device-plugin.
+func getEffectiveStateLabels(config, mode string) map[string]string {
+	base, ok := gpuStateLabels[config]
+	if !ok {
+		return nil
+	}
+	if config == gpuWorkloadConfigVMPassthrough && mode == "kata" {
```
cdesiniotis: Why are we only doing this for `vm-passthrough`? I would assume we should also do this transformation for `vm-vgpu`.

controllers/state_manager.go
```diff
@@ -338,6 +340,27 @@ func getWorkloadConfig(labels map[string]string, sandboxEnabled bool) (string, e
...
+			if k == "nvidia.com/gpu.deploy.sandbox-device-plugin" {
+				continue
+			}
+			out[k] = v
+		}
+		out[kataDevicePluginDeployLabelKey] = "true"
+		return out
+	}
+	return base
+}
```
cdesiniotis, line 368: What about simplifying to below?
```suggestion
func getEffectiveStateLabels(config, mode string) map[string]string {
	labels, ok := gpuStateLabels[config]
	if !ok {
		return nil
	}
	
	if mode != "kata" {
	  return labels
	}
	
	delete(labels, "nvidia.com/gpu.deploy.sandbox-device-plugin")
	labels[kataDevicePluginDeployLabelKey] = "true"
	return labels
}
```

controllers/state_manager.go
```diff
@@ -350,6 +373,10 @@ func removeAllGPUStateLabels(labels map[string]string) bool {
 			}
 		}
 	}
+	if _, ok := labels[kataDevicePluginDeployLabelKey]; ok {
+		delete(labels, kataDevicePluginDeployLabelKey)
+		modified = true
+	}
```
cdesiniotis, line 385: As an alternative -- what if we added additional keys to the `gpuStateLabels` map, `vm-passthrough-kata` and `vm-vgpu-kata`, which stored the list of labels for kata? Would that simplify the code in this file?

controllers/state_manager.go
```diff
@@ -1011,6 +1046,8 @@ func (n ClusterPolicyController) isStateEnabled(stateName string) bool {
 		return clusterPolicySpec.NodeStatusExporter.IsEnabled()
 	case "state-sandbox-device-plugin":
 		return n.sandboxEnabled && clusterPolicySpec.SandboxDevicePlugin.IsEnabled()
+	case "state-kata-device-plugin":
+		return n.sandboxEnabled && clusterPolicySpec.SandboxWorkloads.Mode == "kata"
```
cdesiniotis:
```suggestion
		return n.sandboxEnabled && clusterPolicySpec.KataSandboxDevicePlugin.IsEnabled() && clusterPolicySpec.SandboxWorkloads.Mode == "kata"
```

controllers/state_manager.go
```diff
@@ -1011,6 +1046,8 @@ func (n ClusterPolicyController) isStateEnabled(stateName string) bool {
 		return clusterPolicySpec.NodeStatusExporter.IsEnabled()
 	case "state-sandbox-device-plugin":
 		return n.sandboxEnabled && clusterPolicySpec.SandboxDevicePlugin.IsEnabled()
+	case "state-kata-device-plugin":
```
cdesiniotis, line 1100: Also update the case for `state-sandbox-device-plugin` to only return true when `clusterPolicySoec.SandboxWorkloads.Mode == kubevirt`

controllers/state_manager.go
cdesiniotis: nit: what about also adding a new type for `SandboxWorkloadsMode` and corresponding constants for this type in `api/nvidia/v1/clusterpolicy.go` similar to: https://github.com/NVIDIA/gpu-operator/blob/69ba9c34a97a0275a6f9526a44028417ee937eab/api/nvidia/v1/clusterpolicy_types.go#L100-L112

controllers/state_manager.go
jojimt: vm-passthrough is the only one we are testing. When we have vm-gpu, we could enable this.

controllers/state_manager.go
jojimt: I don't think it simplifies the code overall. It just moves the complexity elsewhere and probably makes for less readable code.

controllers/state_manager.go
jojimt: Simplified on similar lines, without a copy.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -124,6 +126,16 @@ func (r Runtime) String() string {
 	}
 }
 
+// SandboxWorkloadsMode defines the mode for sandbox workloads
+type SandboxWorkloadsMode string
+
+const (
+	// kubevirt mode
+	KubeVirt SandboxWorkloadsMode = "kubevirt"
+	// kata more
```
```suggestion
	// Kata is the SandboxWorkloadsMode value for enabling Kata Container based workloads.
```

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -124,6 +126,16 @@ func (r Runtime) String() string {
 	}
 }
 
+// SandboxWorkloadsMode defines the mode for sandbox workloads
+type SandboxWorkloadsMode string
+
+const (
+	// kubevirt mode
```
```suggestion
	// KubeVirt is the SandboxWorkloadsMode value for enabling KubeVirt based workloads.
```

tariq1890: This adheres to the `golint` standards for godoc

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -197,6 +209,10 @@ type SandboxWorkloadsSpec struct {
 	// +kubebuilder:validation:Enum=container;vm-passthrough;vm-vgpu
 	// +kubebuilder:default=container
 	DefaultWorkload string `json:"defaultWorkload,omitempty"`
+	// Mode indicates the sandbox mode (e.g. kubevirt vs kata)
```
```suggestion
	// Mode indicates the sandbox mode. Accepted values are "kubevirt" and "kata". The default value is "kubevirt".
```

controllers/state_manager.go
```diff
@@ -117,6 +119,7 @@ var gpuNodeLabels = map[string]string{
 
 type gpuWorkloadConfiguration struct {
 	config string
+	mode   string // SandboxWorkloads.Mode (e.g. "kubevirt", "kata") — only affects vm-passthrough labels
```
tariq1890: In general I'd like to avoid comments as much as possible as the code itself should be self-explanatory. With that in mind, can we do this?
```suggestion
	sandboxMode   string // SandboxWorkloads.Mode (e.g. "kubevirt", "kata") — only affects vm-passthrough labels
```

controllers/state_manager.go
```diff
@@ -1010,7 +1044,9 @@ func (n ClusterPolicyController) isStateEnabled(stateName string) bool {
 	case "state-node-status-exporter":
 		return clusterPolicySpec.NodeStatusExporter.IsEnabled()
 	case "state-sandbox-device-plugin":
-		return n.sandboxEnabled && clusterPolicySpec.SandboxDevicePlugin.IsEnabled()
+		return n.sandboxEnabled && clusterPolicySpec.SandboxDevicePlugin.IsEnabled() && clusterPolicySpec.SandboxWorkloads.Mode == "kubevirt"
+	case "state-kata-device-plugin":
+		return n.sandboxEnabled && clusterPolicySpec.SandboxWorkloads.Mode == "kata"
```
tariq1890: Don't we also need a `KataSandboxDevicePlugin.IsEnabled()` check?

deployments/gpu-operator/values.yaml
```diff
@@ -508,6 +510,17 @@ sandboxDevicePlugin:
   env: []
   resources: {}
 
+# Kata sandbox device plugin (used when sandboxWorkloads.mode is "kata"). No enabled flag — implied by mode.
+kataSandboxDevicePlugin:
+  repository: nvcr.io/nvidia
+  image: nvidia-sandbox-device-plugin
```
cdesiniotis, line 517: We also need to add this image to the OLM bundle. Specifically [here](https://github.com/NVIDIA/gpu-operator/blob/11e2596aab0143d8bdee7101adbb26f453de1ff1/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L198) and [here](https://github.com/NVIDIA/gpu-operator/blob/11e2596aab0143d8bdee7101adbb26f453de1ff1/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L907).

As an aside, I noticed we are missing the CC Manager image in the OLM bundle. We may want to include that as well in a follow-up.

assets/state-kata-device-plugin/0500_daemonset.yaml
```diff
@@ -0,0 +1,82 @@
...
+  selector:
+    matchLabels:
+      app: nvidia-kata-sandbox-device-plugin-daemonset
+  template:
+    metadata:
+      labels:
+        app: nvidia-kata-sandbox-device-plugin-daemonset
+    spec:
+      nodeSelector:
+        nvidia.com/gpu.deploy.kata-device-plugin: "true"
```
cdesiniotis: nit: I would prefer if the component label aligned with the name of the component elsewhere.
```suggestion
        nvidia.com/gpu.deploy.kata-sandbox-device-plugin: "true"
```

assets/state-kata-device-plugin/0500_daemonset.yaml
```diff
@@ -0,0 +1,82 @@
...
+      tolerations:
+        - key: nvidia.com/gpu
+          operator: Exists
+          effect: NoSchedule
+      priorityClassName: system-node-critical
+      serviceAccountName: nvidia-kata-sandbox-device-plugin
+      initContainers:
+        - name: kata-validation
```
cdesiniotis: nit: let's align the name of this container with the name used in the `sandbox-device-plugin` daemonset. In general, the name of the validation init containers should be `<component>-validation`, where `<component>` indicates what component we are validating. For reference, see https://github.com/NVIDIA/gpu-operator/blob/11e2596aab0143d8bdee7101adbb26f453de1ff1/assets/state-sandbox-device-plugin/0500_daemonset.yaml#L28 and https://github.com/NVIDIA/gpu-operator/blob/11e2596aab0143d8bdee7101adbb26f453de1ff1/controllers/object_controls.go#L3663-L3707


```suggestion
        - name: vfio-pci-validation
```

controllers/state_manager.go
cdesiniotis: +1

controllers/state_manager.go
cdesiniotis: +1 let's add an enabled field and an additional check here

tariq1890:
Let's update the sample clusterpolicy JSON [here](https://github.com/NVIDIA/gpu-operator/blob/75f47b38729ba8a920d20e496f13a9453f60a8d3/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L24) to reflect the latest clusterpolicy API updates

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -225,6 +229,8 @@ spec:
       image: ghcr.io/nvidia/k8s-driver-manager:69eec274
     - name: sandbox-device-plugin-image
       image: nvcr.io/nvidia/kubevirt-gpu-device-plugin:v1.4.0@sha256:119de9a331a47203858b99901f44d0c4a8052961b4e60327f4b100d0ab8c9df0
+    - name: kata-sandbox-device-plugin-image
+      image: ghcr.io/nvidia/nvidia-sandbox-device-plugin:3a8894b2
```
tariq1890, line 233: I don't see this image being used in `values.yaml`.  We should use the same image in both helm and OLM for consistency's sake.  

Also, are we okay with pinning a `ghcr.io` image in our releases?

cdesiniotis: It is fine to point to `ghcr.io` for now as long as we update this to refer to a released image on `nvcr.io` before our GPU Operator release. 

But agree, let's have both the OLM bundle and helm chart point to the same image on `ghcr.io` in this PR.

jojimt: This is the correct image, I have updated helm. We can switch to release version before the release.

controllers/state_manager.go
```diff
@@ -340,6 +343,25 @@ func getWorkloadConfig(labels map[string]string, sandboxEnabled bool) (string, e
...
+	}
+
+	if config == gpuWorkloadConfigVMPassthrough && gpuv1.SandboxWorkloadsMode(mode) == gpuv1.Kata {
+		delete(labels, kubevirtDevicePluginDeployLabelKey)
+		labels[kataDevicePluginDeployLabelKey] = "true"
+		return labels
+	}
+
+	delete(labels, kataDevicePluginDeployLabelKey)
+	labels[kubevirtDevicePluginDeployLabelKey] = "true"
```
cdesiniotis: Is this really necessary? IIUC the original `labels` fetched from the `gpuStateLabels` map will never contain `kataDevicePluginDeployLabelKey` as a key.


```suggestion
```

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -38,6 +38,7 @@ metadata:
             "sandboxWorkloads": {
               "enabled": false,
               "defaultWorkload": "container"
```
tariq1890: Missing a comma at the end of L40

PR #2144, opened 2026-02-20
eliminate redundant node list calls when retrieving the GPU Node OS Info

tariq1890:
The commits dcb3f478620ba95988eb1dc22e9f53192db09632 and efaac78606e40e32e6eedbfed365a6767ab01c45 have
removed the dependency on the host-mounted /etc/os-release file and replaced it with NFD labels as the source of information. This commit builds upon those changes by eliminating the redundant calls to Node List
operations. This makes the gpu-operator scale better when it applies transformations especially in large clusters.

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -662,67 +662,6 @@ func (n ClusterPolicyController) getKernelVersionsMap() (map[string]string, erro
 	return kernelVersionMap, nil
 }
 
-func kernelFullVersion(n ClusterPolicyController) (string, string, string) {
-	ctx := n.ctx
-	logger := n.logger.WithValues("Request.Namespace", "default", "Request.Name", "Node")
-	// We need the node labels to fetch the correct container
-	opts := []client.ListOption{
-		client.MatchingLabels{"nvidia.com/gpu.present": "true"},
-	}
-
-	list := &corev1.NodeList{}
-	err := n.client.List(ctx, list, opts...)
-	if err != nil {
-		logger.Info("Could not get NodeList", "ERROR", err)
-		return "", "", ""
-	}
-
-	if len(list.Items) == 0 {
-		// none of the nodes matched nvidia GPU label
-		// either the nodes do not have GPUs, or NFD is not running
-		logger.Info("Could not get any nodes to match nvidia.com/gpu.present label", "ERROR", "")
-		return "", "", ""
-	}
-
-	// Assuming all nodes are running the same kernel version,
-	// One could easily add driver-kernel-versions for each node.
-	node := list.Items[0]
-	labels := node.GetLabels()
-
-	var ok bool
-	kFVersion, ok := labels[nfdKernelLabelKey]
-	if ok {
-		logger.Info(kFVersion)
-	} else {
-		err := apierrors.NewNotFound(schema.GroupResource{Group: "Node", Resource: "Label"}, nfdKernelLabelKey)
-		logger.Info("Couldn't get kernelVersion, did you run the node feature discovery?", "Error", err)
-		return "", "", ""
-	}
-
-	osName, ok := labels[nfdOSReleaseIDLabelKey]
-	if !ok {
-		return kFVersion, "", ""
-	}
-	osVersion, ok := labels[nfdOSVersionIDLabelKey]
-	if !ok {
-		return kFVersion, "", ""
-	}
-	osMajorVersion := strings.Split(osVersion, ".")[0]
-	osMajorNumber, err := strconv.Atoi(osMajorVersion)
-	if err != nil {
-		return kFVersion, "", ""
-	}
-
-	// If the OS is RockyLinux or RHEL 10 & above, we will omit the minor version when constructing the os image tag
-	if osName == "rocky" || (osName == "rhel" && osMajorNumber >= 10) {
-		osVersion = osMajorVersion
-	}
-
-	osTag := fmt.Sprintf("%s%s", osName, osVersion)
-
-	return kFVersion, osTag, osVersion
-}
-
 func preprocessService(obj *corev1.Service, n ClusterPolicyController) error {
 	logger := n.logger.WithValues("Service", obj.Name)
 	transformations := map[string]func(*corev1.Service, *gpuv1.ClusterPolicySpec) error{
@@ -3206,12 +3145,6 @@ func transformOpenShiftDriverToolkitContainer(obj *appsv1.DaemonSet, config *gpu
 
 // resolveDriverTag resolves image tag based on the OS of the worker node
 func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string, error) {
-	// obtain os version
-	kvers, osTag, _ := kernelFullVersion(n)
-	if kvers == "" {
-		return "", fmt.Errorf("ERROR: Could not find kernel full version: ('%s', '%s')", kvers, osTag)
-	}
-
 	// obtain image path
 	var image string
 	var err error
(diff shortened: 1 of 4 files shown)


controllers/object_controls.go
```diff
@@ -662,7 +662,7 @@ func (n ClusterPolicyController) getKernelVersionsMap() (map[string]string, erro
 	return kernelVersionMap, nil
 }
 
-func kernelFullVersion(n ClusterPolicyController) (string, string, string) {
```
tariq1890, line 665: After checking all the callers of this method, I saw that both the first and last return arguments aren't really used. I believe it is safe to remove them

controllers/object_controls.go
```diff
@@ -662,7 +662,7 @@ func (n ClusterPolicyController) getKernelVersionsMap() (map[string]string, erro
...
+func (n ClusterPolicyController) getGPUNodeOSTag() string {
 	ctx := n.ctx
 	logger := n.logger.WithValues("Request.Namespace", "default", "Request.Name", "Node")
 	// We need the node labels to fetch the correct container
```
karthikvetrivel: nit: I think this comment is stale now and can be removed.

controllers/object_controls.go
```diff
@@ -3267,54 +3265,38 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
-	osID, osTag, err := n.getGPUNodeOSID()
-	if err != nil {
-		return "", err
-	}
+func getRepoConfigPath(osID string) (string, error) {
 	if path, ok := RepoConfigPathMap[osID]; ok {
 		return path, nil
 	}
-	return "", fmt.Errorf("repository configuration not supported for distribution %s", osTag)
+	return "", fmt.Errorf("repository configuration not supported for distribution %s", osID)
```
karthikvetrivel: Question: I see that we're now longer passing the full `osTag` anymore, just an `osID`.  Is that fine?

rahulait: Yes, because we have osID as key here and not the full version/osTag: https://github.com/NVIDIA/gpu-operator/blob/main/controllers/object_controls.go#L205-L307
And these paths are consistent across different versions of same os. Like they remain same for ubuntu22.04, ubuntu20.04, ubuntu24.04 ...

controllers/object_controls.go
tariq1890: removed

controllers/object_controls.go
```diff
@@ -662,55 +662,44 @@ func (n ClusterPolicyController) getKernelVersionsMap() (map[string]string, erro
...
 }
 
-func kernelFullVersion(n ClusterPolicyController) (string, string, string) {
+func (n ClusterPolicyController) getGPUNodeOSTag() string {
 	ctx := n.ctx
 	logger := n.logger.WithValues("Request.Namespace", "default", "Request.Name", "Node")
-	// We need the node labels to fetch the correct container
+
 	opts := []client.ListOption{
 		client.MatchingLabels{"nvidia.com/gpu.present": "true"},
```
rajathagasthya: optional: you can also add `client.Limit(1)` as a list option since we only ever look at one node and get another micro optimization (reduced memory allocations).

tariq1890: Great suggestion, thanks!

tariq1890: Done

rajathagasthya (approved):

tariq1890:
I have validated this PR on OpenShift

controllers/object_controls.go
```diff
@@ -1028,26 +1013,31 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 		return err
 	}
 
+	osTag := n.getGPUNodeOSTag()
```
cdesiniotis: As an alternative, what if we fetched this information (GPU Node os tag) before transforming / rendering any daemonsets? For example, we could fetch this information in `init()` after labeling all GPU nodes: https://github.com/NVIDIA/gpu-operator/blob/69ba9c34a97a0275a6f9526a44028417ee937eab/controllers/state_manager.go#L855. We could save the GPU Node os tag in a field of the ClusterPolicyController struct and use it as needed in the transformation functions. That way, we only make one `List` call and we don't have to change any of the function signatures for the transform functions in this file.

tariq1890: That makes sense. Let's see what I can do

tariq1890: I've addressed this. Please check now

controllers/state_manager.go
```diff
@@ -499,6 +501,30 @@ func (n *ClusterPolicyController) labelGPUNodes() (bool, int, error) {
...
+				return false, 0, fmt.Errorf("unable to retrieve OS name from label %s", nfdOSReleaseIDLabelKey)
+			}
+			osVersion, ok := labels[nfdOSVersionIDLabelKey]
+			if !ok {
+				return false, 0, fmt.Errorf("unable to retrieve OS version from label %s", nfdOSVersionIDLabelKey)
+			}
+			osMajorVersion := strings.Split(osVersion, ".")[0]
+			osMajorNumber, err := strconv.Atoi(osMajorVersion)
+			if err != nil {
+				return false, 0, fmt.Errorf("error processing OS major version %s, err %w", osMajorVersion, err)
```
cdesiniotis:
```suggestion
				return false, 0, fmt.Errorf("error processing OS major version %s: %w", osMajorVersion, err)
```

controllers/state_manager.go
```diff
@@ -499,6 +501,30 @@ func (n *ClusterPolicyController) labelGPUNodes() (bool, int, error) {
 		if !clusterHasNFDLabels {
 			clusterHasNFDLabels = hasNFDLabels(labels)
 		}
+
+		if len(n.osTag) == 0 {
```
cdesiniotis: This for loop iterates over *all* nodes. We need to make sure the os tag we construct is coming from a GPU node. As currently implemented, this is not guaranteed to be the case.

I understand why you moved this logic to this method -- we list all nodes in this method, and therefore, could easily construct the os tag here. However, for readability sake (and to not overload this method), what about moving this code in a separate method that gets invoked *after* `labelGPUNodes()` gets invoked in `init()`?

controllers/state_manager.go
```diff
@@ -161,6 +162,7 @@ type ClusterPolicyController struct {
 	ocpDriverToolkit OpenShiftDriverToolkit
 
 	runtime        gpuv1.Runtime
+	osTag          string
```
cdesiniotis: Question -- thoughts on renaming this field to `gpuNodeOSTag` so that it is clear this OS tag is constructed from a GPU node?

tariq1890: Yes, agreed. That's a better name

controllers/state_manager.go
tariq1890: Ok so that would mean extracting the node List call out of the `labelGPUNodes()` method right?

cdesiniotis: To minimize the diff here, I was thinking of not extracting the node List call out of `labelGPUNodes()`. It would mean having to make an additional List call in the `getGPUNodeOSTag()` helper (like you originally had). I think that is fine for now.

cdesiniotis: The second List call can filter for just GPU nodes.

controllers/state_manager.go
tariq1890: This is done

controllers/state_manager.go
tariq1890: This is done

controllers/state_manager.go
tariq1890: > To minimize the diff here, I was thinking of not extracting the node List call out of labelGPUNodes().

~I've extracted the node List call out of `labelGPUNodes()`. Figured it's the best optimisation we have. Please check now~

tariq1890: I've implemented your suggestion. Please check now

cdesiniotis (approved):
Thanks @tariq1890 this looks great now!

controllers/state_manager.go
cdesiniotis: Thanks, this is much more readable now.

PR #2162, opened 2026-02-24
Mark defaultRuntime as optional, remove validation constraints

empovit:
Commit cf964b8c marked `defaultRuntime` as deprecated, removed it from alm-examples in the OpenShift operator. However, the validations on this field remained in the CRD. As a result,  trying to create a default `ClusterPolicy` based on alm-examples failed with the following error:
    
```
spec.operator.defaultRuntime: Unsupported value: "": supported values:
"docker", "crio", "containerd"
```
    
This fixes the issue by making the field optional and removing the validations.

## Checklist

## Testing

No code changes to affect unit-tests. The fix was tested manually with a custom bundle image.

### DIFF
--- a/api/nvidia/v1/clusterpolicy_types.go
+++ b/api/nvidia/v1/clusterpolicy_types.go
@@ -139,9 +139,8 @@ const (
 // OperatorSpec describes configuration options for the operator
 type OperatorSpec struct {
 	// Deprecated: DefaultRuntime is no longer used by the gpu-operator. This is instead, detected at runtime.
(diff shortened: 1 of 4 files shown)


rahulait:
This now becomes a breaking change as the field behavior is changed. We'll need to evaluate if we want to do this or add back the optional field we removed in clusterpolicy json for now.

Update: it looks like we were not using this field previously. So any value set was not getting used. Since the field was not used, this change doesn't break any functionality and is not a breaking change then. Its safe to accept this change then.

rahulait (approved):

PR #2167, opened 2026-02-25
bump toolkit to v1.19.0-rc.5

rahulait:
## Description

See: https://github.com/NVIDIA/nvidia-container-toolkit/releases/tag/v1.19.0-rc.5

## Checklist


Makefile
rajathagasthya: We can remove `-extldflags=-Wl,-z,lazy` as they are no longer relevant for non-CGO builds (I believe they are ignored).

rahulait: Good catch, updated.

PR #2177, opened 2026-02-27
Make kataManager.config conditional in ClusterPolicy template

rajathagasthya:
## Description

Default `kataManager.config` is empty and results in a `null` value which ClusterPolicy rejects.

## Checklist

### DIFF
--- a/deployments/gpu-operator/templates/clusterpolicy.yaml
+++ b/deployments/gpu-operator/templates/clusterpolicy.yaml
@@ -300,7 +300,9 @@ spec:
       {{- end }}
   kataManager:
     enabled: {{ .Values.kataManager.enabled }}
+    {{- if .Values.kataManager.config }}
     config: {{ toYaml .Values.kataManager.config | nindent 6 }}
+    {{- end }}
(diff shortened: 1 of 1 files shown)


deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -300,7 +300,9 @@ spec:
       {{- end }}
   kataManager:
     enabled: {{ .Values.kataManager.enabled }}
+    {{- if .Values.kataManager.config }}
```
tariq1890, line 303:
```suggestion
    {{- if and .Values.kataManager.enabled .Values.kataManager.config }}
```

rajathagasthya: None of the other keys like `image`, `version`, etc. are conditionally filled on `enabled` for any operand. Is it necessary to special case the `config` key? It will be ignored if it's not enabled anyway.

tariq1890: Fair point. Will resolve

Issue #772, opened 2024-06-15
"leader election lost" gpu-operator pod restarts


alnhk:
### 1. Quick Debug Information
We deployed 24.3.0 helm chart gpu-opeator on the kubernetes
* RHEL 8.9
* Kernel Version:  4.18.0-513.18.1.el8_9.x86_64
* Container Runtime Type : CRI-O 1.27.4
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS): k8s
* GPU Operator Version: 24.3.0 

### 2. Issue or feature description
The pod "gpu-operator" crashes with below complete log 
``` 
{"level":"info","ts":1718417750.9473174,"logger":"controllers.Upgrade","msg":"Upgrades in progress","currently in progress":0,"max parallel upgrades":1,"upgrade slots available":1,"currently unavailable nodes":0,"total number of nodes":2,"maximum nodes that can be unavailable":1}
{"level":"info","ts":1718417750.9473245,"logger":"controllers.Upgrade","msg":"ProcessDoneOrUnknownNodes"}
{"level":"info","ts":1718417750.9473295,"logger":"controllers.Upgrade","msg":"ProcessDoneOrUnknownNodes"}
{"level":"info","ts":1718417750.9629183,"logger":"controllers.Upgrade","msg":"ProcessUpgradeRequiredNodes"}
{"level":"info","ts":1718417750.9629385,"logger":"controllers.Upgrade","msg":"ProcessCordonRequiredNodes"}
{"level":"info","ts":1718417750.9629438,"logger":"controllers.Upgrade","msg":"ProcessWaitForJobsRequiredNodes"}
{"level":"info","ts":1718417750.9629483,"logger":"controllers.Upgrade","msg":"ProcessPodDeletionRequiredNodes"}
{"level":"info","ts":1718417750.9629524,"logger":"controllers.Upgrade","msg":"ProcessDrainNodes"}
{"level":"info","ts":1718417750.9629562,"logger":"controllers.Upgrade","msg":"Node drain is disabled by policy, skipping this step"}
{"level":"info","ts":1718417750.962961,"logger":"controllers.Upgrade","msg":"ProcessPodRestartNodes"}
{"level":"info","ts":1718417750.9629655,"logger":"controllers.Upgrade","msg":"Starting Pod Delete"}
{"level":"info","ts":1718417750.9629693,"logger":"controllers.Upgrade","msg":"No pods scheduled to restart"}
[...]

PR #2185, opened 2026-03-02
Add leader election configuration to operator deployment

yuyue9284:
## Description

Add config to operator's leader election, for single replica, disable the leader election can prevent the unwanted crash of gpu-operator due to transient error.

https://github.com/NVIDIA/gpu-operator/issues/772

## Checklist

### DIFF
--- a/deployments/gpu-operator/templates/operator.yaml
+++ b/deployments/gpu-operator/templates/operator.yaml
@@ -39,7 +39,9 @@ spec:
         imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
         command: ["gpu-operator"]
         args:
+        {{- if .Values.operator.leaderElection }}
         - --leader-elect
+        {{- end }}
       {{- if .Values.operator.logging.develMode }}
         - --zap-devel
       {{- else }}
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -74,6 +74,8 @@ operator:
   priorityClassName: system-node-critical
   runtimeClass: nvidia
   use_ocp_driver_toolkit: false
+  # Enable leader election
+  leaderElection: true
   # cleanup CRD on chart un-install
   cleanupCRD: false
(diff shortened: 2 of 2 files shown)


rajathagasthya:
@yuyue9284 We've changed our CI workflows, so could you rebase your PR? Thanks.

yuyue9284:
> @yuyue9284 We've changed our CI workflows, so could you rebase your PR? Thanks.

Thanks, updated.

tariq1890:
Is there a good reason to actually disable leader election here? @yuyue9284 The issue you've linked is doesn't seem like the failed leader election is the problem, but a rather a symptom of a failing API server. I am open to changing my view on this, but I am not convinced that we need a setting for this right now.

yuyue9284:
> Is there a good reason to actually disable leader election here? @yuyue9284 The issue you've linked is doesn't seem like the failed leader election is the problem, but a rather a symptom of a failing API server. I am open to changing my view on this, but I am not convinced that we need a setting for this right now.

Hi @tariq1890, we run gpu-operator on a large cluster, during the other system components update, the API server may face the temporally performance degrade during that time, and as a result the gpu-operator might fail to renew the lease and crash/re-list at that time, make the API server perf worse, since we are only running one instance of operator, we want to disable the leader election to avoid restart at that code path, but it is not supported by the chart..

rajathagasthya:
> as a result the gpu-operator might fail to renew the lease and crash/re-list at that time, make the API server perf worse

If API server isn't healthy for leader election, GPU Operator operations will fail as well. Plus I'm fairly sure other code paths during manager creation in controller-runtime fail and causes a restart/re-list. I'm not convinced one fewer leader election moves the needle.

tariq1890:
I agree with @rajathagasthya, if the leader election process fails, that is a symptom of a deeper issue that needs to be addressed first. The way I see it, if the API server issue which is at the root is addressed, the leader election should resolve itself and unblock the gpu-operator controller in a restart.

I am closing this PR for now.

PR #2187, opened 2026-03-04
Bump go version to 1.26.0

cdesiniotis:
Needed by https://github.com/NVIDIA/gpu-operator/pull/2180


tariq1890 (dismissed):

tariq1890:
Can you update the [Dockerfile](https://github.com/NVIDIA/gpu-operator/blob/main/docker/Dockerfile#L17) too?

cdesiniotis:
> Can you update the [Dockerfile](https://github.com/NVIDIA/gpu-operator/blob/main/docker/Dockerfile#L17) too?

Good catch. Updated. Ideally the Go version in the Dockerfile would be parameterized so that we only have to update the version in one place.

PR #2195, opened 2026-03-06
avoid unicode character that messes up pod rendering

rajatchopra:
## Description

When using the '<' character, the script is rendered like this:
`if [ \"$(\u003c/run/nvidia/validations/workload-type)\" != \"vm-passthrough\" ]; then echo kata not needed, skipping validation; exit 0; fi;`

This PR changes the `<` to `cat` so that the init check is functional.

## Checklist

## Testing

Manual out of band testing.

### DIFF
--- a/assets/state-kata-device-plugin/0500_daemonset.yaml
+++ b/assets/state-kata-device-plugin/0500_daemonset.yaml
@@ -30,7 +30,7 @@ spec:
           command: ['sh', '-c']
           args:
             - until [ -f /run/nvidia/validations/workload-type ]; do echo waiting for workload type status file; sleep 5; done;
-              if [ "$(</run/nvidia/validations/workload-type)" != "vm-passthrough" ]; then echo kata not needed, skipping validation; exit 0; fi;
+              if [ "$(cat /run/nvidia/validations/workload-type)" != "vm-passthrough" ]; then echo kata not needed, skipping validation; exit 0; fi;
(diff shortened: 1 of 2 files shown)


rajatchopra:
cc @jojimt

rajathagasthya (approved):

cdesiniotis:
What about Line 49? Why is this issue only being discovered now? We have the same entrypoint scripts in the kubevirt-gpu-device-plugin and no one has reported issues. https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-sandbox-device-plugin/0500_daemonset.yaml

cdesiniotis:
Also note, based on the PR description I believe you intended to update https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-kata-device-plugin/0500_daemonset.yaml?

rajatchopra:
> Also note, based on the PR description I believe you intended to update https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-kata-device-plugin/0500_daemonset.yaml?

Correct. Fixed in that file.

rajatchopra:
> What about Line 49? Why is this issue only being discovered now? We have the same entrypoint scripts in the kubevirt-gpu-device-plugin and no one has reported issues. https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-sandbox-device-plugin/0500_daemonset.yaml

The issue was always there. Never caught it because it only prevents a race condition. The bug was just not preventing the race condition - which finally happened with QA testing.

cdesiniotis (approved):

PR #2202, opened 2026-03-11
[WIP] : remove unnecessary permissions, add validating-admissions-policies

rahulait:
## Description
This is still work in progress. PR will be updated as we get more feedback.

Key changes include:

1. Adding ValidatingAdmissionPolicy for gpu-operator service account
    a. Restrict to certain runtime handlers when creating a runtime class
    b. Restrict to certain node labels when labeling nodes
    c. Restrict to certain labels that can be added to namespaces
2. Remove privileges which are not required

Openshift specific OLM bundle generation still needs to be updated in the PR

## Checklist


deployments/gpu-operator/templates/validating-admission-policies.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - v1
+      operations:
+      - UPDATE
+      resources:
+      - namespaces
+  matchConditions:
+  - name: gpu-operator-service-account
+    expression: request.userInfo.username == 'system:serviceaccount:{{ .Release.Namespace }}:gpu-operator'
+  - name: target-namespace
+    expression: object.metadata.name in ['{{ .Release.Namespace }}', 'nvidia-gpu-operator']
```
rahulait, line 65: need inputs here if these are the only namespaces we label or there can be others as well in case of openshift

deployments/gpu-operator/templates/validating-admission-policies.yaml
```diff
@@ -0,0 +1,141 @@
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy" }}
```
rahulait, line 1: This can be further restricted to be installed if someone requests validatingadmissionpolicy guardrails in values.yaml. I haven't added added that and looking for feedback if we should do something like that or let these be always installed if supported by k8s.

PR #2224, opened 2026-03-18
Bump project version v26.3.0

cdesiniotis:


rahulait:
Do we need to bump version here as well? 
https://github.com/NVIDIA/gpu-operator/blob/main/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L21
https://github.com/NVIDIA/gpu-operator/blob/main/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L201

cdesiniotis:
> https://github.com/NVIDIA/gpu-operator/blob/main/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L201

Yes, good catch! I have updated this.

Issue #2239, opened 2026-03-21
[Bug]: Latest relese does not work on Talos Linux because of incorrect version parsing


finomen:
**Describe the bug**
Latest release fails to list gpus because it fails to process OS version:
{"level":"error","ts":1774112196.4380157,"msg":"Reconciler error","controller":"clusterpolicy-controller","object":{"name":"cluster-policy"},"namespace":"","name":"cluster-policy","reconcileID":"991b62e2-34f1-4844-9ed3-fe049d7a5c04","error":"failed to retrieve GPU node OS tag: error processing OS major version v1: strconv.Atoi: parsing \"v1\": invalid syntax"}

Annotations on node are the following:
"feature.node.kubernetes.io/system-os_release.ID": "talos",
"feature.node.kubernetes.io/system-os_release.VERSION_ID": "v1.12.6",

Operator 25.10.1 works fine. Likely, it was broken in https://github.com/NVIDIA/gpu-operator/commit/3fb6dd3095e9ec109597a541252f19d957a4bef0

**To Reproduce**
Install on Talos linux

**Expected behavior**
OPerator works

**Environment (please provide the following information):**
 - GPU Operator Version: 26.3.0
 - OS: Talos 1.12.6
 - Kernel Version: 6.18.18-talos
 - Container Runtime Version: 2.1.6
 - Kubernetes Distro and Version: v1.34.0

rahulait:
Thanks @finomen for filing this bug. Does the following image works fine for you? I've provided sample values.yaml change to pull this fix until we release a newer patched version.

```
operator:
  repository: ghcr.io/nvidia
  image: gpu-operator
  version: 670be908
```

gimmw:
Thanks @rahulait, I can confirm that image resolved the same issue I had (talos 1.10.6)

ch9hn:
We checked have the same issue with the above provided Image and Talos 1.12.6 even when we define the new image.

`failed to retrieve GPU node OS tag: unable to retrieve OS name from label feature.node.kubernetes.io/system-os_release.ID`

```
    helm:
      values: |
        operator:
          repository: ghcr.io/nvidia
          image: gpu-operator
          version: 670be908
          defaultRuntime: containerd
```

rahulait:
@ch9hn can you share what node labels you have on the node? You can also specify the version as "main-latest" as the fix has been merged to main. I would like to see what labels your node has and why its failing there.

ch9hn:
[...]

PR #2244, opened 2026-03-24
store os release name and use it than recalculating from tag which is errorprone

rahulait:
## Description
Fixes https://github.com/NVIDIA/gpu-operator/issues/2239

Instead of re-calculating os release from the tag (which is errorprone as there can be different combinations and all might not be covered by code), store the os-release when its fetched from node label and then re-use it where required.

## Checklist

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -3298,16 +3298,9 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
 	return image, nil
 }
 
-func getOSName(osTag string) string {
-	// Extract base OS ID by stripping version suffix from osTag
-	// Examples: "rhel10" -> "rhel", "ubuntu22.04" -> "ubuntu", "rocky9" -> "rocky"
-	osID := strings.TrimRight(osTag, "0123456789.")
-	return osID
-}
-
 // getRepoConfigPath returns the standard OS specific path for repository configuration files.
 func (n ClusterPolicyController) getRepoConfigPath() (string, error) {
-	osID := getOSName(n.gpuNodeOSTag)
+	osID := n.gpuNodeOSRelease
 	if path, ok := RepoConfigPathMap[osID]; ok {
 		return path, nil
 	}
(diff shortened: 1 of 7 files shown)


controllers/object_controls.go
```diff
@@ -3298,16 +3298,12 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
...
-	// Examples: "rhel10" -> "rhel", "ubuntu22.04" -> "ubuntu", "rocky9" -> "rocky"
-	osID := strings.TrimRight(osTag, "0123456789.")
-	return osID
-}
-
 // getRepoConfigPath returns the standard OS specific path for repository configuration files.
 func (n ClusterPolicyController) getRepoConfigPath() (string, error) {
-	osID := getOSName(n.gpuNodeOSTag)
+	osID := n.gpuNodeOSRelease
+	if osID == "" {
```
tariq1890: Can we not throw this error at the time of setting this property? There is a never a working scenario where this property would be empty, right?

rahulait: We only sets this field when hasGPUNodes is [true](https://github.com/NVIDIA/gpu-operator/blob/652724dd65926e85ea411da0f1525467439a2a8d/controllers/state_manager.go#L941), so there is a valid reconcile path where the property stays empty (for example, no GPU nodes yet).

controllers/object_controls_test.go
```diff
@@ -225,13 +225,14 @@ func setup() error {
 	if gpuNodeCount == 0 {
 		return fmt.Errorf("no gpu nodes in mock cluster")
 	}
-	gpuNodeOSTag, err := clusterPolicyController.getGPUNodeOSTag()
+	gpuNodeOSRelease, gpuNodeOSTag, err := clusterPolicyController.getGPUNodeOSInfo()
```
tariq1890, line 228: @rahulait Can we not throw an error here if the gpuNodeOSRelease and gpuNodeOSTag is empty? This code path is only executed after gpu nodes are detected. This would eliminate some redundant code

rahulait: Updated.

tariq1890:
Looks like the unit tests are failing

controllers/object_controls.go
```diff
@@ -3618,7 +3611,6 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 		}
 	}
 
-	osID = getOSName(n.gpuNodeOSTag)
```
tariq1890, line 3621: Instead of removing this line, we want it to be this instead

```
osID := n.gpuNodeOSRelease
```

rahulait: Its set on line 3551.

rahulait:
> Looks like the unit tests are failing

Should be fixed now.

Issue #363, opened 2022-06-28
Prometheus service discovery can't find DCGM exporter


TarekMSayed:
Prometheus service discovery can't find DCGM exporter usually that needs to add extra labels/annotations to the pods ex: `release: "prometheus-operator"` or create service monitor with these labels/annotations.
I can't find a way to add it to the helm chart

shivamerla:
@TarekMSayed This has been missing with operator for a while now, will consider adding this with next release(setting pod labels/annotations for dcgm-exporter).

TarekMSayed:
@shivamerla Also adding an optional service monitor for the DCGM exporter metric service will be nice, I forked the helm chart and add one manually to the templates
dcgm-servicemonitor.yaml
```yaml
{{- if .Values.dcgmExporter.metrics.serviceMonitor.enabled -}}
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: {{ include "gpu-operator.fullname" . }}-dcgm-exporter 
{{- if .Values.dcgmExporter.metrics.serviceMonitor.namespace }}
  namespace: {{ .Values.dcgmExporter.metrics.serviceMonitor.namespace | quote }}
{{- end }}
  labels:
    {{- include "gpu-operator.labels" . | nindent 4 }}
    app: nvidia-dcgm-exporter
  {{- if .Values.dcgmExporter.metrics.serviceMonitor.additionalLabels }}
    {{- toYaml .Values.dcgmExporter.metrics.serviceMonitor.additionalLabels | nindent 4 }}
  {{- end }}
spec:
  endpoints:
    - port: gpu-metrics
      interval: {{ .Values.dcgmExporter.metrics.serviceMonitor.scrapeInterval }}
    {{- if .Values.dcgmExporter.metrics.serviceMonitor.honorLabels }}
      honorLabels: true
    {{- end }}
    {{- if .Values.dcgmExporter.metrics.serviceMonitor.metricRelabelings }}
      metricRelabelings: {{ toYaml .Values.dcgmExporter.metrics.serviceMonitor.metricRelabelings | nindent 8 }}
    {{- end }}
{{- if .Values.dcgmExporter.metrics.serviceMonitor.jobLabel }}
  jobLabel: {{ .Values.dcgmExporter.metrics.serviceMonitor.jobLabel | quote }}
{{- end }}
{{- if .Values.dcgmExporter.metrics.serviceMonitor.namespaceSelector }}
[...]

Issue #305, opened 2022-01-10
No ServiceMonitor set in Prometheus ? 


seb-835:
Hi,
i have install the gpu-operator, i see the dcgm-exporter running, i can pull metrics manually with curl.
But i do not see any ServiceMonitor set in prometheus, so there is no gpu metrics in my Dashboard.

Do i have to manually setup a Prometheus ServiceMonitor ?  or do i missed an option in the helm chart settings.
Thanks for help.

shivamerla:
@seb-835 Yes, currently this needs to be manually created(Only auto created for RedHat OCP). We will add this for K8s too with next release.

rhaps0dy:
You can now do `--set dcgmExporter.serviceMonitor.enabled=true` and the chart creates the ServiceMonitor for you

PR #2262, opened 2026-04-01
dcgm-exporter: enable ServiceMonitor by default and skip gracefully when Prometheus CRD is absent

VincentG1234:
### Problem

The DCGM Exporter `ServiceMonitor` has been opt-in (`enabled: false`) since the feature was introduced in 2022. Every user who wants GPU metrics scraped by Prometheus must remember to set `serviceMonitor.enabled: true` — **a silent misconfiguration that causes hours of debugging** (#305 #363) .

On top of that, when `enabled: true` is set and the Prometheus `ServiceMonitor` CRD is absent, the operator returns `NotReady` and blocks the entire reconcile loop (re-queuing every 5 seconds). This was documented in release 23.3 with added logging, but the underlying blocking behavior was never fixed. As a result, missing Prometheus CRDs prevent GFD pods from starting.

### Solution

Two minimal changes:

1. **Set `serviceMonitor.enabled: true` by default** in `values.yaml`.
2. **Fix the blocking behavior**: when the Prometheus `ServiceMonitor` CRD is absent, the operator now returns `Ready` and skips gracefully instead of blocking the reconcile loop. This aligns `state-dcgm-exporter` with the existing behavior of `state-operator-metrics`, which already handled this case correctly.

An explicit `enabled: false` continues to disable and remove the resource.

| Scenario | Before | After |
|---|---|---|
| `enabled: true` + CRD absent | `NotReady` (blocks reconcile loop, GFD stalls) | `Ready` (silent skip) |
| `enabled: true` + CRD present | created | created (unchanged) |
| `enabled: false` explicit | `Disabled` | `Disabled` (unchanged) |

### Changes

- **`deployments/gpu-operator/values.yaml`** — `serviceMonitor.enabled: false` → `enabled: true`.
- **`controllers/object_controls.go`** — `ServiceMonitor()`: CRD-absent path returns `Ready` instead of `NotReady`, consistent with `state-operator-metrics`.

### Testing

```bash
go test ./controllers/... -run TestServiceMonitor -v
```

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -4996,9 +4996,10 @@ func ServiceMonitor(n ClusterPolicyController) (gpuv1.State, error) {
 			return gpuv1.Disabled, nil
 		}
 
+		// If Prometheus CRD is missing, skip gracefully
 		if !serviceMonitorCRDExists {
-			logger.Error(fmt.Errorf("couldn't find ServiceMonitor CRD"), "Install Prometheus and necessary CRDs for gathering GPU metrics!")
-			return gpuv1.NotReady, nil
+			logger.V(1).Info("ServiceMonitor CRD not found, skipping DCGM Exporter ServiceMonitor creation")
+			return gpuv1.Ready, nil
 		}
 
 		// Apply custom edits for DCGM Exporter
--- a/controllers/object_controls_test.go
+++ b/controllers/object_controls_test.go
@@ -1325,7 +1325,7 @@ func TestServiceMonitor(t *testing.T) {
 			expectedServiceMonitor: nil,
 		},
 		{
-			description: "dcgm-exporter SM enabled, CRD missing -> NotReady",
+			description: "dcgm-exporter SM explicitly enabled, CRD missing -> Ready (skip gracefully)",
 			stateName:   "state-dcgm-exporter",
 			k8sObjects:  nil,
 			clusterPolicySpec: gpuv1.ClusterPolicySpec{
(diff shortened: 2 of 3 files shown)


VincentG1234:
Hi, just following up on this PR.

The main goal here is to fix an inconsistency in the operator:
- gpu-operator metrics ServiceMonitor is enabled by default
- dcgm-exporter ServiceMonitor is not

Since both rely on the same Prometheus Operator setup, this often leads to confusion and missing GPU metrics in practice, and significantly increases debugging time for users.

The change is backward-compatible:
- explicit disable still works
- CRD absence is handled gracefully (no blocking reconcile)

Happy to adjust the approach if you’d prefer a different direction (e.g. keeping default=false and only fixing the reconciliation behavior).

Would appreciate your feedback when you have time.

VincentG1234:
@rahulait hey, gentle ping on this PR, would truly appreciate your feedback when you have time
happy to adjust if needed

rahulait:
Hey @VincentG1234, thanks for this contribution. I'll plan to review this next week. Please feel free to ping again if you don't hear back by next weekend.

rahulait:
Thank Vincent, let me check with team and get back to you. In the meanwhile, can you please rebase your PR branch so that CI can run correctly, we added some fixes today.

VincentG1234:
just rebased, should be up to date now

rahulait:
@tariq1890 @cdesiniotis @shivamerla  let me know if you see any issue with this change as I don't have the historical context to this. The change looks good to me otherwise.

VincentG1234:
Hi @rahulait, quick follow-up on this PR — not sure if it's waiting on additional internal feedback, but I was wondering if there are still any blockers or updates on it?

tariq1890:
Since it does [align](https://github.com/NVIDIA/gpu-operator/blob/b2b4d1f4398a6be78172c578b895d6460ea3a3ae/controllers/object_controls.go#L5074) with how we render `state-operator-metrics`, I am good to approve this PR

tariq1890 (dismissed):

rahulait (approved):

tariq1890:
@VincentG1234 Can you squash your commit history?

PR #2264, opened 2026-04-02
Run toolkit validation in operand init containers

rajathagasthya:
## Description

The toolkit-validation init containers in operand DaemonSets previously
polled for a toolkit-ready sentinel file on the host. During driver reinstall
cycles, it is possible for operands to (in unknown situations) to find a
stale toolkit-ready file from a previous cycle, passing the init gate
while nvidia-smi would actually fail.

Replace the shell-based sentinel file check with a nvidia-validator
check. This runs nvidia-smi through the toolkit runtime wrapper and
retries until it succeeds, validating both toolkit injection and driver
module readiness without depending on host sentinel files.

## Checklist

### DIFF
--- a/assets/state-dcgm-exporter/0800_daemonset.yaml
+++ b/assets/state-dcgm-exporter/0800_daemonset.yaml
@@ -28,13 +28,20 @@ spec:
       - name: toolkit-validation
         image: "FILLED BY THE OPERATOR"
         command: ['sh', '-c']
-        args: ["until [ -f /run/nvidia/validations/toolkit-ready ]; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
+        args: ["nvidia-validator"]
+        env:
+        - name: NVIDIA_VISIBLE_DEVICES
+          value: "all"
+        - name: WITH_WAIT
+          value: "true"
+        - name: COMPONENT
+          value: toolkit
         securityContext:
           privileged: true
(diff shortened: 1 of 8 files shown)


assets/state-dcgm-exporter/0800_daemonset.yaml
```diff
@@ -28,13 +28,20 @@ spec:
...
+        - name: COMPONENT
+          value: toolkit
         securityContext:
           privileged: true
         volumeMounts:
           - name: run-nvidia
-            mountPath: "/run/nvidia"
-            mountPropagation: HostToContainer
+            mountPath: /run/nvidia
+            mountPropagation: Bidirectional
```
rajatchopra, line 44: Noob question: why does the nvidia-validator want bidrectional mount? It isn't playing with any files is it?

tariq1890:
Closing this PR for now as the bug this was created in response to is still not reliably reproducible.

assets/state-dcgm-exporter/0800_daemonset.yaml
rajathagasthya: It does create and delete status files: https://github.com/NVIDIA/gpu-operator/blob/35d35715dd3f2441ddf8323e7f01a2f006116824/cmd/nvidia-validator/main.go#L1151-L1171

PR #2269, opened 2026-04-06
Remove the CC_CAPABLE_DEVICE_IDS variable

manuelh-dev:
The use of this variable has been obsoleted in k8s-cc-manager, hence removing the variable from values.yaml

### DIFF
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -513,9 +513,6 @@ ccManager:
   version: v0.3.0
   imagePullPolicy: IfNotPresent
   imagePullSecrets: []
-  env:
-    - name: CC_CAPABLE_DEVICE_IDS
-      value: "0x2339,0x2331,0x2330,0x2324,0x2322,0x233d"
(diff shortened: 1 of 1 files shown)


rajatchopra:
Looks good. Just wondering why https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-cc-manager/0500_daemonset.yaml never had the env passed into it.

manuelh-dev:
@cdesiniotis @tariq1890 - could you please take a look?

cdesiniotis (approved):

cdesiniotis:
> Looks good. Just wondering why https://github.com/NVIDIA/gpu-operator/blob/main/assets/state-cc-manager/0500_daemonset.yaml never had the env passed into it.

This env used to be hardcoded in the daemonset template, but was removed in https://github.com/NVIDIA/gpu-operator/commit/eea29f1f47a36f27332056ba55daf07fff10b0d8. Looks like we forgot to remove it from the helm values as well. Thanks @manuelh-dev for the contribution!

PR #2281, opened 2026-04-09
default to the R580 driver instead of R595

tariq1890:
We want to continue defaulting to R580 as the upcoming release of gpu-operator is a patch release. Pointing to a new driver branch should be done in a major/minor release of gpu-operator.


rahulait:
Should we keep main on 595 and release-26.3 on 580?

tariq1890:
Good suggestion, I am open to that

tariq1890:
Created a new PR - https://github.com/NVIDIA/gpu-operator/pull/2283

Issue #2271, opened 2026-04-07
[Feature]: Allow specific annotation to be set on dgcm-exporter pod's daemonset


MadJlzz:
We're still using annotation based service discovery for prometheus to scrape our metrics endpoints. Right now, afaik it is not possible  to specify annotations to pods created by the dcgm exporter. 

Same remark for the service that exposes the exporter.

Is it something we would like to support?

francisguillier:
hi @MadJlzz,

is the request to be able to set custom annotations for the DCGM exporter pod and the nvidia-dcgm-exporter K8s service?

MadJlzz:
Yes exactly so that we can use the service discovery of prometheus based on annotations as well as the CRD one.

rahulait:
Does setting annotations for all daemonsets in your environment work?
https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/values.yaml#L38-L39

MadJlzz:
The answer is yes and no 😅 

Yes, it does work - setting `daemonsets.annotations` sets annotations properly for both the `DaemonSet` and it's underlying `Pods`.

No, it does not work because it creates unreachable `Targets` for prometheus since not every pods of the daemonsets are exposing metrics.

See:

<img width="1864" height="852" alt="Image" src="https://github.com/user-attachments/assets/c65967c2-c3c5-4d1b-8d32-700c2948e64b" />

rahulait:
Can you add additional relabel_configs to filter based on pod labels? So annotation will select all candidates and then filter on specific labels/podnames which are exposing metrics?

MadJlzz:
It's an option that I wanted to avoid but it's doable if we do not want to implement `extraAnnotations` for the `dgcm-exporter` only.
Imho, it complexifies prometheus configuration for nothing but I guess that's my own preference - annotations are made to filter what I scrape already.

rahulait:
We haven't seen such requests from customers and don't have any immediate plans to implement extra annotations, but contributions are welcomed. Please feel free to open a PR with your implementation and we can review and look to get it in.

MadJlzz:
Sure, I will try to see if I can propose something.

rahulait:
Just for the notes, there is another previous duplicate of this feature request: https://github.com/NVIDIA/gpu-operator/issues/1716
Hopefully we'll get this merged soon and can support this feature request.

MadJlzz:
[...]

PR #2292, opened 2026-04-13
[Feature]: Allow specific annotation to be set on dgcm-exporter pod's daemonset

MadJlzz:
## Description

Add support to set extra annotations to `dcgm-exporter` only.
Solves #2271 

## Checklist

## Testing

Added multiple units tests for now:
- add annotations on `dcgm-exporter` when `daemonsets.annotations` is empty
- combine annotations of `dcgm-exporter` and `daemonsets.annotations` when both are set

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1755,6 +1755,12 @@ func TransformDCGMExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 	if len(config.DCGMExporter.ImagePullSecrets) > 0 {
 		addPullSecrets(&obj.Spec.Template.Spec, config.DCGMExporter.ImagePullSecrets)
 	}
+
+	// merge extra annotations at the pod template level
+	if len(config.DCGMExporter.Annotations) > 0 {
+		addExtraAnnotations(obj, config.DCGMExporter.Annotations)
+	}
+
 	// set resource limits
 	if config.DCGMExporter.Resources != nil {
 		// apply resource limits to all containers
@@ -1849,6 +1855,15 @@ func TransformDCGMExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
 	}
 
 	return nil
+}
+
+func addExtraAnnotations(obj *appsv1.DaemonSet, annotations map[string]string) {
+	if obj.Spec.Template.Annotations == nil {
+		obj.Spec.Template.Annotations = make(map[string]string)
+	}
+	for k, v := range annotations {
+		obj.Spec.Template.Annotations[k] = v
+	}
 }
 
 // TransformDCGM transforms dcgm daemonset with required config as per ClusterPolicy
--- a/api/nvidia/v1/clusterpolicy_types.go
+++ b/api/nvidia/v1/clusterpolicy_types.go
@@ -970,6 +970,11 @@ type DCGMExporterSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced,urn:alm:descriptor:com.tectonic.ui:text"
 	Args []string `json:"args,omitempty"`
 
+	// Optional: Annotations is an unstructured key value map stored with a resource that may be
+	// set by external tools to store and retrieve arbitrary metadata. They are not
+	// queryable and should be preserved when modifying objects.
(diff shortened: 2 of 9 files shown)


MadJlzz:
Hello there maintainers 👋🏼 

This is a naive approach to add extra annotations to `dcgm-exporter` daemonset as well as it's pod. I have nothing against reviewing the entire logic - this is my first time touching a Kubernetes operator.

rahulait:
Thanks @MadJlzz. You'll also need to update helm chart templates to render newly added annotations field.
See here: https://github.com/NVIDIA/gpu-operator/blob/8d9d357d5e49b53b4843ac7b1149aa95aa394a8e/deployments/gpu-operator/templates/clusterpolicy.yaml#L541-L585

And then update dcgmExporter's values in values.yaml with an example
https://github.com/NVIDIA/gpu-operator/blob/8d9d357d5e49b53b4843ac7b1149aa95aa394a8e/deployments/gpu-operator/values.yaml#L280-L328

rahulait:
Thanks @MadJlzz. You can test these changes and see if the built image works fine for you.

Steps to verify:
1. Checkout this branch.
2. Create a custom values.yaml which includes dcgmExporter with annotations and also overwrite gpu-operator image with this values.yaml:
```
operator:
  repository: ghcr.io/nvidia
  image: gpu-operator
  version: bef9213a

dcgmExporter:
  annotations:
    <your annotations here>
```
3. Helm install
```
helm install gpu-operator ./deployments/gpu-operator -n gpu-operator -f values.yaml
```

MadJlzz:
It looks like everything is working as expected.

image

.gitignore
```diff
@@ -33,3 +33,6 @@ kubeconfig
 
 # Bundle artifacts
 bundle.Dockerfile
+
+# User specific tooling
+mise.toml
```
rahulait, line 38: nit: missing newline

rahulait:
Overall, the PR looks good to me. I'll let other maintainers take a look as well and merge as required.

.gitignore
tariq1890: @MadJlzz Can you revert this change? This seems to be specific to your environment. I recommend using a global gitignore in your workstation to handle cases like this.

MadJlzz: @tariq1890 it's done

tariq1890: Thank you!

controllers/object_controls.go
```diff
@@ -1851,6 +1857,19 @@ func TransformDCGMExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
...
 
+func addExtraAnnotations(obj *appsv1.DaemonSet, annotations map[string]string) {
+	if obj.Annotations == nil {
+		obj.Annotations = make(map[string]string)
+	}
+	if obj.Spec.Template.Annotations == nil {
+		obj.Spec.Template.Annotations = make(map[string]string)
+	}
+	for k, v := range annotations {
+		obj.Annotations[k] = v
```
tariq1890, line 1868: Just curious - Do you have a requirement to add annotations to the daemonset level? In most (if not all) cases, it is sufficient to add the annotations at the pod-template level, correct?

MadJlzz: To my knowledge, yes. I went down the path of beeing less strict about it, just in case someone has the need to set annotations at this level too.

tariq1890: The `spec.daemonsets.annotations` field only sets the pod template annotations. For consistency, it would be good to do the same here as well.

MadJlzz: Ok I will do the same then.
Do you want me to rename the values field? Right now it's called `dcgmExporter.annotations`. I could call it `dcgmExporter.podAnnotations` or even `dcgmExporter.podExtraAnnotations`
What do you think?

tariq1890: Right now, I am inclined to just keep the name as-is as that keeps it consistent with what we have now. If we find ourselves needing to add top-level annotations, we will just update the current code to update both daemonset-level and pod-level annotations for all `annotations` fields in the ClusterPolicy

MadJlzz: Cool, I have everything I need - I will come back with the updates soon. (hopefully I will work on this tonight)

MadJlzz: @tariq1890 Changes are pushed!

rahulait (approved):

tariq1890 (approved):
Thank you @MadJlzz !

Issue #2398, opened 2026-04-27
[Bug]: regression in GPU Operator `v26.3.1` around `kataManager` Helm defaults.


Ayush-Rathor:
### Bug description

There appears to be a regression in GPU Operator `v26.3.1` around `kataManager` Helm defaults.

In `v25.10.0`, setting only:

```yaml
kataManager:
  enabled: true
```
was enough for the chart to render a usable ClusterPolicy.spec.kataManager.

In `v26.3.1`, enabling `kataManager` with the same style of values produces an incomplete `ClusterPolicy.spec.kataManager` block that lacks image metadata (repository, image, version), and the operator later panics during reconcile in the Kata runtime class path.

### Environment
- GPU Operator: v26.3.1
- Helm chart: gpu-operator-v26.3.1
- Kubernetes: v1.35.3
- Sandbox workloads:
  - sandboxWorkloads.enabled=true
  - sandboxWorkloads.mode=kata
### Reproduction

Use Helm values like:
```yaml
cdi:
  enabled: true

driver:
  enabled: true
  kernelModuleType: open

toolkit:
  enabled: true

devicePlugin:
  enabled: false

ccManager:
  enabled: true

sandboxWorkloads:
  enabled: true
  defaultWorkload: "container"
  mode: "kata"

kataSandboxDevicePlugin:
  enabled: true

kataManager:
  enabled: true
```

Install / upgrade the chart and inspect the rendered/applied ClusterPolicy.

#### Actual behavior

The resulting ClusterPolicy.spec.kataManager is incomplete, e.g.:
```yaml
kataManager:
  enabled: true
  imagePullPolicy: IfNotPresent
```
with no repository, image, or version.

After that, the operator hits a reconcile panic in the Kata runtime class path, for example:
- `transformKataRuntimeClasses`
- `RuntimeClasses`

```yaml
status:
  conditions:
  - lastTransitionTime: "2026-04-27T08:30:56Z"
    message: ""
    reason: Error
    status: "False"
    type: Ready
  - lastTransitionTime: "2026-04-27T08:30:56Z"
    message: 'Failed to reconcile state-kata-manager: empty image path provided through
      both ClusterPolicy CR and ENV KATA_MANAGER_IMAGE'
    reason: ReconcileFailed
    status: "True"
    type: Error
  namespace: gpu-operator
  state: notReady
```
with a nil pointer dereference.

#### Expected behavior

If kataManager.enabled=true, the chart should either:

[...]

PR #2403, opened 2026-04-27
helm: restore kataManager defaults

Ayush-Rathor:
## Description

This fixes a regression in the Helm chart defaults for `kataManager`, issue [2398](https://github.com/NVIDIA/gpu-operator/issues/2398).

In `v25.10.0`, the chart shipped a complete default `kataManager` block including:
- `config`
- `repository`
- `image`
- `version`

In `v26.3.1`, those defaults were removed, leaving `kataManager` incomplete when enabled through values. This caused the rendered `ClusterPolicy.spec.kataManager` to miss required image metadata and broke reconciliation.

## What changed

Restored the missing default `kataManager` values in `deployments/gpu-operator/values.yaml`.

## Validation

I verified the following:

1. Stock `v26.3.1` reproduced the issue and the `ClusterPolicy` reported:
   - `Failed to reconcile state-kata-manager: empty image path provided through both ClusterPolicy CR and ENV KATA_MANAGER_IMAGE`

2. With this patch applied and the chart rendered locally, the `kataManager` block was complete again.

3. Upgrading from the patched local chart resulted in:
   - `ClusterPolicy.status.state: ready`
   - `ClusterPolicy is ready as all resources have been successfully reconciled`

## Notes

For local source-based validation only, I temporarily adjusted `Chart.yaml` so the raw source chart would not use `main-latest`, but that change is not part of this PR.

### DIFF
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -492,7 +492,23 @@ vfioManager:
 
 kataManager:
   enabled: false
-  config: {}
+  config:
+    artifactsDir: "/opt/nvidia-gpu-operator/artifacts/runtimeclasses"
+    runtimeClasses:
+      - name: kata-nvidia-gpu
+        nodeSelector: {}
+        artifacts:
+          url: nvcr.io/nvidia/cloud-native/kata-gpu-artifacts:ubuntu22.04-535.54.03
+          pullSecret: ""
+      - name: kata-nvidia-gpu-snp
+        nodeSelector:
+          "nvidia.com/cc.capable": "true"
+        artifacts:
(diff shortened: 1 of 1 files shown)


rajatchopra (commented):
@Ayush-Rathor thanks for this PR.
However, we intend to deprecate KataManager. It is not needed anymore as functionality has moved to cc-manager and sandbox-device-plugin.

`the operator hits a reconcile panic in the Kata runtime class path` is what we really need to fix here.

Ayush-Rathor-1:
Got that, just saw your PR for the same

rajatchopra:
@Ayush-Rathor If you could help review the PR - https://github.com/NVIDIA/gpu-operator/pull/2407
Thanks.

Closing this one out meanwhile. Please re-open if you feel otherwise.

PR #2406, opened 2026-04-27
Add DCGM exporter pod metadata enrichment API

### LINKED ISSUE #2009: Pod Label not visible in DCGM Exporter Metrics
_**Important Note:  NVIDIA AI Enterprise customers can get support from NVIDIA Enterprise support. Please open a case [here](https://enterprise-support.nvidia.com/s/create-case)**._

**Describe the bug**
The GPU Operator is the recommended install mode for DCGM Exporter however it doesn't seem to support enabling Pod Labels for metrics.

**To Reproduce**
Deploy GPU-Operator and enable DCGM Exporter with extra environment variables

- DCGM_EXPORTER_KUBERNETES_ENABLE_POD_LABELS
- DCGM_EXPORTER_KUBERNETES_ENABLE_POD_UID

**Expected behavior**
Deployment should create

- ClusterRole and ClusterRoleBinding to ServiceAccount used by dcgm export pods
- AutoMount Service Account Token to allow Pods read Kubernetes API
- Mount kubelet path as a volume by DCGM pods

 - GPU Operator Version: v23.6.1
 - OS: Rocky-Linux-8.10
 - Kernel Version: [e.g. 6.8.0-generic]
 - Container Runtime Version: containerd  v1.7.1
 - Kubernetes Distro and Version: K8s v1.24.12

**Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/)** (optional if deemed irrelevant)

 - [ ] kubernetes pods status: `kubectl get pods -n OPERATOR_NAMESPACE`
 - [ ] kubernetes daemonset status: `kubectl get ds -n OPERATOR_NAMESPACE`
 - [ ] If a pod/ds is in an error state or pending state `kubectl describe pod -n OPERATOR_NAMESPACE POD_NAME`
 - [ ] If a pod/ds is in an error state or pending state `kubectl logs -n OPERATOR_NAMESPACE POD_NAME --all-containers`
 - [ ] Output from running `nvidia-smi` from the driver container: `kubectl exec DRIVER_POD_NAME -n OPERATOR_NAMESPACE -c nvidia-driver-ctr -- nvidia-smi`
 - [ ] containerd logs `journalctl -u containerd > containerd.log`

Collecting full debug bundle (optional):

```
curl -o must-gather.sh -L https://raw.githubusercontent.com/NVIDIA/gpu-operator/main/hack/must-gather.sh
chmod +x must-gather.sh
./must-gather.sh
```
**NOTE**: please refer to the [must-gather](https://raw.githubusercontent.com/NVIDIA/gpu-operator/main/hack/must-gather.sh) script for debug data collected.

This bundle can be submitted to us via email: **operator_feedback@nvidia.com**

karthikvetrivel:
## Description

Resolves https://github.com/NVIDIA/gpu-operator/issues/2009. 

Previously, there was no way of enabling pod metrics when using DCGM exporter through GPU Operator. This fix introduces introduces first-class Helm values that provision RBAC and allow pod metrics to include Pod UID & Pod Label. 

**Design Choices:** 

- Introduced enabling pod metrics as first-class Helm values instead of making arbitrary env vars provision RBAC to match the upstream DCGM metrics helm chart. 
- I intentionally did not add resourceSlices RBAC. Standalone dcgm-exporter Helm includes resourceSlices because it also supports DRA-related Kubernetes enrichment. GPU Operator is only exposing pod metadata enrichment here because it does not support DRA yet. I am open to changing this. 

## Checklist

- [X] Test cases are added for new code paths

## Testing

Added unit tests.
Confirmed the operator reconciled the expected DCGM exporter resources:
- `nvidia-dcgm-exporter` DaemonSet rolled out successfully.
- DCGM exporter container received:
      - `DCGM_EXPORTER_KUBERNETES_ENABLE_POD_LABELS=true`
      - `DCGM_EXPORTER_KUBERNETES_ENABLE_POD_UID=true`
      - `DCGM_EXPORTER_KUBERNETES_POD_LABEL_ALLOWLIST_REGEX=^gpu_test_label$`
    - Pod template had `automountServiceAccountToken: true`.
    - `nvidia-dcgm-exporter-read-pods` ClusterRole and ClusterRoleBinding were created.
- Exporter service account could `get`, `list`, and `watch` pods cluster-wide.
    
Scraped DCGM exporter metrics from a labeled GPU pod and confirmed metrics included `gpu_test_label="live-test"` and `pod_uid="..."`


controllers/object_controls.go
```diff
@@ -466,6 +466,19 @@ func ClusterRole(n ClusterPolicyController) (gpuv1.State, error) {
...
 	}
 
+	// For state-dcgm-exporter, cluster-scoped pod-read RBAC is only needed
+	// when pod-metadata enrichment is opted in.
+	if n.stateNames[n.idx] == "state-dcgm-exporter" &&
+		obj.Name == "nvidia-dcgm-exporter-read-pods" &&
+		!n.singleton.Spec.DCGMExporter.IsKubernetesPodMetadataEnabled() {
+		err := n.client.Delete(ctx, obj)
+		if err != nil && !apierrors.IsNotFound(err) {
+			logger.Info("Couldn't delete", "Error", err)
```
rahulait: This should be Error() instead of Info().

controllers/object_controls.go
```diff
@@ -507,6 +520,18 @@ func ClusterRoleBinding(n ClusterPolicyController) (gpuv1.State, error) {
 		return gpuv1.Disabled, nil
 	}
 
+	// Mirror the ClusterRole gate.
+	if n.stateNames[n.idx] == "state-dcgm-exporter" &&
+		obj.Name == "nvidia-dcgm-exporter-read-pods" &&
+		!n.singleton.Spec.DCGMExporter.IsKubernetesPodMetadataEnabled() {
+		err := n.client.Delete(ctx, obj)
+		if err != nil && !apierrors.IsNotFound(err) {
+			logger.Info("Couldn't delete", "Error", err)
```
rahulait: Error instead of Info

rahulait (approved):
LGTM

controllers/object_controls.go
karthikvetrivel: Changed. I want to flag that the file uses this pattern: `logger.Info("Couldn't delete", "Error", err)` though. I still changed it, but something to keep note of.

assets/state-dcgm-exporter/0210_clusterrole.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nvidia-dcgm-exporter-read-pods
+  labels:
+    app: nvidia-dcgm-exporter
+# Resourceslices are DRA-only and should be added when GPU Operator exposes DRA exporter support.
```
tariq1890: Is this a TODO comment?

controllers/object_controls_test.go
```diff
@@ -1151,6 +1151,12 @@ func getDCGMExporterTestInput(testCase string) *gpuv1.ClusterPolicy {
 	case "standalone-dcgm":
 		dcgmEnabled := true
 		cp.Spec.DCGM.Enabled = &dcgmEnabled
+	case "pod-labels-enabled":
+		enabled := true
+		cp.Spec.DCGMExporter.EnablePodLabels = &enabled
```
tariq1890: why not just do this?
```suggestion
		cp.Spec.DCGMExporter.EnablePodLabels = ptr.To(true)
```

assets/state-dcgm-exporter/0210_clusterrole.yaml
karthikvetrivel: This was just to explain why there isn't parity with the upstream DCGM Exporter clusterrole, which contains resource slices. Can remove if you'd like?

controllers/object_controls_test.go
karthikvetrivel: Fixed, good callout

assets/state-dcgm-exporter/0210_clusterrole.yaml
tariq1890: Yes, either add a TODO comment with the intended change in the future or just remove

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -1077,6 +1077,28 @@ type DCGMExporterSpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="HPC Job Mapping Configuration"
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:advanced"
 	HPCJobMapping *DCGMExporterHPCJobMappingConfig `json:"hpcJobMapping,omitempty"`
+
+	// Enable Kubernetes pod labels in metrics. Requires cluster-level read access to pods.
```
tariq1890: This feels a bit too terse. Can we expand on this a bit more? Let's also clarify that this setting adds pod label as a label dimension to the DCGM exporter prometheus metrics. 

Also, the "Requires cluster-level read access to pods" is a bit unnecessary IMO. If we do want to mention this, can we enclose it in parenthesis?

assets/state-dcgm-exporter/0210_clusterrole.yaml
karthikvetrivel: Updated to be a TODO!

api/nvidia/v1/clusterpolicy_types.go
karthikvetrivel: Yep, expanded on the label and enclosed "Requires cluster-level read access to pods" in parentheses.

assets/state-dcgm-exporter/0210_clusterrole.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
rajathagasthya, line 2: This will always add a ClusterRole regardless of `EnablePodLabels`. Is this intended?

rajathagasthya: Or, is this being deleted in `object_controls.go` if `EnablePodLabels=false`?

karthikvetrivel: Yep, this is gated in `objects_controls.go`. I have the YAML live in the asset dir unconditionally because that's how `state-dcgm-exporter` loads its manifest set I believe.

rajathagasthya: Does this first create this ClusterRole unconditionally and then cleanup during reconcile when it reads `EnablePodLabels=false`?

tariq1890: The existence of this file in the assets folder doesn't mean that the manifest will be created in the live cluster. During reconciliation, the candidate manifests would be loaded from file. After the transformation methods are called, these manifests would either be excluded or included depending on the ClusterPolicy values and transformation logic in `object_controls.go`. The final list of resources (in memory) are then created or updated in the cluster after all the transformations

karthikvetrivel: No, it is gated before the `client.Create` call, so it is never called for `nvidia-dcgm-exporter-read-pods` when `EnablePodLabels=false`.

rajathagasthya: Got it, thanks for clarifying!

rajathagasthya (approved):
LGTM.

assets/state-dcgm-exporter/0800_daemonset.yaml
```diff
@@ -24,6 +24,7 @@ spec:
           effect: NoSchedule
       priorityClassName: system-node-critical
       serviceAccountName: nvidia-dcgm-exporter
+      automountServiceAccountToken: false
```
tariq1890, line 27: I am a bit wary of this change. This changes the default behaviour of the dcgm-exporter that is deployed by gpu-operator

karthikvetrivel: That is a fair concern, but this aligns gpu-operator's `dcgm-exporter` daemonset with upstream dcgm-exporter's own packaging defaults — both the standalone manifest and the Helm chart already ship with `automountServiceAccountToken: false`, with the chart gating it to true only when pod label/UID enrichment or DRA is enabled. This PR mirrors that contract on the gpu-operator side.

controllers/object_controls.go
```diff
@@ -507,6 +520,18 @@ func ClusterRoleBinding(n ClusterPolicyController) (gpuv1.State, error) {
 		return gpuv1.Disabled, nil
 	}
 
+	// Mirror the ClusterRole gate.
+	if n.stateNames[n.idx] == "state-dcgm-exporter" &&
```
tariq1890: In daemonsets, we have state-specific (or operand-specific) methods that are called through the `preProcessDaemonSet` method which is invoked by `DaemonSet()`. Can we have something similar in the `ClusterRole()` method?

karthikvetrivel: Updated to use the same pattern. Take another look when you can.

tariq1890: Thanks @karthikvetrivel !

controllers/object_controls.go
```diff
@@ -447,6 +447,20 @@ func RoleBinding(n ClusterPolicyController) (gpuv1.State, error) {
 	return gpuv1.Ready, nil
 }
 
+var rbacGates = map[string]func(*gpuv1.ClusterPolicySpec) bool{
+	"nvidia-dcgm-exporter-read-pods": func(config *gpuv1.ClusterPolicySpec) bool {
+		return config.DCGMExporter.IsKubernetesPodMetadataEnabled()
+	},
+}
+
+func isRBACEnabled(name string, config *gpuv1.ClusterPolicySpec) bool {
```
tariq1890, line 456: Great work here!

tariq1890 (approved):

PR #2424, opened 2026-05-04
Expand DRA Driver Integration for GPU Operator

karthikvetrivel:
## Purpose

Umbrella branch for NVIDIA DRA driver integration into GPU Operator, building on Chris's foundation work in #1541. Refreshes embedded DRA assets to the newer driver layout and adds the operator-side glue needed to make DRA usable.

## Change Log

[`dd7cf4d`](https://github.com/NVIDIA/gpu-operator/pull/2424/changes/dd7cf4d73cdffdb502caae47f660f6e6a45e3569):

- Refresh embedded DRA manifests to the new driver layout: split kubelet-plugin RBAC/SAs into separate `nvidia-dra-driver-kubeletplugin` and `compute-domain-daemon-service-account` identities, update controller + kubelet-plugin workloads, drop the legacy entrypoint ConfigMap, and switch to the `dra-driver-nvidia-gpu` image.
- CRD packaging: refresh `ComputeDomain`, add new `ComputeDomainClique`, wire both into the Helm CRD upgrade hook, Dockerfile, and OLM bundle.

[`3be3986`](https://github.com/NVIDIA/gpu-operator/pull/2424/changes/3be39863d68a8a224de0f91e02fb4e9035cdd04c):

- Adds three rules to `gpu-operator`'s ClusterRole: `pods/patch` (consumed by compute-domain-daemon), `resource.nvidia.com/computedomaincliques` CRUD (the new sub-resource), and `coordination.k8s.io/leases` (DRA controller leader election).

[`f9c67dd`](https://github.com/NVIDIA/gpu-operator/pull/2424/changes/f9c67dd616b959dd88216289f185d3493498f3e0):

- New `draDriver.featureGates` field (`map[string]bool`) on ClusterPolicy. Operator renders it as a `FEATURE_GATES` env var on all three DRA driver containers (controller, `gpus` kubelet plugin, `compute-domains` kubelet plugin) — same channel the upstream `k8s-dra-driver-gpu` Helm chart uses, identical `Key=Value,` format with trailing comma.
- Keys are sorted alphabetically before rendering (upstream chart doesn't — Go map iteration is randomized per call). This avoids spurious pod restarts caused by the env-var string shuffling on each reconcile pass. The binary's parser is order-insensitive, so observable behavior is identical to upstream.


kasia-kujawa:
When the GPU Operator supports the NVIDIA DRA driver, it would be useful to have an option to enable it only on selected nodes. This would allow some workloads to use the DRA, while others continue using the approach with NVIDIA device plugin on different nodes. 

It  would be especially helpful during the migration phase, when users are ready to migrate only some of workloads to DRA.

PR #2434, opened 2026-05-06
Ship static busybox shell in gpu-operator image

### LINKED ISSUE #2435: Ship static busybox shell in gpu-operator image
Part of NVIDIA/cloud-native-team#299.

NVIDIA STIG policy is dropping `-dev` distroless tags as approved
parent images. To move the gpu-operator image onto a non-`-dev` base
without rewriting validator/operand manifests or `hack/must-gather.sh`,
we ship a static busybox shell inside the image.

## Approach

Add a `debian:trixie-slim` shell-builder stage to `docker/Dockerfile`
that apt-installs `busybox-static`. Copy `/bin/busybox` into the final
stage and run `busybox --install -s /bin` to populate `/bin/sh`,
`/bin/rm`, `/bin/ln`, etc. Flip the final base from
`nvcr.io/nvidia/distroless/cc:<ver>-dev` to the matching non-`-dev`
variant.

```dockerfile
FROM debian:trixie-slim AS shell-builder
RUN apt-get update \
 && apt-get install --no-install-recommends -y busybox-static \
 && rm -rf /var/lib/apt/lists/*

FROM nvcr.io/nvidia/distroless/cc:<version>
COPY --from=shell-builder /bin/busybox /bin/busybox
RUN ["/bin/busybox", "--install", "-s", "/bin"]
```

Validator-pod manifests, operand-asset `sh -c` wrappers,
`hack/must-gather.sh` (mounted at `/usr/bin/gather`), and the
`rmglob` preStop helper all remain untouched. Mechanical Dockerfile
change.

Precedent:
[`kubernetes-sigs/dra-driver-nvidia-gpu`](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu)
uses the same pattern with `bash-static`.

## Tracking

Tracked in PR NVIDIA/gpu-operator#2434, being repurposed from the
earlier Go-absorption approach.

## Acceptance

- [ ] Dockerfile no longer references a `-dev` distroless tag
- [ ] `/bin/sh` and friends are present at runtime via static busybox
- [ ] Image builds successfully on amd64 and arm64
- [ ] Validator pods, operand-asset init containers, and the
      `gather` plugin all run as before

rajathagasthya:
Flip the base from `*-dev*` to non-`*-dev*` distroless and source a static busybox from `debian:trixie-slim`. Init container wrappers, lifecycle hooks, and helper scripts continue to work via `/bin/sh` and busybox applet symlinks layered into the final image.

Part of NVIDIA/cloud-native-team#299.
Resolves #2435.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -70,9 +70,20 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
     make && \
     cp ${SAMPLE_NAME} /build/${SAMPLE_NAME}
 
+# Build a static busybox layout: one binary plus applet symlinks (sh, rm,
+# ln, sleep, cat, ...) so PATH-resolved commands in init-container wrappers
+# and lifecycle hooks keep working on the non-*-dev* distroless base.
+FROM debian:trixie-slim AS shell
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends busybox-static \
+ && rm -rf /var/lib/apt/lists/* \
+ && mkdir /busybox \
+ && cp /bin/busybox /busybox/busybox \
+ && /busybox/busybox --install -s /busybox
+
 # The C/C++ distroless image is used as a base since the CUDA vectorAdd
 # sample application depends on C/C++ libraries.
-FROM nvcr.io/nvidia/distroless/cc:v4.0.6-dev
+FROM nvcr.io/nvidia/distroless/cc:v4.0.6
 
 ENV NVIDIA_VISIBLE_DEVICES=void
 
@@ -89,6 +100,12 @@ LABEL description="See summary"
 LABEL vsc-ref=${GIT_COMMIT}
 
 WORKDIR /
+
+COPY --from=shell /busybox /busybox
+USER 0:0
+RUN ["/busybox/ln", "-s", "/busybox/sh", "/bin/sh"]
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/busybox
+
 COPY --from=builder /workspace/gpu-operator /usr/bin/
 COPY --from=builder /workspace/manage-crds /usr/bin/
 COPY --from=builder /workspace/nvidia-validator /usr/bin/


docker/Dockerfile
```diff
@@ -70,9 +70,20 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
...
     cp ${SAMPLE_NAME} /build/${SAMPLE_NAME}
 
+# Static busybox shell + coreutils for the non-*-dev* distroless base.
+FROM debian:trixie-slim AS shell
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends busybox-static \
+ && rm -rf /var/lib/apt/lists/* \
+ && mkdir /busybox \
+ && cp /bin/busybox /busybox/busybox \
+ && /busybox/busybox --install -s /busybox
```
cdesiniotis, line 82: What's the purpose of this statement? What is `--install` actually doing?

docker/Dockerfile
```diff
@@ -70,9 +70,20 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
...
+
 # The C/C++ distroless image is used as a base since the CUDA vectorAdd
 # sample application depends on C/C++ libraries.
-FROM nvcr.io/nvidia/distroless/cc:v4.0.6-dev
+FROM nvcr.io/nvidia/distroless/cc:v4.0.6
+
+USER 0:0
```
cdesiniotis: This is just needed to create the `/bin/sh` symlink, correct? A minor nit would be to move this line immediately before where it is needed for clarity (e.g. L103).

cdesiniotis (commented):
A couple of minor comments / questions. Otherwise this looks good to me!

docker/Dockerfile
rajathagasthya: Correct, it's needed to create the symlink. It makes sense to move that line for clarity.

docker/Dockerfile
rajathagasthya: `--install` creates links for all the busybox internal commands (`rm`, `cat`, etc.). The `-s` makes them a symlink. What we're doing here is installing those symlinks in `/busybox`, copying that directory into the final image (symlinks are preserved) and adding `/busybox` to `PATH` so direct command invocations are resolved correctly.

Here's what `/busybox` directory looks like:

```shell
$ ls -l /busybox
-rwxr-xr-x 1 root root 1975064 May 20 04:48  busybox
lrwxrwxrwx 1 root root      16 May 20 04:48  cat -> /busybox/busybox
lrwxrwxrwx 1 root root      16 May 20 04:48  chgrp -> /busybox/busybox
lrwxrwxrwx 1 root root      16 May 20 04:48  chmod -> /busybox/busybox
lrwxrwxrwx 1 root root      16 May 20 04:48  chown -> /busybox/busybox
lrwxrwxrwx 1 root root      16 May 20 04:48  chroot -> /busybox/busybox
```
My understanding is invocation of, say `cat` (which is resolved via `$PATH`), is dispatched as `/busybox/busybox cat` since `busybox` is a "multi-call binary".

tariq1890: The only reason we used busybox before was because it came with the `distroless-dev` images. Since we are moving away from them, why can't we just add a statically-built bash binary to the final image? Wouldn't bash be much simpler to use than a busybox shell?

rajathagasthya: The deployment manifests call out to coreutils like `rm`, `cat`, etc directly. We don't get those if we just add a static bash binary.

tariq1890: Fair enough

cdesiniotis (approved):

tariq1890 (approved):

PR #2455, opened 2026-05-13
Set go directive to 1.26.0 in module

rajathagasthya:
## Description

Minimum Go version required to build. Matches main branch.

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/NVIDIA/gpu-operator
 
-go 1.25.7
+go 1.26.0
 
 require (
 	github.com/Masterminds/sprig/v3 v3.3.0


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/NVIDIA/gpu-operator
 
-go 1.25.7
+go 1.26.0
```
tariq1890, line 3: To be honest, I prefer updating this indirectly via a dependency update. This way, we better understand our dep tree and what dependencies push our go backward compat limits

rajathagasthya: Good point! I thought I missed this when I bumped k8s.io dependencies manually in this branch in #2445. But the on the main branch Go version was incremented because k8s.io deps were bumped to `v0.36.0` (instead of `v0.35.x` in this branch).

I'll close this PR.

Issue #1357, opened 2025-03-23
Driver validation doesn't succeed because /usr/bin is a symlink


wokalski:
On my OS (NixOS) /usr/bin is a symlink to /run/current-system/sw/bin

```
[root@nvidia-operator-validator-9nk4h /]# ls -lah /host/usr/bin
lrwxrwxrwx 1 root root 26 Feb 16 13:09 /host/usr/bin -> /run/current-system/sw/bin
```

This means that the validation will always fail here:
https://github.com/NVIDIA/gpu-operator/blob/e9b7ec16203987508eddc048fe82a4c44d3b392a/validator/main.go#L703C1-L704C1

If I'm not mistaken we should simply syscal.chroot there but I'm not sure if the following symlink creation would still work.

cdesiniotis:
@wokalski what is the behavior you are observing? From reading https://pkg.go.dev/os#Lstat, my understanding is that the `os.Lstat()` call would not return an error here.

wokalski:
The pod has host fs mounted under /host so a symbolic link that resolves to /foo/bar doesn't work because the real location is /host/foo/bar.

cdesiniotis:
What are the validator pod logs? Based on the below from https://pkg.go.dev/os#Lstat, I was not expecting the link to be followed:

> Lstat makes no attempt to follow the link

wokalski:
There are no interesting logs cuz it's running silently. Notice that /usr/bin is a symlink and it's lstating /use/bin/nvidia-smi so it will never find it if it's not followed 😅. Right my initial description was confusing but the two things are interconnected.

cdesiniotis:
Got it, thanks for the further clarification. The fact that `os.Lstat` does not follow symlinks means that it will think `/host/usr/bin/nvidia-smi` does not exist. 

As you may already know, NixOS is not a [supported OS](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/platform-support.html#supported-operating-systems-and-kubernetes-platforms). It may still be valuable to re-evaluate the validator code and and see if it can be tweaked to work on a system where `/usr/bin` is a symlink. I am not familiar with NixOS so I am not sure if there are other issues that would be encountered besides this.

jcstryker:
this is still an active issue for nixos-managed nodes

rahulait:
This is done and fix will be available starting v26.7.0

PR #2464, opened 2026-05-15
Update Validator for /usr/bin symlink

JunAr7112:
## Description

This MR is in response to this issue: https://github.com/NVIDIA/gpu-operator/issues/1357. In certain scenarios, we might have a symlink with /usr/bin which causes os.Lstat() to be unable to properly parse the nvidia-smi location. Inside the validator container, the host root filesystem is mounted at /host. So the validator tries to resolve: /host/usr/bin/nvidia-smi
But because /host/usr/bin points to an absolute symlink target, /run/.../bin, the path resolution happens from the validator container’s root, not from /host. In other words, it looks for: /run/.../bin/nvidia-smi inside the container namespace, instead of /host/run/.../bin/nvidia-smi. So the validator falsely concludes that the host-installed driver is not present.

## Change
nvidiaSMIPath, err := symlink.FollowSymlinkInScope(filepath.Join(hostRootCtrPath, "usr/bin/nvidia-smi"), hostRootCtrPath)

We will switch to using symlink.FollowSymlinkInScope to resolve absolute paths in a helper function resolveHostNvidiaSMI(hostRootCtrPath string) (os.FileInfo, error). For testing purposes, it can be set to a different value. We will need to update the func (v *VGPUManager) runValidation(...) and the func validateHostDriver(...).

## Testing

Added a Unit test:

local-agadiyar@ipp1-2290:~/gpu-operator$ go test ./cmd/nvidia-validator -run TestResolveHostNvidiaSMI -count=1
ok      github.com/NVIDIA/gpu-operator/cmd/nvidia-validator     0.020s

TestResolveHostNvidiaSMI verifies that the validator can find host nvidia-smi through the mounted host root, including when /usr/bin is an absolute symlink inside that host root. It builds small fake host filesystems in temp dirs, creates regular files or symlinks from the test case map, calls resolveHostNvidiaSMI(hostRoot), and checks that the helper succeeds for valid layouts and fails when nvidia-smi is missing.

### DIFF
--- a/cmd/nvidia-validator/main.go
+++ b/cmd/nvidia-validator/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/NVIDIA/go-nvlib/pkg/nvmdev"
 	"github.com/NVIDIA/go-nvlib/pkg/nvpci"
 	devchar "github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk/system/create-dev-char-symlinks"
+	"github.com/cyphar/filepath-securejoin/pathrs-lite"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert/yaml"
 	cli "github.com/urfave/cli/v3"
@@ -742,6 +743,22 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 	return false, nil
 }
 
+// resolveHostNvidiaSMI opens and stats nvidia-smi within the mounted host root.
+func resolveHostNvidiaSMI(hostRootCtrPath string) (os.FileInfo, error) {
+	f, err := pathrs.OpenInRoot(hostRootCtrPath, "/usr/bin/nvidia-smi")
+	if err != nil {
+		return nil, fmt.Errorf("failed to open 'nvidia-smi' on the host: %w", err)
+	}
+	defer f.Close()
+
+	fileInfo, err := f.Stat()
+	if err != nil {
+		return nil, fmt.Errorf("failed to stat 'nvidia-smi' on the host: %w", err)
+	}
+
+	return fileInfo, nil
+}
+
 func validateHostDriver(silent bool) error {
 	log.Info("Attempting to validate a pre-installed driver on the host")
 	if fileInfo, err := os.Lstat(filepath.Join("/host", wslNvidiaSMIPath)); err == nil && fileInfo.Size() != 0 {
@@ -749,9 +766,10 @@ func validateHostDriver(silent bool) error {
 		disableDevCharSymlinkCreation = true
 		return nil
 	}
-	fileInfo, err := os.Lstat("/host/usr/bin/nvidia-smi")
+
+	fileInfo, err := resolveHostNvidiaSMI("/host")
 	if err != nil {
-		return fmt.Errorf("no 'nvidia-smi' file present on the host: %w", err)
+		return err
 	}
 	if fileInfo.Size() == 0 {
 		return fmt.Errorf("empty 'nvidia-smi' file found on the host")
@@ -1753,7 +1771,7 @@ func (v *VGPUManager) runValidation(silent bool) (hostDriver bool, err error) {
 	args := []string{"/run/nvidia/driver", "nvidia-smi"}
 
 	// check if driver is pre-installed on the host and use host path for validation
-	if _, err := os.Lstat("/host/usr/bin/nvidia-smi"); err == nil {
+	if _, err := resolveHostNvidiaSMI("/host"); err == nil {
 		args = []string{"/host", "nvidia-smi"}
 		hostDriver = true
 	}
--- a/cmd/nvidia-validator/main_test.go
+++ b/cmd/nvidia-validator/main_test.go
@@ -18,9 +18,83 @@ package main
 
 import (
 	"context"
+	"maps"
 	"os"
+	"path/filepath"
+	"slices"
+	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
+
+func TestResolveHostNvidiaSMI(t *testing.T) {
+	testCases := []struct {
+		description  string
+		contents     map[string]string
+		expectsError bool
+	}{
+		{
+			description: "nvidia-smi exists in /usr/bin",
+			contents: map[string]string{
+				"/usr/bin/nvidia-smi": "fake nvidia-smi",
+			},
+		},
+		{
+			description: "nvidia-smi exists through absolute /usr/bin symlink",
+			contents: map[string]string{
+				"/run/current-system/sw/bin/nvidia-smi": "fake nvidia-smi",
+				"/usr/bin":                              "symlink=/run/current-system/sw/bin",
+			},
+		},
+		{
+			description: "nvidia-smi exists through relative /usr/bin symlink",
+			contents: map[string]string{
+				"/run/current-system/sw/bin/nvidia-smi": "fake nvidia-smi",
+				"/usr/bin":                              "symlink=../run/current-system/sw/bin",
+			},
+		},
+		{
+			description: "parent dir is symlink to path not within root",
+			contents: map[string]string{
+				"/usr/bin": "symlink=../../",
+			},
+			expectsError: true,
+		},
+		{
+			description:  "nvidia-smi does not exist",
+			expectsError: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			hostRoot := t.TempDir()
+			// Iterate in sorted order so parent symlinks are created before paths under them.
+			for _, name := range slices.Sorted(maps.Keys(tc.contents)) {
+				contents := tc.contents[name]
+				target := filepath.Join(hostRoot, name)
+				require.NoError(t, os.MkdirAll(filepath.Dir(target), 0755))
+
+				if strings.HasPrefix(contents, "symlink=") {
+					require.NoError(t, os.Symlink(strings.TrimPrefix(contents, "symlink="), target))
+					continue
+				}
+
+				require.NoError(t, os.WriteFile(target, []byte(contents), 0600))
+			}
+
+			fileInfo, err := resolveHostNvidiaSMI(hostRoot)
+			if tc.expectsError {
+				require.Error(t, err)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotZero(t, fileInfo.Size())
+		})
+	}
+}
 
 func Test_isValidComponent(t *testing.T) {
 	tests := []struct {
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/NVIDIA/k8s-kata-manager v0.2.3
 	github.com/NVIDIA/k8s-operator-libs v0.0.0-20260505175649-fa6a3643c441
 	github.com/NVIDIA/nvidia-container-toolkit v1.19.1
+	github.com/cyphar/filepath-securejoin v0.6.1
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/go-logr/logr v1.4.3
 	github.com/go-logr/zapr v1.3.0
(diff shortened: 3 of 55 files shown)


cmd/nvidia-validator/main.go
```diff
@@ -749,7 +749,17 @@ func validateHostDriver(silent bool) error {
 		disableDevCharSymlinkCreation = true
 		return nil
 	}
-	fileInfo, err := os.Lstat("/host/usr/bin/nvidia-smi")
+
+	nvidiaSMIPath := "/host/usr/bin/nvidia-smi"
```
rahulait: Can we move this newly added logic to a new function which correctly resolves the nvidia-smi path? Then, we can easily write few unit-tests for it as well.

rahulait:
Overall, LGTM. Just a minor suggestion to have a separate function for it and also add unit-tests for it.

cmd/nvidia-validator/main.go
cdesiniotis: +1. Let's please write unit tests in this PR.

cmd/nvidia-validator/main.go
```diff
@@ -749,7 +749,17 @@ func validateHostDriver(silent bool) error {
...
+	if target, err := os.Readlink("/host/usr/bin"); err == nil {
+		if filepath.IsAbs(target) {
+			nvidiaSMIPath = filepath.Join("/host", strings.TrimPrefix(target, "/"), "nvidia-smi")
+		} else {
+			nvidiaSMIPath = filepath.Join("/host/usr", target, "nvidia-smi")
```
cdesiniotis: I am not following this logic. For example, what happens if `/host/usr/bin/nvidia-smi` is a symlink to `../../nvidia-smi`?

cdesiniotis (commented):
I would recommend we use existing packages for symlink resolution within a root. For example, we could use moby's [FollowSymlinkInScope](https://pkg.go.dev/github.com/moby/sys/symlink#FollowSymlinkInScope) or https://pkg.go.dev/github.com/cyphar/filepath-securejoin. The former will probably suffice since the host path to `nvidia-smi` is assumed to be trustworthy in this context.

cmd/nvidia-validator/main.go
JunAr7112: I can try using symlinkinScope here. wrt the current format, if /usr/bin points to an absolute host path, it prefixes /host; if it is relative, it resolves it under /host/usr

cmd/nvidia-validator/main.go
JunAr7112: Added Test_validateHostDriverWithAbsoluteUsrBinSymlink(). For the unit test design, a variable hostRootControlPath was created that defaults to /host.

cdesiniotis: The original suggestion was to move the symlink resolution code to its own function. I would strongly prefer that we do that since 1) we resolve the path to the host's nvidia-smi in two places, and 2) it greatly simplifies the unit test code.

cmd/nvidia-validator/main_test.go
```diff
@@ -19,9 +19,96 @@ package main
 import (
 	"context"
 	"os"
+	"path/filepath"
+	"reflect"
 	"testing"
 )
 
+func Test_validateHostDriverWithAbsoluteUsrBinSymlink(t *testing.T) {
```
cdesiniotis: This feels overly complicated. As suggested in a prior comment, let's move the nvidia-smi resolution code to its own function and just write a unit test method for that code.

cdesiniotis: For reference, here is a method which tests path resolution in a root: https://github.com/NVIDIA/nvidia-container-toolkit/blob/2925ef90ea40767ada6599c03a1c6a949bae85ee/cmd/nvidia-cdi-hook/cudacompat/container-root_test.go#L300-L391

cmd/nvidia-validator/main.go
```diff
@@ -742,16 +743,32 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 	return false, nil
 }
 
+// resolveHostNvidiaSMI resolves and stats nvidia-smi within the mounted host root.
+func resolveHostNvidiaSMI(hostRootCtrPath string) (os.FileInfo, error) {
+	nvidiaSMIPath, err := symlink.FollowSymlinkInScope(filepath.Join(hostRootCtrPath, "usr/bin/nvidia-smi"), hostRootCtrPath)
```
tariq1890: @cdesiniotis Should we use `cyphar.SecureJoin` here instead?

cmd/nvidia-validator/main_test.go
JunAr7112: Ok I updated to realign the format more to match this

cmd/nvidia-validator/main.go
JunAr7112: I created a helper function resolveHostNvidiaSMI(hostRootCtrPath string)

cmd/nvidia-validator/main.go
JunAr7112: I think cyphar.SecureJoin is legacy code. libpathrs is recommended instead. Would that be needed for this use case though? It would be a larger dependency

tariq1890: Can you elaborate?

JunAr7112: This library was originally just an implementation of SecureJoin which was https://github.com/golang/go/issues/20126 as a safer filepath.Join that would restrict the path lookup to be inside a root directory.

The implementation was based on code that existed in several container runtimes. Unfortunately, this API is fundamentally unsafe against attackers that can modify path components after SecureJoin returns and before the caller uses the path, allowing for some fairly trivial TOCTOU attacks.

https://github.com/cyphar/filepath-securejoin

In the readme they were recommending libpathrs. I think it might be easier to use a smaller dependency though, unless we need anything else from libpathrs

cdesiniotis: As mentioned in https://github.com/NVIDIA/gpu-operator/pull/2464#pullrequestreview-4312140793, either should work. Using `pathrs` is a more secure approach since it only operates on file descriptors. In contrast `symlink.FollowSymlinkInScope` returns a path which must be resolved again when opening / executing the file. Since the host's root filesystem is a "trusted" filesystem in this context, I am not too concerned with using `symlink.FollowSymlinkInScope`.

However, with that said, @JunAr7112 let's switch to using https://github.com/cyphar/filepath-securejoin/tree/main/pathrs-lite in this PR. This is more secure and aligns with the work we recently did in the toolkit. The pseudocode would be:

```
f, _ := pathrs.OpenInRoot("/host", "/usr/bin/nvidia-smi")
fInfo, _ := f.Stat()
. . .
```

JunAr7112: Alright, I can switch to using pathrs for more consistency

JunAr7112: Ok @cdesiniotis I updated the MR to use pathrs

cmd/nvidia-validator/main.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/NVIDIA/go-nvlib/pkg/nvmdev"
 	"github.com/NVIDIA/go-nvlib/pkg/nvpci"
 	devchar "github.com/NVIDIA/nvidia-container-toolkit/cmd/nvidia-ctk/system/create-dev-char-symlinks"
+	pathrs "github.com/cyphar/filepath-securejoin/pathrs-lite"
```
tariq1890: This is redundant. The external module name is already `pathrs`
```suggestion
	"github.com/cyphar/filepath-securejoin/pathrs-lite"
```

cmd/nvidia-validator/main.go
```diff
@@ -742,16 +743,33 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
 	return false, nil
 }
 
+// resolveHostNvidiaSMI opens and stats nvidia-smi within the mounted host root.
+func resolveHostNvidiaSMI(hostRootCtrPath string) (os.FileInfo, error) {
+	f, err := pathrs.OpenInRoot(hostRootCtrPath, "/usr/bin/nvidia-smi")
```
tariq1890, line 748: Should we also attempt paths like `/usr/sbin/nvidia-smi` and `/usr/lib/wsl/lib/nvidia-smi`?

cc @cdesiniotis

cmd/nvidia-validator/main.go
JunAr7112: removed this

cmd/nvidia-validator/main.go
JunAr7112: @tariq1890 In what systems would we see /usr/sbin/nvidia-smi and /usr/lib/wsl/lib/nvidia-smi symlinks? I think the original error was wrt NixOS systems, but I can generalize to include this as well if its a common bug

tariq1890: `/usr/lib/wsl/lib/nvidia-smi` is used in WSL. I believe Talos and other similar distros use `/usr/sbin/nvidia-smi`

cmd/nvidia-validator/main.go
```diff
@@ -742,16 +743,33 @@ func isDriverManagedByOperator(ctx context.Context) (bool, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to open 'nvidia-smi' on the host: %w", err)
+	}
+	defer func() { _ = f.Close() }()
```
cdesiniotis:
```suggestion
	defer f.Close()
```

cmd/nvidia-validator/main.go
```diff
@@ -645,7 +646,7 @@ func validateComponent(ctx context.Context, componentFlag string) error {
 	}
 }
 
-func runCommand(command string, args []string, silent bool) error {
+var runCommand = func(command string, args []string, silent bool) error {
```
cdesiniotis: Let's revert this change since it is no longer needed.

cmd/nvidia-validator/main.go
cdesiniotis: I think we should expand the search paths here, but we could do that in a follow-up. For reference, we have precedent for searching for nvidia-smi in multiple directories, e.g. https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/510086482c09c9cdde2886c292fc0006fc9df926/hack/kubelet-plugin-prestart.sh#L44-L52

cmd/nvidia-validator/main_test.go
```diff
@@ -19,9 +19,65 @@ package main
...
+		expectsError bool
+	}{
+		{
+			description: "nvidia-smi exists in /usr/bin",
+			contents: map[string]string{
+				"/usr/bin/nvidia-smi": "fake nvidia-smi",
+			},
+		},
+		{
+			description: "nvidia-smi exists through absolute /usr/bin symlink",
```
cdesiniotis, line 44: Can we add additional test cases for 1) a relative symlink, and 2) a symlink which escapes the root?

cmd/nvidia-validator/main.go
JunAr7112: Reverted to func

cmd/nvidia-validator/main_test.go
JunAr7112: Updated the test cases

cmd/nvidia-validator/main_test.go
```diff
@@ -19,9 +19,88 @@ package main
...
+		{
+			description: "nvidia-smi symlink escapes root",
+			contents: map[string]string{
+				"/usr/bin": "symlink=../../outside/bin",
+			},
+			setup: func(t *testing.T, hostRoot string) {
+				outsideDir := filepath.Join(filepath.Dir(hostRoot), "outside", "bin")
+				require.NoError(t, os.MkdirAll(outsideDir, 0755))
+				require.NoError(t, os.WriteFile(filepath.Join(outsideDir, "nvidia-smi"), []byte("fake nvidia-smi"), 0600))
+			},
```
cdesiniotis: Do we really need this? Note, the target of the symlink does not need to exist for this test case. I would vote to remove this.

Also, isn't the directory your are creating, `filepath.Join(filepath.Dir(hostRoot), "outside", "bin") `, within the root?

JunAr7112: Ok I simplified the format for this case to just use "/usr/bin":    "symlink=../../", to represent the symlink target being outside the root case and got rid of setup.

For the older format, filepath.Join(filepath.Dir(hostRoot), "outside", "bin") would yield tmp/outside/bin which would be parallel to the hostRoot: /tmp/TestResolveHostNvidiaSMI.../, not within it. This format would've been more explicit with having a folder outside the root but is probably not strictly necessary.

cdesiniotis (approved):

cdesiniotis:
@tariq1890 can you re-review this?

PR #2485, opened 2026-05-26
ci: skip multi-arch manifest publish when push is disabled

fallintoplace:
## Summary
- skip the multi-arch manifest job when `push_on_build` is `false`

## Why
The per-arch image builds already honor `push_on_build` through the build output settings, but the final manifest job still runs unconditionally.

That leaves us with an inconsistent flow in non-push contexts: the workflow can skip pushing the arch images and then still try to create and push a multi-arch manifest that references images that were never published.

This showed up after the CI workflow split introduced `push_on_build` into the modular workflows, but the manifest job never started consuming that signal.

## Validation
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/image-builds.yaml")'`

I did not run the workflow end-to-end locally.

### DIFF
--- a/.github/workflows/image-builds.yaml
+++ b/.github/workflows/image-builds.yaml
@@ -141,6 +141,7 @@ jobs:
 
   build-multi-arch-images:
     needs: [variables, build-gpu-operator-arm64, build-gpu-operator-amd64]
+    if: needs.variables.outputs.push_on_build == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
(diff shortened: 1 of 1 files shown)


kvalliyurnatt (dismissed):

kvalliyurnatt:
@fallintoplace can you please check and make sure you have signed your commit

fallintoplace:
@kvalliyurnatt Let me check real quick. Sorry for that.

kvalliyurnatt (approved):

tariq1890 (approved):

tariq1890:
Thank you for you contribution @fallintoplace !

PR #2507, opened 2026-05-30
feat: add ExtraVolumes and ExtraVolumeMounts to all daemonset specs

### LINKED ISSUE #1532: Ability to add extra volumes and extra volume mounts to daemonsets
We would like to manage daemonsets like `gpu-feature-discovery` and `dcgm-exporter` on GKE via the gpu-operator where the driver is installed in a non standard location `/home/kubernetes/bin/nvidia/`

We do not want to use the `device-plugin` and `container-toolkit` via the gpu-operator because 
- GKE NAP does not allow labelling nodes to disable the GKE's default device plugin 
- Just enabling container-toolkit via the gpu-operator can cause race conditions where GKE's device plugin registers devices before gpu-operator's toolkit gets chance to run

Sadly this means we have to play by rules of GKE's `runc` which calls GKE's own `nvidia-container-cli` variant which does not inject devices, nvidia-smi and libnvml to `gpu-feature-discovery` and `dcgm-exporter` since these pods do not ask for any gpu devices using `resources.limits`

Ability to add volumes and mounts will allow to bypass this limitation by manually exposing devices, binaries and libraries

```
      volumes:
        - name: dev
          hostPath:
            path: /dev
            type: ''
        - name: nvidia-install-dir-host
          hostPath:
            path: /home/kubernetes/bin/nvidia
            type: ''
        - name: nvidia-config
          hostPath:
            path: /etc/nvidia
            type: ''
      containers:
        - name: nvidia-dcgm-exporter
          volumeMounts:
            - name: dev
              mountPath: /dev
            - name: nvidia-install-dir-host
              mountPath: /usr/local/nvidia
            - name: nvidia-config
              mountPath: /etc/nvidia
          env:
            - name: DCGM_EXPORTER_LISTEN
              value: ':9400'
            - name: DCGM_EXPORTER_KUBERNETES
              value: 'true'
            - name: DCGM_EXPORTER_COLLECTORS
              value: /etc/dcgm-exporter/dcp-metrics-included.csv
            - name: NODE_NAME
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: spec.nodeName
            - name: DCGM_EXPORTER_KUBERNETES_GPU_ID_TYPE
              value: uid
            - name: NVIDIA_INSTALL_DIR_HOST
              value: /home/kubernetes/bin/nvidia
            - name: NVIDIA_INSTALL_DIR_CONTAINER
              value: /usr/local/nvidia
            - name: LD_LIBRARY_PATH
              value: >-
                /usr/local/nvidia/lib64:/usr/local/nvidia/lib:/usr/local/cuda/lib64
```

75asu:
Resolves #1532

cc @guptaNswati -- you added the `help wanted` + `feature` labels and invited a PR. Here it is.

## Description

Adds `ExtraVolumes` (`[]corev1.Volume`) and `ExtraVolumeMounts` (`[]corev1.VolumeMount`) fields to every daemonset-managing component spec, so ClusterPolicy users can mount arbitrary host paths or other volume sources into the operator-managed daemonsets. The motivating case (issue #1532) is GKE, where NVIDIA driver libraries live at a non-standard path (`/home/kubernetes/bin/nvidia/`) and have to be exposed to `dcgm-exporter` and `gpu-feature-discovery` pods that don't request a GPU resource.

The PR is "full sweep" across all 16 daemonset specs in `ClusterPolicySpec` -- not just the two mentioned in the issue body -- for consistency:

`DriverSpec`, `ToolkitSpec`, `DevicePluginSpec`, `SandboxDevicePluginSpec`, `KataDevicePluginSpec`, `DCGMExporterSpec`, `DCGMSpec`, `NodeStatusExporterSpec`, `GPUFeatureDiscoverySpec`, `MIGManagerSpec`, `KataManagerSpec`, `CCManagerSpec`, `VFIOManagerSpec`, `VGPUManagerSpec`, `VGPUDeviceManagerSpec`, `ValidatorSpec`.

### How it works

A single helper `applyExtraVolumes(obj, vols, mounts)` is added near the existing `addExtraAnnotations` helper. Each `Transform*` daemonset function gets one new line at the end, after `applyHostNetworkConfig`, calling the helper with the matching `config.<Component>.ExtraVolumes` / `ExtraVolumeMounts`. Mounts are appended to every container in the pod spec; init containers are intentionally left untouched (they are operator-owned setup steps, not user-facing application containers).

Two transforms share a parent spec by design:
- `TransformMPSControlDaemon` reads from `DevicePluginSpec` (same as `TransformDevicePlugin`)
- `TransformSandboxValidator` reads from `ValidatorSpec` (same as `TransformValidator`)

Both are commented inline to make the sharing explicit.

### Example usage after this change

```yaml
apiVersion: nvidia.com/v1
kind: ClusterPolicy
spec:
  dcgmExporter:
    extraVolumes:
      - name: nvidia-install-dir-host
        hostPath: { path: /home/kubernetes/bin/nvidia }
    extraVolumeMounts:
      - name: nvidia-install-dir-host
        mountPath: /usr/local/nvidia
```

### Note on diff size

The CRD yaml files (`config/crd/bases/`, `bundle/manifests/`, `deployments/gpu-operator/crds/`) gain ~36k lines each. That is the expected expansion: `corev1.Volume` is a tagged union over ~30 volume source types, and the OpenAPI schema for it inlines all of them. Multiplied across 16 specs and mirrored to three CRD locations, the line count grows fast. All three yaml files are output of `make manifests` + `make sync-crds`; only the four hand-written files matter for review:
- `api/nvidia/v1/clusterpolicy_types.go` -- the 32 field declarations
- `controllers/object_controls.go` -- the helper + 18 call sites
- `controllers/transforms_test.go` -- new test case for `TransformDCGMExporter`
- `api/nvidia/v1/zz_generated.deepcopy.go` -- autogen from controller-gen `object`

## Checklist

- [x] Lint checks passing (`make lint`) -- 0 issues
- [x] Go mod artifacts in-sync (`make validate-modules`) -- all modules verified

## Testing

- New sub-case `transform_dcgm_exporter_with_extra_volumes_and_volume_mounts` added to `TestTransformDCGMExporter` exercising both `ExtraVolumes` (single hostPath volume) and `ExtraVolumeMounts` (single mount). Verifies the volume is appended to the pod spec and the mount is appended to every container.
- Full `controllers` test suite passes locally.
- The other 15 daemonsets follow the identical pattern, so the single test exercises the helper that every other transform calls. Happy to add per-component tests in a follow-up if maintainers prefer; wanted to keep this PR scoped.


cdesiniotis:
@75asu thanks for the contribution. However, adding the ability for users to specify arbitrary volumes / volume mounts for daemonsets deployed by GPU Operator is not something we are currently willing to accept. Arbitrary host path volume mounts opens up the security surface significantly. 

Note, we already support GKE and their custom driver installer: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/google-gke.html. In particular, the `hostPaths.driverInstallDir` field in ClusterPolicy should address your use case.

Closing this PR as not planned.

PR #2513, opened 2026-06-02
Add GPUClusterConfig CRD and controller for DRA-based stack

karthikvetrivel:
## 1. Overview

We introduce a new CRD named `GPUClusterConfig` and a new controller for reconciling it. Like `ClusterPolicy` today, it is a singleton, cluster-scoped CRD that configures the operands needed to enable GPUs in Kubernetes. `GPUClusterConfig` represents the new **DRA-based** software-enablement stack; it is an evolution of `ClusterPolicy`. 

## Change Log 

[9f08bec](https://github.com/NVIDIA/gpu-operator/pull/2513/commits/9f08bec547b52efde5eb6d9d423a202b246e5cff):
- Defined `GPUClusterConfig` Go types in `api/nvidia/v1alpha1`, cluster-scoped + singleton, with kubebuilder validation/default markers for every operand block. Wire `AddToScheme`. Generated the CRD manifest + deepcopy.
- Tested: `make manifests generate` produces the CRD yaml and deepcopy. `kubectl apply` the CRD succeeds. 

[73c9d30](https://github.com/NVIDIA/gpu-operator/pull/2513/commits/73c9d30ef7398c0db40ed6babaef053119607d85): 
- Introduced a new controller built on the existing `state.Manager` / `SyncState()` engine (the same
  pattern `NVIDIADriver` uses), registered in `cmd/gpu-operator/main.go`.
- Singleton enforcement (first-wins): a single instance owns reconciliation; any additional
  instance is marked `Ignored` and skipped. Mirrors how `ClusterPolicy` handles duplicates.
  
[ccc0f7a](https://github.com/NVIDIA/gpu-operator/pull/2513/commits/ccc0f7ae3c22c3aacb59219304c8992224b80792):
- Added cmd/dra-driver-validator, the init container binary for the DRA kubelet-plugin DaemonSet. It runs before the `gpus` and `computeDomains` containers start, validates that the NVIDIA driver is installed, and writes `/run/nvidia/validations/driver-ready` with the two env vars the kubelet-plugin containers source on startup (`NVIDIA_DRIVER_ROOT`, `DRIVER_ROOT_CTR_PATH`).
- Tested: unit tests end-to-end with fake driver, and against a real NVIDIA 595.58.03 driver on an A100 node.

[87fa6c0](https://github.com/NVIDIA/gpu-operator/pull/2513/commits/87fa6c0b8a0c0e5a23f1195258f736bc5fcde6e7): 
- Adds the DRA driver operand to the GPUClusterConfig controller (a new state-dra-driver state with its manifests wired into the state manager) 

[aff2736](https://github.com/NVIDIA/gpu-operator/pull/2513/commits/aff2736e22e41fed313d604df7720bc7f3c37661)
- Adds a Helm install path for the DRA stack and makes it mutually exclusive with ClusterPolicy. Setting `gpuClusterConfig.enabled=true` renders a new `deployments/gpu-operator/templates/gpuclusterconfig.yaml` that creates the singleton gpu-cluster-config CR (full draDriver spec exposed via values, reusing the shared hostPaths/daemonsets), and gates the clusterpolicy.yaml template off so the chart never deploys both paths at once.


api/nvidia/v1alpha1/gpuclusterconfig_types.go
```diff
@@ -0,0 +1,238 @@
...
+// +genclient
+// +genclient:nonNamespaced
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Cluster,shortName={"gcc"}
+//+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.state`,priority=0
+//+kubebuilder:printcolumn:name="Age",type=string,JSONPath=`.metadata.creationTimestamp`,priority=0
+
+// GPUClusterConfig is the Schema for the gpuclusterconfigs API
+type GPUClusterConfig struct {
```
rajathagasthya, line 205: Suggestion: how about calling this just `GPUCluster`? `Config` part is fairly obvious. It would also read well as `GPUClusterSpec` (specification for the GPU cluster) and `GPUClusterStatus` (status of the GPU cluster).

karthikvetrivel: I'm not opposed! Great point re: Spec, Status.

karthikvetrivel:
Superseded by #2571, which re-creates this with the head branch on `NVIDIA/gpu-operator` (instead of my fork) so it can serve as a base for stacked PRs. The naming discussion here (`GPUClusterConfig` → `GPUCluster`) is still tracked and will be addressed in #2571.

PR #2514, opened 2026-06-03
fix: drop containerd config mounts when NRI is enabled

frezbo:
When NRI is enabled there's no need to setup containerd config file path mounts as the installer is a no-op. This fixes issues on immutable hosts where the containerd paths are non-writable and prevents the NRI pod from coming up since kubelet tries to create mount-points for those.

Part of upstream discusssion at: https://github.com/NVIDIA/gpu-operator/discussions/2385

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1379,11 +1379,50 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n 
 func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, container *corev1.Container) error {
 	setContainerEnv(container, "RUNTIME", runtime)
 
-	if runtime == gpuv1.Containerd.String() {
-		// Set the runtime class name that is to be configured for containerd
-		setContainerEnv(container, "CONTAINERD_RUNTIME_CLASS", getRuntimeClassName(config))
+	// In NRI plugin mode the toolkit installer runs the noop runtime configurer
+	// (see nvidia-ctk-installer's NewConfigurer): it does not read or modify the
+	// container runtime configuration and does not restart the runtime. Mounting the
+	// runtime config files and socket is therefore unnecessary in this mode, and the
+	// hostPath mounts (e.g. /etc/containerd/conf.d with DirectoryOrCreate) break on
+	// read-only hosts such as Talos. Only the NRI socket mount (below) is required.
+	if config.CDI.IsNRIPluginEnabled() {
+		transformNRISocketMounts(obj, container)
+	} else {
+		if runtime == gpuv1.Containerd.String() {
+			// Set the runtime class name that is to be configured for containerd
+			setContainerEnv(container, "CONTAINERD_RUNTIME_CLASS", getRuntimeClassName(config))
+		}
+
+		if err := transformRuntimeConfigAndSocketMounts(obj, runtime, container); err != nil {
+			return err
+		}
 	}
 
+	return nil
+}
+
+// transformNRISocketMounts adds the required NRI socket environment variable and
+// sets up the NRI socket mount
+func transformNRISocketMounts(obj *appsv1.DaemonSet, container *corev1.Container) {
+	// setup mounts for the runtime NRI socket file
+	nriSocketFile := getContainerEnv(container, "NRI_SOCKET")
+	if nriSocketFile == "" {
+		nriSocketFile = DefaultRuntimeNRISocketFile
+	}
+	setContainerEnv(container, "NRI_SOCKET", path.Join(DefaultRuntimeNRISocketTargetDir, path.Base(nriSocketFile)))
+
+	nriVolMountSocketName := "nri-socket"
+	nriVolMountSocket := corev1.VolumeMount{Name: nriVolMountSocketName, MountPath: DefaultRuntimeNRISocketTargetDir}
+	container.VolumeMounts = append(container.VolumeMounts, nriVolMountSocket)
+
+	nriSocketVol := corev1.Volume{Name: nriVolMountSocketName, VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: path.Dir(nriSocketFile), Type: ptr.To(corev1.HostPathDirectoryOrCreate)}}}
+	obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, nriSocketVol)
+}
+
+// transformRuntimeConfigAndSocketMounts configures the toolkit container with the
+// mounts and environment required for the toolkit installer to update the container
+// runtime configuration (top-level config, drop-in config, and runtime socket).
+func transformRuntimeConfigAndSocketMounts(obj *appsv1.DaemonSet, runtime string, container *corev1.Container) error {
 	// For runtime config files we have top-level configs and drop-in files.
 	// These are supported as follows:
 	//   * Docker only supports top-level config files.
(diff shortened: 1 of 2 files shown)


frezbo:
Relates to: https://github.com/NVIDIA/nvidia-container-toolkit/pull/1861

controllers/object_controls.go
```diff
@@ -1384,6 +1384,43 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
 		setContainerEnv(container, "CONTAINERD_RUNTIME_CLASS", getRuntimeClassName(config))
 	}
```
cdesiniotis: Question -- do we need to continue setting these envvars when the NRI Plugin is enabled?

frezbo: I'm not sure, I'll do another set of test dropping them

frezbo: yes, not needed, moved down

cdesiniotis:
@tariq1890 requesting your review

controllers/transforms_test.go
```diff
@@ -558,11 +559,44 @@ func TestTransformForRuntime(t *testing.T) {
...
+			},
+			input: NewDaemonset().
+				WithContainer(corev1.Container{Name: "test-ctr"}),
+			expectedOutput: NewDaemonset().
+				WithHostPathVolume("nri-socket", filepath.Dir(DefaultRuntimeNRISocketFile), ptr.To(corev1.HostPathDirectoryOrCreate)).
+				WithContainer(corev1.Container{
+					Name: "test-ctr",
+					Env: []corev1.EnvVar{
+						{Name: "RUNTIME", Value: gpuv1.Containerd.String()},
+						{Name: "CONTAINERD_RUNTIME_CLASS", Value: DefaultRuntimeClass},
```
cdesiniotis: This should address the unit test failure
```suggestion
```

frezbo: ahh I missed this, will fixup

controllers/object_controls.go
```diff
@@ -1384,6 +1384,43 @@ func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec,
...
+			nriSocketFile = DefaultRuntimeNRISocketFile
+		}
+		setContainerEnv(container, "NRI_SOCKET", DefaultRuntimeNRISocketTargetDir+path.Base(nriSocketFile))
+
+		nriVolMountSocketName := "nri-socket"
+		nriVolMountSocket := corev1.VolumeMount{Name: nriVolMountSocketName, MountPath: DefaultRuntimeNRISocketTargetDir}
+		container.VolumeMounts = append(container.VolumeMounts, nriVolMountSocket)
+
+		nriSocketVol := corev1.Volume{Name: nriVolMountSocketName, VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: path.Dir(nriSocketFile), Type: ptr.To(corev1.HostPathDirectoryOrCreate)}}}
+		obj.Spec.Template.Spec.Volumes = append(obj.Spec.Template.Spec.Volumes, nriSocketVol)
```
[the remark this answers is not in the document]
tariq1890: @frezbo This is a valid comment. Can you address this?

controllers/object_controls.go
```diff
@@ -1379,11 +1379,48 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
 func transformForRuntime(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, runtime string, container *corev1.Container) error {
 	setContainerEnv(container, "RUNTIME", runtime)
 
+	// In NRI plugin mode the toolkit installer runs the noop runtime configurer
+	// (see nvidia-ctk-installer's NewConfigurer): it does not read or modify the
+	// container runtime configuration and does not restart the runtime. Mounting the
+	// runtime config files and socket is therefore unnecessary in this mode, and the
+	// hostPath mounts (e.g. /etc/containerd/conf.d with DirectoryOrCreate) break on
+	// read-only hosts such as Talos. Only the NRI socket mount (below) is required.
+	if config.CDI.IsNRIPluginEnabled() {
```
tariq1890, line 1388: I understand that you placed this conditional block here so that we could return early before we get to mounting the container runtime socket and config file. However, I think it's more intuitive if we place the container runtime socket and config file mount code in a conditional instead. So, we would explicitly skip that code block if the NRI Plugin is enabled.

NOTE: I also get that the pattern of returning early is used elsewhere in the code, but I'd like to move away from it.

controllers/object_controls.go
```diff
@@ -3804,7 +3825,8 @@ func transformDriverContainer(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicy
 }
 
 func createSecretEnvReference(ctx context.Context, ctrlClient client.Client, secretName string,
-	namespace string, container *corev1.Container) error {
+	namespace string, container *corev1.Container,
```
tariq1890: Is this diff relevant to the changes you are trying to implement?

frezbo: that my IDE, i can fix up

controllers/object_controls.go
frezbo: I can fix up, i'm also used to returning early

controllers/object_controls.go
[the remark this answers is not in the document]
frezbo: yes, I wasn't sure if existing code that got moved around needed fixing, i'll fix up

tariq1890:
Thank you very much @frezbo for your contribution here!

controllers/object_controls.go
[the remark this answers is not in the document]
tariq1890: Ahh, I didn't realise that this was existing code. In that case, this comment is not a blocker. I'll leave it to you to decide if this needs to be fixed this as a part of this PR or not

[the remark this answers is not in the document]
frezbo: fixed

controllers/object_controls.go
frezbo: fixed

controllers/object_controls.go
frezbo: fixed

controllers/object_controls.go
```diff
@@ -1379,11 +1379,46 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
 	}
 
+	return nil
+}
+
+// transformRuntimeConfigAndSocketMounts configures the toolkit container with the
+// mounts and environment required for the toolkit installer to update the container
+// runtime configuration (top-level config, drop-in config, and runtime socket). This
+// is not required when the NRI plugin is enabled, since the installer does not modify
+// the runtime configuration in that mode.
```
tariq1890: Let's leave out the parts that are more or less obvious.
```suggestion
// transformRuntimeConfigAndSocketMounts configures the toolkit container with the
// mounts and environment required for the toolkit installer to update the container
// runtime configuration (top-level config, drop-in config, and runtime socket).
```

controllers/object_controls.go
```diff
@@ -1379,11 +1379,46 @@ func TransformToolkit(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n
...
-		// Set the runtime class name that is to be configured for containerd
-		setContainerEnv(container, "CONTAINERD_RUNTIME_CLASS", getRuntimeClassName(config))
+	// In NRI plugin mode the toolkit installer runs the noop runtime configurer
+	// (see nvidia-ctk-installer's NewConfigurer): it does not read or modify the
+	// container runtime configuration and does not restart the runtime. Mounting the
+	// runtime config files and socket is therefore unnecessary in this mode, and the
+	// hostPath mounts (e.g. /etc/containerd/conf.d with DirectoryOrCreate) break on
+	// read-only hosts such as Talos. Only the NRI socket mount (below) is required.
+	if config.CDI.IsNRIPluginEnabled() {
+		// setup mounts for the runtime NRI socket file
```
tariq1890: Why not move L1389-L1401 to a private the same way as you have done in `transformRuntimeConfigAndSocketMounts`?

controllers/transforms_test.go
```diff
@@ -390,6 +391,7 @@ func TestTransformForRuntime(t *testing.T) {
 	testCases := []struct {
 		description    string
 		runtime        gpuv1.Runtime
+		clusterPolicy  *gpuv1.ClusterPolicySpec
```
tariq1890: Instead of modifying the test case struct here, you can add a new test case to the `TestTransformToolkit` method in `object_controls_test.go`. The test case struct over there already has the `ClusterPolicySpec` object as a struct field.

frezbo: TestTransformToolkit is in controllers/transforms_test.go, or should I add in object_controls_test.go 🤔

tariq1890: Whoops, my bad :). It is indeed in `transforms_test.go`. To be precise, I mean [here](https://github.com/NVIDIA/gpu-operator/blob/main/controllers/transforms_test.go#L849)

controllers/object_controls.go
frezbo: fixed

controllers/transforms_test.go
frezbo: done

tariq1890 (approved):

PR #2515, opened 2026-06-03
Prevent driver rollout caused by helm.sh/chart label change

rajathagasthya:
## Description

On a Helm chart upgrade, the operator copies the chart's `helm.sh/chart` label (which encodes the chart version) onto operand pod templates. The changed value alters the DaemonSet's controller revision hash, which the upgrade controller treats as a spec change and rolls the driver DaemonSet — disrupting GPU workloads even on patch upgrades where the driver itself is unchanged.

This change stops propagating `helm.sh/chart` to operand pod templates in both the ClusterPolicy and NVIDIADriver paths, while still applying it to the DaemonSet object metadata for chart traceability. To keep the label on the running pods without recreating them, the operator reconciles it directly onto the live operand and driver pods on each reconcile. The label is also injected into the NVIDIADriver CR so its pods are covered the same way as the ClusterPolicy-managed operands.

## Checklist

## Testing

Unit tests added/updated for the pod-template exclusion (both paths) and the live-pod label reconcilers (`TestApplyCommonDaemonsetMetadata`, `TestDriverPodTemplateExcludesChartLabel`, `TestReconcileOperandPodLabels`, `TestReconcileDriverPodLabels`). `go build`, `go vet`, and the full `controllers` and `internal/state` suites pass. `helm template` confirms the NVIDIADriver CR renders the `helm.sh/chart` label. Not yet validated on a live cluster.

### DIFF
--- a/controllers/clusterpolicy_controller.go
+++ b/controllers/clusterpolicy_controller.go
@@ -170,6 +170,12 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Keep the helm.sh/chart label on running operand pods current. Best-effort: a failure
+	// must not fail the reconcile, since the authoritative chart label lives on the DaemonSets.
+	if err := clusterPolicyCtrl.reconcileOperandPodLabels(ctx); err != nil {
+		r.Log.Error(err, "failed to reconcile chart label on operand pods; continuing")
+	}
+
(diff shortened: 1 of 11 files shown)


rajathagasthya:
Closing this draft. The pod-template label-exclusion approach special-cases a single label (`helm.sh/chart`), and we'd rather pursue a more general solution for non-disruptive patch upgrades. Will follow up with a revised design.

PR #2516, opened 2026-06-04
Support for precompiled GDRCopy

just1not2:
### Context 

Add a new `usePrecompiled` field to the GDRCopy CRD and corresponding spec in the GPU Operator.

When enabled, this option allows the operator to use precompiled GDRCopy driver images instead of building the gdrdrv kernel module on the target node.

### Motivation

Some environments rely on container images that already contain a prebuilt `gdrdrv` kernel module. These images are built specifically for a given kernel version and require the image tag to include a suffix matching the node's current kernel.

Introducing `usePrecompiled` provides an explicit way to opt into this workflow and enables the operator to select images that contain a kernel-specific precompiled `gdrdrv` module, avoiding the need for on-node compilation.

## Checklist

## Testing

This has been tested for months at Datadog Inc. (see [here](https://github.com/DataDog/gpu-operator/commit/fcf23499d03bb43ec2742aef9e6cd6f2edcc83c2))

### DIFF
--- a/api/nvidia/v1alpha1/nvidiadriver_types.go
+++ b/api/nvidia/v1alpha1/nvidiadriver_types.go
@@ -363,6 +363,12 @@ type GDRCopySpec struct {
 	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
 	Enabled *bool `json:"enabled,omitempty"`
 
+	// UsePrecompiled indicates if deployment of GDRCopy using pre-compiled modules is enabled
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors=true
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.displayName="Enable GDRCopy deployment using pre-compiled modules"
+	// +operator-sdk:gen-csv:customresourcedefinitions.specDescriptors.x-descriptors="urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	UsePrecompiled *bool `json:"usePrecompiled,omitempty"`
+
 	// GDRCopy diver image repository
 	// +kubebuilder:validation:Optional
 	Repository string `json:"repository,omitempty"`
@@ -599,6 +605,14 @@ func (d *GPUDirectStorageSpec) GetImagePath(osVersion string) (string, error) {
 	return image, nil
 }
 
+// UsePrecompiledDrivers returns true if usePrecompiled option is enabled in spec
+func (d *GDRCopySpec) UsePrecompiledDrivers() bool {
+	if d.UsePrecompiled == nil {
+		return false
+	}
+	return *d.UsePrecompiled
+}
+
 // GetImagePath returns the gdrcopy driver image path given the information
 // provided in GDRCopySpec and the osVersion passed as an argument.
 // The driver image path will be in the following format unless the spec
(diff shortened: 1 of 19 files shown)


deployments/gpu-operator/values.yaml
```diff
@@ -450,6 +450,7 @@ gds:
 
 gdrcopy:
   enabled: false
+  usePrecompiled: false
```
rahulait, line 453: This toggle is not added to clusterpolicy or nvidiadriver helm templates, so setting it has no effect.

controllers/object_controls.go
```diff
@@ -3425,9 +3425,14 @@ func resolveDriverTag(n ClusterPolicyController, driverSpec interface{}) (string
 		}
 	case *gpuv1.GDRCopySpec:
 		spec := driverSpec.(*gpuv1.GDRCopySpec)
-		image, err = gpuv1.ImagePath(spec)
-		if err != nil {
-			return "", err
+		if spec.UsePrecompiledDrivers() {
+			// use per kernel version tag
+			image = spec.Repository + "/" + spec.Image + ":" + spec.Version + "-" + n.currentKernelVersion
```
rahulait: This assumes repository, image and version are all separately specified. Existing configuration supports providing full image reference in image with empty repository/version.
https://github.com/NVIDIA/gpu-operator/blob/57752060b8cd83ffa4a54a58b2de093e48f8bb5e/api/nvidia/v1/clusterpolicy_types.go#L2054-L2083

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -599,6 +605,14 @@ func (d *GPUDirectStorageSpec) GetImagePath(osVersion string) (string, error) {
...
 }
 
+// UsePrecompiledDrivers returns true if usePrecompiled option is enabled in spec
+func (d *GDRCopySpec) UsePrecompiledDrivers() bool {
+	if d.UsePrecompiled == nil {
+		return false
+	}
+	return *d.UsePrecompiled
+}
+
```
rahulait, line 615: Nvidiadriver reconciler still marks any `usePrecompiled: true` plus enabled GDRCopy as unsupported before it reaches the new precompiled-image logic. Thus spec.gdrcopy.usePrecompiled: true cannot work for NVIDIADriver resources.
https://github.com/NVIDIA/gpu-operator/blob/57752060b8cd83ffa4a54a58b2de093e48f8bb5e/controllers/nvidiadriver_controller.go#L156

deployments/gpu-operator/values.yaml
just1not2: fixed here: https://github.com/NVIDIA/gpu-operator/pull/2516/changes/e5b7c379da8dafe228211376379abdf8714ad85b

controllers/object_controls.go
just1not2: fixed here: https://github.com/NVIDIA/gpu-operator/pull/2516/changes/a5c1f8adfb5598f169be5e829b2175ec032d18dd

api/nvidia/v1alpha1/nvidiadriver_types.go
just1not2: fixed here: https://github.com/NVIDIA/gpu-operator/pull/2516/changes/d2460580ec2e686ae34321a3197847fde8337855

tariq1890:
@just1not2 Thank you for your contributions and patience thus far.

After an internal discussion, we have decided to not accept this change. The main reason being that it may not be worth maintaining this feature given that GDRCopy v2.6 provides support for `mmap()` operations without the `gdrdrv` kernel module. This feature should be available to users so long as they are on **CUDA 13.3 + R610 (or later) + libgdrapi.so v2.6**. We want to encourage users to adopt this model as it removes the need to enable the gdrcopy sidecar container in the driver daemonset thus making operations way simpler down the line.

You can refer to the following links for the more information on this:

- https://docs.nvidia.com/cuda/archive/13.3.0/cuda-toolkit-release-notes/index.html#new-features
- https://github.com/NVIDIA/gdrcopy/releases/tag/v2.6

Issue #2353, opened 2026-04-15
Support migration from Clusterpolicy to NVIDADriver CR for driver section only


myeolenv:

rahulait:
PR has been merged. This is now implemented and released in v26.7.0 release. Marking the task as done.

Issue #2355, opened 2026-04-15
Enable NVIDIADriver default CR to co-exist with user defined NVIDIADriver CRs


myeolenv:
Enable NVIDIADriver default CR to co-exist with user defined NVIDIADriver CRs

Note: The default CR must be deleted, as its nodeSelector targets all nodes in the cluster and overrides the nodeSelectors defined in user-created NVIDIADriver CRs

rahulait:
We have added support for default nvidiadriver CR. This will be available starting v26.7.0. Marking the ticket as done.

PR #2518, opened 2026-06-04
add support for default nvidiadriver and migration from clusterpolicy to nvidiadriver

rahulait:
## Fixes:
https://github.com/NVIDIA/gpu-operator/issues/2353
https://github.com/NVIDIA/gpu-operator/issues/2355

## Summary

This PR adds support for a default `NVIDIADriver` CR and enables controlled migration from legacy `ClusterPolicy` driver management to `NVIDIADriver`-based driver management.

When `driver.nvidiaDriverCRD.enabled=true` and `driver.nvidiaDriverCRD.deployDefaultCR=true`, Helm renders a default `NVIDIADriver` CR. This CR is identified by `spec.default=true`:

```yaml
spec:
  default: true
```

The CR name is not semantically important. Helm renders it as `default`, but users can create an arbitrarily named `NVIDIADriver` and mark it as the fallback driver by setting `spec.default=true`.

## Behavior

- Non-default `NVIDIADriver` CRs own nodes selected by their `spec.nodeSelector`.
- The default `NVIDIADriver` acts as fallback for GPU nodes that do not match any non-default driver.
- If no `NVIDIADriver` has `spec.default=true`, fallback ownership is skipped.
- If more than one `NVIDIADriver` has `spec.default=true`, reconciliation fails closed.
- A default `NVIDIADriver` cannot define `spec.nodeSelector`; the fallback always applies to remaining GPU nodes that do not match non-default drivers.
- If `spec.default` is changed to `false`, that CR becomes a normal user-defined `NVIDIADriver` and participates in nodeSelector conflict detection.
- User-defined driver conflicts do not relabel nodes to the default driver or clear existing owner labels.
- `NVIDIADriver` node selectors cannot include the operator-managed owner label `nvidia.com/gpu-operator.driver.owner`.
- In `NVIDIADriver` mode, `NVIDIADriver` ownership conflicts mark `ClusterPolicy` as `notReady` and surface the conflict in its status conditions.

## Migration Support

This PR supports migration from `ClusterPolicy` driver management to `NVIDIADriver` driver management by:

- rendering a default `NVIDIADriver` from Helm values when CRD mode is enabled
- preserving legacy `ClusterPolicy` driver fields when `driver.nvidiaDriverCRD.enabled=false`
- routing GPU nodes to `NVIDIADriver` ownership through the operator-managed node label `nvidia.com/gpu.driver.owner`
- cleaning up `ClusterPolicy`-owned driver daemonsets after `NVIDIADriver` ownership takes over
- preserving running driver pods during migration by orphaning old `ClusterPolicy` driver daemonsets before `NVIDIADriver` rolls replacement pods

Migration requires the target GPU nodes to be matched by at least one `NVIDIADriver` CR. If `driver.nvidiaDriverCRD.enabled=true` but no matching `NVIDIADriver` exists, the controller has no driver owner to assign to those nodes, so fallback ownership is skipped until a matching CR is created.

## Upgrade Example

When upgrading from a release that does not contain this migration support, migration from `ClusterPolicy` driver management to `NVIDIADriver` driver management must be performed in two upgrades. This applies only to clusters that are currently using `ClusterPolicy` for driver management and are switching to `NVIDIADriver`. It applies to releases older than `v26.7.0`.

Clusters that are already using `NVIDIADriver` in an older release do not need this two-phase migration flow.

Starting with `v26.7.0`, the migration support is already present in the running controller. Clusters already running `v26.7.0` or newer can switch to `NVIDIADriver` mode directly by upgrading or updating configuration with `driver.nvidiaDriverCRD.enabled=true`, as long as the target GPU nodes are matched by `NVIDIADriver` CRs. The recommended path is to also set `driver.nvidiaDriverCRD.deployDefaultCR=true` so Helm renders the default fallback CR. Alternatively, users can create their own `NVIDIADriver` CRs and mark one as the default fallback with `spec.default=true`. See the `Behavior` section for details on default and non-default ownership.

Do not jump directly from an old release to the new release with `driver.nvidiaDriverCRD.enabled=true`. The old controller does not know about the controlled migration flow. If the old controller exits while the new release is already configured for `NVIDIADriver` mode, it can tear down all existing driver pods immediately before the new controller has a chance to take ownership.

Use this sequence instead:

1. Upgrade to the new release with `ClusterPolicy` driver management still enabled.

   Example Helm settings:

   ```yaml
   driver:
     enabled: true
     nvidiaDriverCRD:
       enabled: false
       deployDefaultCR: false
   ```

   This gets the new controller logic running while the existing `ClusterPolicy`-owned driver pods remain in place.

2. Upgrade again to the same new release, this time enabling `NVIDIADriver` mode.

   Example Helm settings:

   ```yaml
   driver:
     enabled: true
     nvidiaDriverCRD:
       enabled: true
       deployDefaultCR: true
   ```

   This renders the default `NVIDIADriver` and performs the controlled migration from `ClusterPolicy` ownership to `NVIDIADriver` ownership.

## Issues Found And Fixed

During testing, we found several edge cases:

- Treating the literal name `default` as special was too restrictive and could conflict with user-created CRs. The default driver is now identified by `spec.default=true` instead.
- If multiple default drivers existed, reconciliation could behave ambiguously. The controller now fails closed when multiple defaults are found.
- Allowing the default driver to carry a node selector made fallback behavior ambiguous. Default drivers now reject non-empty `spec.nodeSelector`.
- On user-driver nodeSelector conflicts, nodes could be relabeled away from their existing owner, causing existing driver pods to disappear. Owner assignment now preflights conflicts and preserves existing owner labels on conflict.
- User-supplied `spec.nodeSelector` could include `nvidia.com/gpu-operator.driver.owner` and override the operator-managed owner selector used by driver daemonsets. The controller, admission validation, and daemonset nodeSelector construction now reject or defensively preserve that invariant.
- `ClusterPolicy` reconciliation previously attempted `NVIDIADriver` owner assignment even when `NVIDIADriver` mode was disabled. Owner assignment and orphaned-pod migration are now gated on `driver.enabled=true` and `driver.nvidiaDriverCRD.enabled=true`.
- Moving default ownership is now a spec change through `spec.default`, so it bumps `.metadata.generation` and appears in the `kubectl get nvidiadriver` `Default` print column.
- Conflict messages did not identify the other conflicting drivers. They now include the first conflicting node and the conflicting `NVIDIADriver` names.

## E2E Coverage

The `NVIDIADriver` e2e flow now covers:

- Helm renders a default `NVIDIADriver` only when `driver.nvidiaDriverCRD.enabled=true` and `driver.nvidiaDriverCRD.deployDefaultCR=true`.
- Helm does not render a default CR when CRD mode is disabled.
- Helm does not render a default CR when `deployDefaultCR=false`.
- Operator installs in `NVIDIADriver` CRD mode.
- Default driver can use an arbitrary name as long as it has `spec.default=true`.
- User-defined `NVIDIADriver` can take ownership of GPU nodes.
- Driver image/version update works through a user-defined `NVIDIADriver`.
- Custom driver pod labels update through a user-defined `NVIDIADriver`.
- Setting `spec.default=false` makes that CR a normal driver and conflict detection applies.
- Conflicts preserve existing node owner labels.
- Multiple default drivers are rejected or fail closed without disrupting existing ownership.
- The ClusterPolicy e2e workflow now performs an in-place migration to `NVIDIADriver` mode before teardown. It verifies the legacy driver daemonset is removed, the existing legacy pod is orphaned, the default `NVIDIADriver` takes ownership, the orphaned pod is deleted by the upgrade flow within 5 minutes, and the replacement driver pod becomes ready.
- Helm uninstall remains covered by the existing e2e teardown.

## Checklist

## Testing

Added unit-tests, e2e tests and also did manual testing on a 3 node cluster with multiple nvidiadriver CRs. Everything worked as expected even with migration.

### DIFF
--- a/api/nvidia/v1alpha1/nvidiadriver_types.go
+++ b/api/nvidia/v1alpha1/nvidiadriver_types.go
@@ -36,10 +36,18 @@ const (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
-// NVIDIADriverSpec defines the desired state of NVIDIADriver
+// NVIDIADriverSpec defines the desired state of NVIDIADriver.
+// The CEL validation allows non-default drivers to use nodeSelector, but requires
+// default drivers to leave nodeSelector unset or empty.
+// +kubebuilder:validation:XValidation:rule="has(self.default) && self.default ? !has(self.nodeSelector) || size(self.nodeSelector) == 0 : true",message="default NVIDIADriver must not use nodeSelector"
 type NVIDIADriverSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+
+	// Default indicates that this NVIDIADriver acts as the fallback driver daemon set manager for GPU nodes
+	// that do not match any non-default NVIDIADriver nodeSelector.
+	// +kubebuilder:default=false
+	Default bool `json:"default"`
 
 	// +kubebuilder:validation:Enum=gpu;vgpu;vgpu-host-manager
 	// +kubebuilder:default=gpu
@@ -504,6 +512,7 @@ type NVIDIADriverStatus struct {
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:scope=Cluster,shortName={"nvd","nvdriver","nvdrivers"}
 //+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.state`,priority=0
+//+kubebuilder:printcolumn:name="Default",type=boolean,JSONPath=`.spec.default`,priority=0
 //+kubebuilder:printcolumn:name="Age",type=string,JSONPath=`.metadata.creationTimestamp`,priority=0
 
 // NVIDIADriver is the Schema for the nvidiadrivers API
@@ -522,6 +531,31 @@ type NVIDIADriverList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []NVIDIADriver `json:"items"`
+}
+
+// IsDefault returns true when the NVIDIADriver is marked as the fallback driver.
+func (d *NVIDIADriver) IsDefault() bool {
+	return d != nil && d.Spec.Default
+}
+
+// HasDeletionTimestamp returns true when the NVIDIADriver is marked for deletion.
+func (d *NVIDIADriver) HasDeletionTimestamp() bool {
+	return d != nil && !d.GetDeletionTimestamp().IsZero()
+}
+
+// ValidateNodeSelector rejects selectors that use operator-managed routing labels
+// or scope the default fallback driver.
+func (d *NVIDIADriver) ValidateNodeSelector() error {
+	if d == nil || d.Spec.NodeSelector == nil {
+		return nil
+	}
+	if d.IsDefault() && len(d.Spec.NodeSelector) > 0 {
+		return fmt.Errorf("default NVIDIADriver %q cannot use nodeSelector", d.Name)
+	}
+	if _, ok := d.Spec.NodeSelector[consts.NVIDIADriverOwnerLabel]; ok {
+		return fmt.Errorf("NVIDIADriver %q nodeSelector cannot use reserved label %q", d.Name, consts.NVIDIADriverOwnerLabel)
+	}
+	return nil
 }
 
 // UsePrecompiledDrivers returns true if usePrecompiled option is enabled in spec
--- a/controllers/nvidiadriver_controller.go
+++ b/controllers/nvidiadriver_controller.go
@@ -45,6 +45,7 @@ import (
 	"github.com/NVIDIA/gpu-operator/controllers/clusterinfo"
 	"github.com/NVIDIA/gpu-operator/internal/conditions"
 	"github.com/NVIDIA/gpu-operator/internal/consts"
+	nvidiadriverutil "github.com/NVIDIA/gpu-operator/internal/nvidiadriver"
 	"github.com/NVIDIA/gpu-operator/internal/state"
 	"github.com/NVIDIA/gpu-operator/internal/validator"
 )
@@ -83,9 +84,8 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if err := r.Get(ctx, req.NamespacedName, instance); err != nil {
 		if apierrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			return reconcile.Result{}, nil
+			// Re-run owner assignment so deleting the last NVIDIADriver clears stale node owner labels.
+			return r.cleanupNVIDIADriverOwnerLabels(ctx)
 		}
 		wrappedErr := fmt.Errorf("error getting NVIDIADriver object: %w", err)
 		logger.Error(err, "error getting NVIDIADriver object")
@@ -95,6 +95,10 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, wrappedErr
+	}
+	if instance.HasDeletionTimestamp() {
+		logger.Info("NVIDIADriver delete requested; cleaning up owner labels")
+		return r.cleanupNVIDIADriverOwnerLabels(ctx)
 	}
 
 	// Get the singleton NVIDIA ClusterPolicy object in the cluster.
@@ -152,6 +156,19 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return reconcile.Result{}, nil
 	}
 
+	changed, err := nvidiadriverutil.AssignOwners(ctx, r.Client)
+	if err != nil {
+		logger.Error(err, "failed to assign NVIDIADriver owners to nodes")
+		instance.Status.State = nvidiav1alpha1.NotReady
+		if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, err.Error()); condErr != nil {
+			logger.Error(condErr, "failed to set condition")
+		}
+		return reconcile.Result{}, err
+	}
+	if changed {
+		return reconcile.Result{RequeueAfter: time.Second}, nil
+	}
+
 	if instance.Spec.UsePrecompiledDrivers() && (instance.Spec.IsGDSEnabled() || instance.Spec.IsGDRCopyEnabled()) {
 		err := errors.New("GPUDirect Storage driver (nvidia-fs) and/or GDRCopy driver is not supported along with pre-compiled NVIDIA drivers")
 		logger.Error(err, "unsupported driver combination detected")
@@ -187,6 +204,11 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// Sync state and update status
 	managerStatus := r.stateManager.SyncState(ctx, instance, infoCatalog)
+
+	if err := r.labelNodesWithOrphanedDriverPods(ctx); err != nil {
+		logger.Error(err, "failed to label nodes with orphaned NVIDIA driver pods")
+		return reconcile.Result{}, err
+	}
 
 	// update CR status
 	if err := r.updateCrStatus(ctx, instance, managerStatus); err != nil {
@@ -276,6 +298,63 @@ func (r *NVIDIADriverReconciler) enqueueAllNVIDIADrivers(ctx context.Context) []
 	return reconcileRequests
 }
 
+// enqueueNVIDIADriverReconcilers enqueues the NVIDIADriver that triggered the
+// event and all current NVIDIADriver instances. The triggering object is
+// included even for delete events so the NotFound reconcile path can clear
+// stale node owner labels.
+func (r *NVIDIADriverReconciler) enqueueNVIDIADriverReconcilers(ctx context.Context, driver *nvidiav1alpha1.NVIDIADriver) []reconcile.Request {
+	requests := r.enqueueAllNVIDIADrivers(ctx)
+	if driver != nil {
+		requests = append(requests, reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      driver.GetName(),
+				Namespace: driver.GetNamespace(),
+			},
+		})
+	}
+	return dedupeReconcileRequests(requests)
+}
+
+func dedupeReconcileRequests(requests []reconcile.Request) []reconcile.Request {
+	seen := map[types.NamespacedName]struct{}{}
+	deduped := make([]reconcile.Request, 0, len(requests))
+	for _, request := range requests {
+		if _, ok := seen[request.NamespacedName]; ok {
+			continue
+		}
+		seen[request.NamespacedName] = struct{}{}
+		deduped = append(deduped, request)
+	}
+	return deduped
+}
+
+// cleanupNVIDIADriverOwnerLabels re-runs owner assignment after a delete event
+// when ClusterPolicy is still configured for NVIDIADriver mode. This lets
+// AssignOwners remove stale nvidia.com/gpu-operator.driver.owner labels when no remaining
+// NVIDIADriver matches a GPU node, including deletion of the last NVIDIADriver.
+func (r *NVIDIADriverReconciler) cleanupNVIDIADriverOwnerLabels(ctx context.Context) (reconcile.Result, error) {
+	clusterPolicyList := &gpuv1.ClusterPolicyList{}
+	if err := r.List(ctx, clusterPolicyList); err != nil {
+		wrappedErr := fmt.Errorf("error getting ClusterPolicy list: %w", err)
+		log.FromContext(ctx).Error(wrappedErr, "failed to cleanup NVIDIADriver owner labels")
+		return reconcile.Result{}, wrappedErr
+	}
+
+	if len(clusterPolicyList.Items) == 0 {
+		return reconcile.Result{}, nil
+	}
+
+	if !clusterPolicyList.Items[0].Spec.Driver.UseNvidiaDriverCRDType() {
+		return reconcile.Result{}, nil
+	}
+
+	if _, err := nvidiadriverutil.AssignOwners(ctx, r.Client); err != nil {
+		log.FromContext(ctx).Error(err, "failed to cleanup NVIDIADriver owner labels")
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *NVIDIADriverReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	// Create state manager
@@ -307,8 +386,8 @@ func (r *NVIDIADriverReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
 
 	// Watch for changes to NVIDIADriver CRs. Whenever an event is generated for a NVIDIADriver CR,
 	// enqueue a reconcile request for all NVIDIADriver instances.
-	nvidiaDriverMapFn := func(ctx context.Context, _ *nvidiav1alpha1.NVIDIADriver) []reconcile.Request {
-		return r.enqueueAllNVIDIADrivers(ctx)
+	nvidiaDriverMapFn := func(ctx context.Context, driver *nvidiav1alpha1.NVIDIADriver) []reconcile.Request {
+		return r.enqueueNVIDIADriverReconcilers(ctx, driver)
 	}
 
 	// Watch for changes to the primary resource NVIDIADriver
--- a/controllers/nvidiadriver_migration.go
+++ b/controllers/nvidiadriver_migration.go
@@ -1,0 +1,109 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/NVIDIA/k8s-operator-libs/pkg/upgrade"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	nvidiav1alpha1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1alpha1"
+	"github.com/NVIDIA/gpu-operator/internal/consts"
+)
+
+// labelNodesWithOrphanedDriverPods marks NVIDIADriver-owned nodes that still have orphaned
+// ClusterPolicy driver pods so the driver upgrade controller can replace those pods in
+// the normal controlled upgrade flow.
+func (r *NVIDIADriverReconciler) labelNodesWithOrphanedDriverPods(ctx context.Context) error {
+	nvidiaDrivers := &nvidiav1alpha1.NVIDIADriverList{}
+	if err := r.List(ctx, nvidiaDrivers); err != nil {
+		return fmt.Errorf("failed to list NVIDIADriver CRs: %w", err)
+	}
+	if len(nvidiaDrivers.Items) == 0 {
+		return nil
+	}
+
+	pods := &corev1.PodList{}
+	if err := r.List(ctx, pods,
+		client.InNamespace(r.Namespace),
+		client.MatchingLabels{AppComponentLabelKey: DriverAppComponentLabelValue},
+	); err != nil {
+		return fmt.Errorf("failed to list NVIDIA driver pods: %w", err)
+	}
+
+	for _, pod := range pods.Items {
+		if len(pod.OwnerReferences) > 0 || pod.Status.Phase != corev1.PodRunning || pod.Spec.NodeName == "" {
+			continue
+		}
+
+		node := &corev1.Node{}
+		if err := r.Get(ctx, types.NamespacedName{Name: pod.Spec.NodeName}, node); err != nil {
+			log.FromContext(ctx).Error(err, "failed to get node for orphaned NVIDIA driver pod", "pod", pod.Name, "node", pod.Spec.NodeName)
+			continue
+		}
+		if !nodeOwnedByNVIDIADriver(node, nvidiaDrivers.Items) {
+			continue
+		}
+
+		upgradeStateLabel := upgrade.GetUpgradeStateLabelKey()
+		upgradeState := upgrade.UpgradeStateUnknown
+		if node.Labels != nil {
+			upgradeState = node.Labels[upgradeStateLabel]
+		}
+		if !isDriverUpgradeRequestAllowed(upgradeState) {
+			continue
+		}
+		originalNode := node.DeepCopy()
+		if node.Labels == nil {
+			node.Labels = map[string]string{}
+		}
+		node.Labels[upgradeStateLabel] = upgrade.UpgradeStateUpgradeRequired
+		if err := r.Patch(ctx, node, client.MergeFrom(originalNode)); err != nil {
+			return fmt.Errorf("failed to label node %q with orphaned NVIDIA driver pod %q: %w", node.Name, pod.Name, err)
+		}
+	}
+
+	return nil
+}
+
+// isDriverUpgradeRequestAllowed returns true when migration can request a
+// driver upgrade without rewinding an active or failed upgrade state.
+func isDriverUpgradeRequestAllowed(upgradeState string) bool {
+	return upgradeState == upgrade.UpgradeStateUnknown || upgradeState == upgrade.UpgradeStateDone
+}
+
+// nodeOwnedByNVIDIADriver returns true when the node has an owner label for a live NVIDIADriver.
+func nodeOwnedByNVIDIADriver(node *corev1.Node, nvidiaDrivers []nvidiav1alpha1.NVIDIADriver) bool {
+	if node.Labels == nil || node.Labels[consts.NVIDIADriverOwnerLabel] == "" {
+		return false
+	}
+
+	for _, nvidiaDriver := range nvidiaDrivers {
+		if nvidiaDriver.HasDeletionTimestamp() {
+			continue
+		}
+		if node.Labels[consts.NVIDIADriverOwnerLabel] == nvidiaDriver.Name {
+			return true
+		}
+	}
+	return false
+}
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -4264,7 +4264,7 @@ func ocpHasDriverToolkitImageStream(n *ClusterPolicyController) (bool, error) {
 	return true, nil
 }
 
-func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context) error {
+func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context, propagationPolicy metav1.DeletionPropagation) error {
 	// Get all DaemonSets owned by ClusterPolicy
 	//
 	// (cdesiniotis) There is a limitation with the controller-runtime client where only a single field selector
@@ -4280,8 +4280,10 @@ func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context)
 		ds := ds
 		// filter out DaemonSets which are not the NVIDIA driver/vgpu-manager
 		if strings.HasPrefix(ds.Name, commonDriverDaemonsetName) || strings.HasPrefix(ds.Name, commonVGPUManagerDaemonsetName) {
-			n.logger.Info("Deleting NVIDIA driver daemonset owned by ClusterPolicy", "Name", ds.Name)
-			err = n.client.Delete(ctx, &ds)
+			n.logger.Info("Deleting NVIDIA driver daemonset owned by ClusterPolicy", "Name", ds.Name, "PropagationPolicy", propagationPolicy)
+			err = n.client.Delete(ctx, &ds, &client.DeleteOptions{
+				PropagationPolicy: &propagationPolicy,
+			})
 			if err != nil {
 				return fmt.Errorf("error deleting NVIDIA driver daemonset: %w", err)
 			}
--- a/controllers/state_manager.go
+++ b/controllers/state_manager.go
@@ -1051,7 +1051,7 @@ func (n *ClusterPolicyController) initOCPParams() error {
 func (n *ClusterPolicyController) step() (gpuv1.State, error) {
 	result := gpuv1.Ready
 
-	// Skip state-driver if NVIDIADriver CRD is enabled
+	// Skip driver daemonset states if NVIDIADriver CRD is enabled
 	// TODO:
 	//   - Properly clean up any k8s object associated with 'state-driver'
 	//     and owned by the Clusterpolicy controller.
@@ -1062,8 +1062,9 @@ func (n *ClusterPolicyController) step() (gpuv1.State, error) {
 		n.singleton.Spec.Driver.UseNvidiaDriverCRDType() {
 		n.logger.Info("NVIDIADriver CRD is enabled, cleaning up all NVIDIA driver daemonsets owned by ClusterPolicy")
 		n.idx++
-		// Cleanup all driver daemonsets owned by ClusterPolicy.
-		err := n.cleanupAllDriverDaemonSets(n.ctx)
+		// Cleanup all driver daemonsets owned by ClusterPolicy while keeping the
+		// running driver pods available until NVIDIADriver rolls replacements.
+		err := n.cleanupAllDriverDaemonSets(n.ctx, metav1.DeletePropagationOrphan)
 		if err != nil {
 			return gpuv1.NotReady, fmt.Errorf("failed to cleanup all NVIDIA driver daemonsets owned by ClusterPolicy: %w", err)
 		}
--- a/deployments/gpu-operator/templates/clusterpolicy.yaml
+++ b/deployments/gpu-operator/templates/clusterpolicy.yaml
@@ -167,6 +167,7 @@ spec:
   driver:
     enabled: {{ .Values.driver.enabled }}
     useNvidiaDriverCRD: {{ .Values.driver.nvidiaDriverCRD.enabled }}
+    {{- if not .Values.driver.nvidiaDriverCRD.enabled }}
     kernelModuleType: {{ .Values.driver.kernelModuleType }}
     usePrecompiled: {{ .Values.driver.usePrecompiled }}
     {{- if .Values.driver.repository }}
@@ -238,6 +239,7 @@ spec:
     {{- end }}
     {{- if .Values.driver.args }}
     args: {{ toYaml .Values.driver.args | nindent 6 }}
+    {{- end }}
     {{- end }}
     {{- if .Values.driver.upgradePolicy }}
     upgradePolicy:
--- a/internal/consts/consts.go
+++ b/internal/consts/consts.go
@@ -65,6 +65,11 @@ const (
 	// NVIDIADriverControllerIndexKey provides quick lookups for DaemonSets owned by an NVIDIADriver instance
 	NVIDIADriverControllerIndexKey = "metadata.nvidiadriver.controller"
 
+	// DefaultNVIDIADriverName is the Helm-managed fallback NVIDIADriver.
+	DefaultNVIDIADriverName = "default"
+	// NVIDIADriverOwnerLabel is an operator-managed node label used to route each GPU node to one NVIDIADriver.
+	NVIDIADriverOwnerLabel = "nvidia.com/gpu-operator.driver.owner"
+
 	// MinimumGDSVersionForOpenRM indicates the minimum GDS version that is supported only with OpenRM driver
 	MinimumGDSVersionForOpenRM = "v2.17.5"
 )
--- a/internal/nvidiadriver/nvidiadriver.go
+++ b/internal/nvidiadriver/nvidiadriver.go
@@ -1,0 +1,159 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package nvidiadriver
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nvidiav1alpha1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1alpha1"
+	"github.com/NVIDIA/gpu-operator/internal/consts"
+)
+
+// nodeMatchesSelector reports whether nodeLabels satisfy the NVIDIADriver nodeSelector.
+func nodeMatchesSelector(nodeLabels map[string]string, selector map[string]string) bool {
+	return labels.SelectorFromSet(selector).Matches(labels.Set(nodeLabels))
+}
+
+// AssignOwners labels GPU nodes with the NVIDIADriver that should manage their driver pods.
+// Non-default NVIDIADrivers take precedence over the default fallback, and conflicts fail closed before
+// node owner labels are changed. On success, it returns true when any node owner label was changed.
+func AssignOwners(ctx context.Context, c client.Client) (bool, error) {
+	drivers := &nvidiav1alpha1.NVIDIADriverList{}
+	if err := c.List(ctx, drivers); err != nil {
+		return false, fmt.Errorf("failed to list NVIDIADriver CRs: %w", err)
+	}
+
+	defaultDrivers, nonDefaultDrivers, err := classifyDrivers(drivers.Items)
+	if err != nil {
+		return false, err
+	}
+	if len(defaultDrivers) > 1 {
+		return false, fmt.Errorf("multiple default NVIDIADrivers found: %v", driverNames(defaultDrivers))
+	}
+	defaultOwner := ""
+	if len(defaultDrivers) == 1 {
+		defaultOwner = defaultDrivers[0].Name
+	}
+
+	nodes := &corev1.NodeList{}
+	if err := c.List(ctx, nodes, client.MatchingLabels{consts.GPUPresentLabel: "true"}); err != nil {
+		return false, fmt.Errorf("failed to list GPU nodes: %w", err)
+	}
+
+	desiredOwnersByNode := map[string]string{}
+	for _, node := range nodes.Items {
+		desiredOwner, err := desiredOwnerForNode(&node, nonDefaultDrivers, defaultOwner)
+		if err != nil {
+			return false, err
+		}
+		desiredOwnersByNode[node.Name] = desiredOwner
+	}
+
+	changed := false
+	for _, nodeItem := range nodes.Items {
+		node := nodeItem.DeepCopy()
+		desiredOwner := desiredOwnersByNode[node.Name]
+
+		if !ownerLabelNeedsUpdate(node.Labels, desiredOwner) {
+			continue
+		}
+
+		originalNode := node.DeepCopy()
+		if desiredOwner == "" {
+			delete(node.Labels, consts.NVIDIADriverOwnerLabel)
+		} else {
+			if node.Labels == nil {
+				node.Labels = map[string]string{}
+			}
+			node.Labels[consts.NVIDIADriverOwnerLabel] = desiredOwner
+		}
+
+		if err := c.Patch(ctx, node, client.MergeFrom(originalNode)); err != nil {
+			if desiredOwner == "" {
+				return false, fmt.Errorf("failed to remove NVIDIADriver owner label for node %q: %w", node.Name, err)
+			}
+			return false, fmt.Errorf("failed to update NVIDIADriver owner label for node %q: %w", node.Name, err)
+		}
+		changed = true
+	}
+
+	return changed, nil
+}
+
+// classifyDrivers filters out deleting drivers and splits live drivers by default fallback status.
+func classifyDrivers(drivers []nvidiav1alpha1.NVIDIADriver) ([]nvidiav1alpha1.NVIDIADriver, []nvidiav1alpha1.NVIDIADriver, error) {
+	defaultDrivers := []nvidiav1alpha1.NVIDIADriver{}
+	nonDefaultDrivers := make([]nvidiav1alpha1.NVIDIADriver, 0, len(drivers))
+	for _, driver := range drivers {
+		if driver.HasDeletionTimestamp() {
+			continue
+		}
+		if err := driver.ValidateNodeSelector(); err != nil {
+			return nil, nil, err
+		}
+		if driver.IsDefault() {
+			defaultDrivers = append(defaultDrivers, driver)
+			continue
+		}
+		nonDefaultDrivers = append(nonDefaultDrivers, driver)
+	}
+	return defaultDrivers, nonDefaultDrivers, nil
+}
+
+// driverNames returns the names of the provided drivers in their current order.
+func driverNames(drivers []nvidiav1alpha1.NVIDIADriver) []string {
+	names := make([]string, len(drivers))
+	for i, driver := range drivers {
+		names[i] = driver.Name
+	}
+	return names
+}
+
+// desiredOwnerForNode returns the non-default driver that should own the node, or the fallback owner when none match.
+func desiredOwnerForNode(
+	node *corev1.Node,
+	nonDefaultDrivers []nvidiav1alpha1.NVIDIADriver,
+	defaultOwner string,
+) (string, error) {
+	matchingDrivers := []string{}
+	for _, driver := range nonDefaultDrivers {
+		if nodeMatchesSelector(node.Labels, driver.GetNodeSelector()) {
+			matchingDrivers = append(matchingDrivers, driver.Name)
+		}
+	}
+	if len(matchingDrivers) > 1 {
+		return "", fmt.Errorf("multiple NVIDIADrivers match the same node %s: %v", node.Name, matchingDrivers)
+	}
+	if len(matchingDrivers) == 1 {
+		return matchingDrivers[0], nil
+	}
+	return defaultOwner, nil
+}
+
+// ownerLabelNeedsUpdate reports whether the node owner label differs from the desired owner.
+func ownerLabelNeedsUpdate(nodeLabels map[string]string, desiredOwner string) bool {
+	currentOwner, hasOwnerLabel := nodeLabels[consts.NVIDIADriverOwnerLabel]
+	if desiredOwner == "" {
+		return hasOwnerLabel
+	}
+	return !hasOwnerLabel || (currentOwner != desiredOwner)
+}
--- a/internal/state/driver.go
+++ b/internal/state/driver.go
@@ -56,8 +56,8 @@ const (
 
 	// AppComponentLabelKey indicates the label key of the component
 	AppComponentLabelKey = "app.kubernetes.io/component"
-	// AppComponentLabelValue indicates the label values of the nvidia-gpu-driver component
-	AppComponentLabelValue = "nvidia-driver"
+	// DriverAppComponentLabelValue indicates the label value of the NVIDIA driver component
+	DriverAppComponentLabelValue = "nvidia-driver"
 )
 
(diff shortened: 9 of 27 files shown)


api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -36,11 +36,19 @@ const (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
-// NVIDIADriverSpec defines the desired state of NVIDIADriver
+// NVIDIADriverSpec defines the desired state of NVIDIADriver.
+// The CEL validation allows non-default drivers to use nodeSelector, but requires
+// default drivers to leave nodeSelector unset or empty.
+// +kubebuilder:validation:XValidation:rule="has(self.default) && self.default ? !has(self.nodeSelector) || size(self.nodeSelector) == 0 : true",message="default NVIDIADriver cannot use nodeSelector"
```
rajathagasthya:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.default) && self.default ? !has(self.nodeSelector) || size(self.nodeSelector) == 0 : true",message="default NVIDIADriver must not use nodeSelector"
```

controllers/nvidiadriver_default.go
```diff
@@ -0,0 +1,140 @@
...
+
+	for i := range nodes.Items {
+		matchingDrivers := []string{}
+		for _, driver := range specificDrivers {
+			if nodeMatchesSelector(&nodes.Items[i], driver.GetNodeSelector()) {
+				matchingDrivers = append(matchingDrivers, driver.Name)
+			}
+		}
+		if len(matchingDrivers) > 1 {
+			return fmt.Errorf("conflicting NVIDIADriver NodeSelectors found for node %s: %s", nodes.Items[i].Name, strings.Join(matchingDrivers, ", "))
```
rajathagasthya: nit:
```suggestion
			return fmt.Errorf("multiple NVIDIADrivers match the same node %s: %v", nodes.Items[i].Name, matchingDrivers)
```
Just keep the `%v` for `matchingDrivers`, so it's formatted as `[driver-1 driver-2]` in the error string.

controllers/nvidiadriver_controller.go
```diff
@@ -152,6 +159,15 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return reconcile.Result{}, nil
 	}
 
+	if err := assignNVIDIADriverOwners(ctx, r.Client); err != nil {
```
rajathagasthya: Should we return and requeue after a successful `assignNVIDIADriverOwners` call to ensure `Node` objects later in the reconcile has owner labels?

For example: we label nodes in `assignNVIDIADriverOwners` and in the same reconcile we're eventually invoking `cleanupStaleDriverDaemonsets` which uses `Node` objects from the informer cache. Those objects in the cache wouldn't have the node label updates yet. Will it cause them to cleanup all the driver daemonsets?

controllers/object_controls.go
```diff
@@ -4291,6 +4278,76 @@ func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context)
...
+		}
+
+		upgradeStateLabel := upgrade.GetUpgradeStateLabelKey()
+		if node.Labels != nil && node.Labels[upgradeStateLabel] == upgrade.UpgradeStateUpgradeRequired {
+			continue
+		}
+		if node.Labels == nil {
+			node.Labels = map[string]string{}
+		}
+		node.Labels[upgradeStateLabel] = upgrade.UpgradeStateUpgradeRequired
```
rajathagasthya: What happens if the upgrade controller has moved the node label to `cordon-required`, or any of the intermediate states between `upgrade-required` and `upgrade-done`? This code overwrites it back to `upgrade-required` every time. It might disrupt an already in progress upgrade.

Should we set `upgrade-required` only when this label is absent instead of overwriting an existing state?

controllers/clusterpolicy_controller.go
```diff
@@ -143,6 +144,20 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	clusterPolicyCtrl.operatorMetrics.reconciliationTotal.Inc()
+
+	nvidiaDriverEnabled := nvidiaDriverCRDEnabled(instance)
+	if nvidiaDriverEnabled {
+		if err := assignNVIDIADriverOwners(ctx, r.Client); err != nil {
```
rajathagasthya: Any error here (such as conflicting NVIDIADriver nodeSelectors) will block the whole operation and no other operand gets processed until this is fixed. Are we fine with that? I feel like any error here should pause driver management only.

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+			continue
+		}
+		if node.Labels != nil && node.Labels[consts.NVIDIADriverOwnerLabel] == owner {
+			continue
+		}
+		if node.Labels == nil {
+			node.Labels = map[string]string{}
+		}
+		node.Labels[consts.NVIDIADriverOwnerLabel] = owner
```
rajathagasthya: Can these node labels go stale? For example: on a `helm uninstall` or if NFD reports `gpu.present=false`, are these labels cleaned up? Not sure if this is even possible, but just thinking about edge cases.

controllers/object_controls.go
```diff
@@ -4291,6 +4278,76 @@ func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context)
...
+		if err := n.client.Update(ctx, node); err != nil {
```
rajathagasthya: Can we use `Patch` instead of `Update`? There are a lot of components constantly updating `Node` objects, so we don't want to be stuck in conflict errors.

controllers/nvidiadriver_controller.go
```diff
@@ -120,8 +123,12 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 	clusterPolicyInstance := clusterPolicyList.Items[0]
 
+	if handled, err := r.handleDefaultNVIDIADriverDeletion(ctx, instance); handled || err != nil {
```
rajathagasthya: I'm curious why we need to handle deletion? On CR deletion, wouldn't we just return in L84-L88?

controllers/nvidiadriver_default.go
```diff
@@ -0,0 +1,140 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	gpuv1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1"
+	nvidiav1alpha1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1alpha1"
+	"github.com/NVIDIA/gpu-operator/internal/consts"
+)
+
+// isDefaultNVIDIADriver returns true when the NVIDIADriver is marked as the fallback driver.
+func isDefaultNVIDIADriver(driver *nvidiav1alpha1.NVIDIADriver) bool {
```
rajathagasthya: Can we use the same function from `internal/validator.go` to reduce duplication?

controllers/nvidiadriver_default.go
```diff
@@ -0,0 +1,140 @@
...
+// nvidiaDriverCRDEnabled returns true when ClusterPolicy driver management is enabled through NVIDIADriver CRs.
+func nvidiaDriverCRDEnabled(clusterPolicy *gpuv1.ClusterPolicy) bool {
+	return clusterPolicy != nil &&
+		clusterPolicy.Spec.Driver.IsEnabled() &&
+		clusterPolicy.Spec.Driver.UseNvidiaDriverCRDType()
+}
+
+// validateNVIDIADriverNodeSelector rejects selectors that use operator-managed routing labels
+// or scope the default fallback driver.
+func validateNVIDIADriverNodeSelector(driver *nvidiav1alpha1.NVIDIADriver) error {
```
rajathagasthya: Can we use the same function from `internal/validator.go` to reduce duplication?

rajathagasthya (commented):
Overall looks solid, just a few questions.

api/nvidia/v1alpha1/nvidiadriver_types.go
rahulait: Updated.

controllers/nvidiadriver_default.go
rahulait: Updated

controllers/nvidiadriver_controller.go
rahulait: Yeah, makes sense to return early if there is any assignment change. I've added a check to see if assignment changes and if it does, then return early with requeue.

controllers/object_controls.go
rahulait: Updated

controllers/nvidiadriver_controller.go
rahulait: Yeah, if its still in terminating/deleting state at L84-88, then it can end up here. Better approach would be to look at deletion timestamp. I've updated code to use that for all nvidiadrivers if they ever get past L84-L88

controllers/nvidiadriver_default.go
rahulait: Yes, makes sense. I've updated the code.

controllers/nvidiadriver_default.go
rahulait: Yes, makes sense. I've updated the code.

internal/nvidiadriver/nvidiadriver.go
rahulait: Helm uninstall won't remove the labels, so they remain on the node even after uninstall. helm pre-uninstall hook might not be useful as there are other workflows with which users can install gpu-operator, like openshift doesn't use helm. Even if the label remains, when gpu-operator is installed again, clusterpolicy will update the node label to right value. If its not supposed to have any label, it will remove it. So I think it should be fine if stale labels remain after uninstall. I'll check if we can remove them somehow.

deployments/gpu-operator/templates/clusterpolicy.yaml
```diff
@@ -167,6 +167,7 @@ spec:
   driver:
     enabled: {{ .Values.driver.enabled }}
     useNvidiaDriverCRD: {{ .Values.driver.nvidiaDriverCRD.enabled }}
+    {{- if not .Values.driver.nvidiaDriverCRD.enabled }}
```
cdesiniotis, line 170: Question -- what was the rationale for this change? I am assuming this was for aesthetic reasons? I could see it being easier on the eyes to NOT see these fields populated when NVIDIADriver CRD is enabled (since these fields in ClusterPolicy are no-ops).

controllers/clusterpolicy_controller.go
```diff
@@ -143,6 +145,24 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			updateCRState(ctx, r, req.NamespacedName, gpuv1.NotReady)
+			if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, err.Error()); condErr != nil {
+				r.Log.Error(condErr, "failed to set condition")
+			}
+			return ctrl.Result{}, err
+		}
+		if changed {
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+	}
```
cdesiniotis: I am trying to reason about why we are doing this. Does the below statement (from the PR description) capture the reasoning?
```
ClusterPolicy status did not consistently move to notReady for early NVIDIADriver ownership failures. It now sets status.state to notReady and records the concrete conflict message in status conditions.
```

It feels wrong to call `AssignOwners()` here in the ClusterPolicy controller when it is already being called in the NVIDIADriver controller (and I would argue it should only be called there). If the goal is to mark ClusterPolicy as `notReady` when any NVIDIADriver CR is `notReady` could we not check the status of all the NVIDIADriver CRs in the cluster instead? And even then, I question whether we want the ClusterPolicy status to take the NVIDIADriver CR statuses into account.

controllers/clusterpolicy_controller.go
```diff
@@ -371,6 +397,29 @@ func (r *ClusterPolicyReconciler) SetupWithManager(ctx context.Context, mgr ctrl
 		return err
 	}
 
+	err = c.Watch(
```
cdesiniotis: Just an observation -- it feels wrong for the ClusterPolicy controller to have to watch for updates to NVIDIADriver CRs.

controllers/nvidiadriver_controller.go
```diff
@@ -152,6 +157,19 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		logger.Error(err, "failed to assign NVIDIADriver owners to nodes")
+		instance.Status.State = nvidiav1alpha1.NotReady
+		if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, err.Error()); condErr != nil {
+			logger.Error(condErr, "failed to set condition")
+		}
+		return reconcile.Result{}, err
+	}
+	if changed {
+		return reconcile.Result{RequeueAfter: time.Second}, nil
+	}
```
cdesiniotis, line 170: Question -- why do we return here?

controllers/object_controls.go
```diff
@@ -4291,6 +4302,67 @@ func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context)
...
+		return fmt.Errorf("failed to list NVIDIADriver CRs: %w", err)
+	}
+	if len(nvidiaDrivers.Items) == 0 {
+		return nil
+	}
+
+	pods := &corev1.PodList{}
+	if err := n.client.List(ctx, pods,
+		client.InNamespace(n.operatorNamespace),
+		client.MatchingLabels{AppComponentLabelKey: AppComponentLabelValue},
```
cdesiniotis: nit: renaming `AppComponentLabelValue` to `DriverAppComponentLabelValue` improve readability.

controllers/object_controls.go
cdesiniotis: It feels like we should set `upgrade-required` only if the label is currently absent or set to `upgrade-done` / `upgrade-unknown` (aka an upgrade is not currently in progress).

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,153 @@
...
+	return driver != nil && driver.Spec.Default
+}
+
+// ValidateNodeSelector rejects selectors that use operator-managed routing labels
+// or scope the default fallback driver.
+func ValidateNodeSelector(driver *nvidiav1alpha1.NVIDIADriver) error {
+	if driver == nil || driver.Spec.NodeSelector == nil {
+		return nil
+	}
+	if IsDefault(driver) && len(driver.Spec.NodeSelector) > 0 {
```
cdesiniotis: Question -- should we also return an error if `!IsDefault(driver) && len(driver.Spec.NodeSelector) == 0`?

controllers/clusterpolicy_controller.go
```diff
@@ -371,6 +397,29 @@ func (r *ClusterPolicyReconciler) SetupWithManager(ctx context.Context, mgr ctrl
...
+						NamespacedName: types.NamespacedName{Name: item.GetName()},
+					})
+				}
+				return requests
+			}),
+			nvidiaDriverGenerationChangedPredicate(),
+		),
+	)
+	if err != nil {
+		return err
```
tariq1890: this error should be wrapped

deployments/gpu-operator/templates/clusterpolicy.yaml
rahulait: This change is to make ClusterPolicy stop carrying a second copy of the driver spec when the NVIDIADriver CR is supposed to be the source of truth.

controllers/object_controls.go
```diff
@@ -4264,7 +4267,7 @@ func ocpHasDriverToolkitImageStream(n *ClusterPolicyController) (bool, error) {
 	return true, nil
 }
 
-func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context) error {
+func (n ClusterPolicyController) cleanupAllDriverDaemonSets(ctx context.Context, orphanPods bool) error {
```
tariq1890: Instead of a boolean, let's just use the propagation policy as the argument

internal/nvidiadriver/nvidiadriver.go
rahulait: Empty nodeselector is allowed as it is same as no nodeselector. Hence, we are checking for > 0.

controllers/clusterpolicy_controller.go
cdesiniotis: @rahulait it looks like my comment only got added to L164. I meant to attach this to the entire code block.

internal/nvidiadriver/nvidiadriver.go
cdesiniotis: My comment was about non-default NVIDIADriver CRs. Don't we require that they do NOT have empty node selectors configured?

internal/consts/consts.go
```diff
@@ -65,6 +65,11 @@ const (
 	// NVIDIADriverControllerIndexKey provides quick lookups for DaemonSets owned by an NVIDIADriver instance
 	NVIDIADriverControllerIndexKey = "metadata.nvidiadriver.controller"
 
+	// DefaultNVIDIADriverName is the Helm-managed fallback NVIDIADriver.
+	DefaultNVIDIADriverName = "default"
+	// NVIDIADriverOwnerLabel is an operator-managed node label used to route each GPU node to one NVIDIADriver.
+	NVIDIADriverOwnerLabel = "nvidia.com/gpu.driver.owner"
```
tariq1890: If this is an operator managed label, let the name be suggestive of that. We could use either the `gpu-operator` or `<controller-name>` as the prefix.

controllers/clusterpolicy_controller.go
rahulait: In current implementation, clusterpolicy controller is responsible for adding driver ownership labels to nodes during migration from clusterpolicy to nvidiadriver. In case there is no nvidiadriver, or there is a conflict between nvidiadrivers and it is resolved, clusterpolicy reconciler should wait and then re-evaluate everything and add ownership labels. Hence, its keeping a watch on nvidiadriver CRs

internal/state/driver.go
```diff
@@ -270,7 +270,7 @@ func (s *stateDriver) getManifestObjects(ctx context.Context, cr *nvidiav1alpha1
 	}
 
 	isOpenshift := runtimeSpec.OpenshiftVersion != ""
-	nodePools, err := getNodePools(ctx, s.client, cr.Spec.NodeSelector, cr.Spec.UsePrecompiledDrivers(), isOpenshift)
+	nodePools, err := getNodePools(ctx, s.client, cr.Name, cr.Spec.NodeSelector, cr.Spec.UsePrecompiledDrivers(), isOpenshift)
```
tariq1890: Consider condensing the function signature here by just passing the CR object as the parameter. This is a private method, so it should be a safe change

controllers/clusterpolicy_controller.go
```diff
@@ -371,6 +397,29 @@ func (r *ClusterPolicyReconciler) SetupWithManager(ctx context.Context, mgr ctrl
 		return err
 	}
 
+	err = c.Watch(
+		source.Kind(mgr.GetCache(),
+			&nvidiav1alpha1.NVIDIADriver{},
+			handler.TypedEnqueueRequestsFromMapFunc(func(ctx context.Context, _ *nvidiav1alpha1.NVIDIADriver) []reconcile.Request {
```
tariq1890: Instead of dropping the parameter, should we use it to add a debug log? Perhaps, it would be useful to see which NVIDIADriver CR triggered a ClusterPolicy reconcile if needed.

controllers/nvidiadriver_controller.go
rahulait: Before this step, we patch the nodes with ownership labels. After patch, apiserver has the new labels immediately, but controller cache may not have the cache updated. If we continue in the same reconcile, getNodePools() can read stale node objects and think no nodes match this NVIDIADriver. Subsequent reconciles will fix things (adding label would have queued a reconcile event), but its better we just requeue and retry when node ownership changes.

controllers/clusterpolicy_controller.go
rahulait: I think we can get rid of this defensive approach if we think its cleaner. Right now, it does ownership transfer to nvidiadriver and then deletes daemonset with cascade delete orphan. If we think its better we delete with cascade orphan and nvidiadriver will set the ownership, thats also fine in that case.

controllers/clusterpolicy_controller.go
rahulait: These changes were for migration from clusterpolicy to nvidiadriver. I can see these are causing confusion, let me see if I can optimize them further and only let nvidiadriver add ownership labels.

internal/nvidiadriver/nvidiadriver.go
rahulait: Ah ok. A user can add a user-defined nvidiadriver which covers entire cluster if they want to. But I am ok to change that to say user-specified nvidiadriver should match a subset of nodes if that seems cleaner. Let me know your thoughts on that.

controllers/clusterpolicy_controller.go
rahulait: Regarding status field of clusterpolicy: how should we model ClusterPolicy status when driver management is moved to NVIDIADriver?
With NVIDIADriver enabled, ownership of the driver operand moves out of ClusterPolicy, while the remaining operands are still managed by ClusterPolicy. During a node driver upgrade, the driver pod is replaced first, and then other operands may restart or go through init again.
Do we want ClusterPolicy status to reflect only the operands it still owns, which may cause it to become Ready during driver uninstall/replacement and then flip back to NotReady when dependent operands restart? Or should ClusterPolicy continue to represent the combined health of the overall GPU Operator stack, including NVIDIADriver-managed driver state?

controllers/nvidiadriver_controller.go
```diff
@@ -316,7 +334,7 @@ func (r *NVIDIADriverReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
 		mgr.GetCache(),
 		&nvidiav1alpha1.NVIDIADriver{},
 		handler.TypedEnqueueRequestsFromMapFunc(nvidiaDriverMapFn),
-		predicate.TypedGenerationChangedPredicate[*nvidiav1alpha1.NVIDIADriver]{},
+		nvidiaDriverGenerationChangedPredicate(),
```
rajathagasthya: Why doesn't `TypedGenerationChangedPredicate` work here?

internal/consts/consts.go
rahulait: Sure, how about `nvidia.com/gpu-operator.driver.owner`?

controllers/nvidiadriver_controller.go
rahulait: No, it should work. In previous implementation when label was being used to denote default nvidiadriver, generation wasn't updated on label change, so needed that function. With newer implementation, we can just switch back to old code and remove this additional function. I'll remove it.

rahulait: Updated.

controllers/clusterpolicy_controller.go
rahulait: Updated.

internal/state/driver.go
rahulait: updated.

controllers/object_controls.go
rahulait: updated.

controllers/clusterpolicy_controller.go
rahulait: updated.

controllers/object_controls.go
rahulait: updated.

controllers/clusterpolicy_controller.go
rahulait: Looked further, this is needed because if a NVIDIADriver is created/updated after clusterpolicy has already reconciled,
we want to requeue clusterpolicy so it can label nodes with orphaned legacy driver pods with "upgrade-required" to have upgrade-controller upgrade them in controlled manner.

rajathagasthya: In a post-ClusterPolicy world (DRA with https://github.com/NVIDIA/gpu-operator/pull/2513), who is responsible for the owner assignment? I wonder if there should be a dedicated controller to do this?

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("default NVIDIADriver %q cannot use nodeSelector", driver.Name)
+	}
+	if _, ok := driver.Spec.NodeSelector[consts.NVIDIADriverOwnerLabel]; ok {
+		return fmt.Errorf("NVIDIADriver %q nodeSelector cannot use reserved label %q", driver.Name, consts.NVIDIADriverOwnerLabel)
+	}
+	return nil
+}
+
+// CRDEnabled returns true when ClusterPolicy driver management is enabled through NVIDIADriver CRs.
+func CRDEnabled(clusterPolicy *gpuv1.ClusterPolicy) bool {
```
tariq1890: This should be in the `clusterpolicy_types.go` and `DriverSpec` should be used as the receiver

controllers/clusterpolicy_controller.go
rahulait: nvidiadriver would be responsible for ownership assignment. Right now, just to trigger migration when nvidiadriver is added and clusterpolicy has already reconciled, we need to retrigger its reconciliation so that it can add upgrade_required labels to facilitate migration. But once we are in DRA, we wouldn't need this. So it can be dropped later on. This is required just for clusterpolicy -> nvidiadriver migration and that too the edge case when user didn't added any nvidiadriver initially but later on added it. Basically any time when clusterpolicy ran and didn't find any nvidiadriver and we need it to reconcile when nvidiadriver shows up.

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,153 @@
...
+// node owner labels are changed. It returns true when any node label was changed.
+func AssignOwners(ctx context.Context, c client.Client) (bool, error) {
+	drivers := &nvidiav1alpha1.NVIDIADriverList{}
+	if err := c.List(ctx, drivers); err != nil {
+		return false, fmt.Errorf("failed to list NVIDIADriver CRs: %w", err)
+	}
+
+	var defaultDriver *nvidiav1alpha1.NVIDIADriver
+	defaultDriverNames := []string{}
+	specificDrivers := make([]nvidiav1alpha1.NVIDIADriver, 0, len(drivers.Items))
```
tariq1890: Can we think of a different name for non-default `NVIDIADrivers` than "specificDrivers"?

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,153 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	gpuv1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1"
+	nvidiav1alpha1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1alpha1"
+	"github.com/NVIDIA/gpu-operator/internal/consts"
+)
+
+// IsDefault returns true when the NVIDIADriver is marked as the fallback driver.
+func IsDefault(driver *nvidiav1alpha1.NVIDIADriver) bool {
```
tariq1890: This method can be moved to `nvidiadriver_types.go` with `NVIDIADriver` as the receiver

internal/validator/validator_test.go
```diff
@@ -73,6 +74,12 @@ func nodeSelector(labels map[string]string) driverOptions {
 	}
 }
 
+func defaultDriver() driverOptions {
+	return func(c *nvidiav1alpha1.NVIDIADriver) {
```
tariq1890: Let's avoid closures unless it's really necessary.

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	if len(defaultDriverNames) > 1 {
+		return false, fmt.Errorf("multiple default NVIDIADrivers found: %s", strings.Join(defaultDriverNames, ", "))
+	}
+	nodes := &corev1.NodeList{}
+	if err := c.List(ctx, nodes, client.MatchingLabels{consts.GPUPresentLabel: "true"}); err != nil {
+		return false, fmt.Errorf("failed to list GPU nodes: %w", err)
+	}
+
+	for i := range nodes.Items {
```
tariq1890: The `AssignOwners` method will be called during every reconcile of a NVIDIADriver CR and each reconciler involves a loop through the entire list of gpu nodes. How will the performance be in large clusters ?

tests/scripts/update-nvidiadriver.sh
```diff
@@ -68,20 +249,86 @@ test_custom_labels_override() {
...
+    unset_default_driver "${DEFAULT_NVIDIA_DRIVER_NAME}"
+    wait_for_nvidiadriver_condition_message "${NVIDIA_DRIVER_NAME}" "multiple NVIDIADrivers match the same node"
+    wait_for_nvidiadriver_condition_message "${DEFAULT_NVIDIA_DRIVER_NAME}" "multiple NVIDIADrivers match the same node"
+    assert_nvidiadriver_owner_count "${NVIDIA_DRIVER_NAME}"
+    set_default_driver "${DEFAULT_NVIDIA_DRIVER_NAME}"
+    wait_for_default_nvidiadriver "${DEFAULT_NVIDIA_DRIVER_NAME}"
+    wait_for_nvidiadriver_owner "${NVIDIA_DRIVER_NAME}"
+}
+
+test_multiple_default_drivers() {
```
tariq1890: I am questioning if we need to test this in the e2e test. Won't a unit test be sufficient for this test case?

controllers/clusterpolicy_controller.go
cdesiniotis: >Looked further, this is needed because if a NVIDIADriver is created/updated after clusterpolicy has already reconciled,
we want to requeue clusterpolicy so it can label nodes with orphaned legacy driver pods with "upgrade-required" to have upgrade-controller upgrade them in controlled manner.

Got it. 

Question -- couldn't the NVIDIADriver controller technically be able to do this? For example, a new NVIDIADriver CR is created and upon reconciling it, the NVIDIADriver controller iterates over all nodes that match its node selector and adds the `upgrad-required` label if an orphan driver pod is running there?

cdesiniotis: The NVIDIADriver controller already iterates over all nodes by calling `AssignOwners`, so I am wondering if this logic of setting the `upgrade-required` label can be taken care of there (and eliminating the need to duplicate the `AssignOwners` call in the ClusterPolicy controller).

rahulait: Yes, it could. Let me think more about this.

internal/validator/validator_test.go
rahulait: Updated.

internal/nvidiadriver/nvidiadriver.go
rahulait: Updated.

internal/nvidiadriver/nvidiadriver.go
rahulait: updated. Used `nonDefaultDrivers`

internal/nvidiadriver/nvidiadriver.go
rahulait: updated

internal/consts/consts.go
rahulait: Updated to use this label now.

controllers/object_controls.go
rahulait: Updated to only add label if there is no label or its `upgrade-done`

tests/scripts/update-nvidiadriver.sh
rahulait: I think thats a fair point, unit tests cover it, so we can remove this. I'll update it.

controllers/clusterpolicy_controller.go
rahulait: I think thats how it is even today. The driver pod also controls rollout of other operands when it comes up. So if driver is stuck, others get stuck too and this should be fine. Until the conflict is resolved, reconciles will not do anything.

tests/scripts/update-nvidiadriver.sh
rahulait: Updated.

controllers/clusterpolicy_controller.go
rahulait: updated. moved it to nvidiadriver controller

internal/nvidiadriver/nvidiadriver.go
rahulait: The loop is O(number of GPU nodes × number of nvidiadriver CRs), but the CPU cost is not the main issue. The main cost at large scale is listing/copying thousands of Node objects and patching changed nodes. In an already converged/synced cluster, patches are skipped, so impact should be modest, cost is mostly large cache reads. The worst case is convergence after rollout or selector changes where multiple nodes will be patched together, API writes and watch fan-out dominate. We can mitigate by reusing a single GPU-node list across validation/assignment and rate-limiting node patches per reconcile. I would consider it an enhancement we can do to ease load in case of large clusters.

rahulait: Created a task for this so that we can do enhancements in future PRs.

controllers/clusterpolicy_controller.go
rahulait: As of today and even with this PR, the clusterpolicy state keeps fluctuating when driver is upgraded by nvidiadriver. I have filed a separate task to fix it in a separate followup PR.

controllers/clusterpolicy_controller.go
cdesiniotis: Thank you @rahulait. Looks great!

cdesiniotis (approved):
LGTM. Thanks @rahulait for the hard work and patience on this one!

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -36,11 +36,19 @@ const (
...
-// NVIDIADriverSpec defines the desired state of NVIDIADriver
+// NVIDIADriverSpec defines the desired state of NVIDIADriver.
+// The CEL validation allows non-default drivers to use nodeSelector, but requires
+// default drivers to leave nodeSelector unset or empty.
+// +kubebuilder:validation:XValidation:rule="has(self.default) && self.default ? !has(self.nodeSelector) || size(self.nodeSelector) == 0 : true",message="default NVIDIADriver must not use nodeSelector"
 type NVIDIADriverSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
+	// Default indicates that this NVIDIADriver acts as the fallback driver for GPU nodes
```
tariq1890:
```suggestion
	// Default indicates that this NVIDIADriver acts as the fallback driver daemon set manager for GPU nodes
```

controllers/nvidiadriver_controller.go
```diff
@@ -276,6 +298,65 @@ func (r *NVIDIADriverReconciler) enqueueAllNVIDIADrivers(ctx context.Context) []
...
+		if _, ok := seen[request.NamespacedName]; ok {
+			continue
+		}
+		seen[request.NamespacedName] = struct{}{}
+		deduped = append(deduped, request)
+	}
+	return deduped
+}
+
+func (r *NVIDIADriverReconciler) cleanupNVIDIADriverOwnerLabelsAndReturn(ctx context.Context) (reconcile.Result, error) {
```
tariq1890: Is this method needed? I don't see `cleanupNVIDIADriverOwnerLabels` being used outside of this, so we should just invoke this method directly. The `cleanupNVIDIADriverOwnerLabels` is already long and we should avoid creating longer method names IMO

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,143 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nvidiav1alpha1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1alpha1"
+	"github.com/NVIDIA/gpu-operator/internal/consts"
+)
+
+// ValidateNodeSelector rejects selectors that use operator-managed routing labels
+// or scope the default fallback driver.
+func ValidateNodeSelector(driver *nvidiav1alpha1.NVIDIADriver) error {
```
tariq1890: Can we make `driver *nvidiaalpha1.NVIDIADriver` the method receiver here?

internal/nvidiadriver/nvidiadriver.go
tariq1890: Anything we can add to the instrumentation so that we can closely monitor this? Or are the current metrics  we expose sufficient?

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,143 @@
...
+		return fmt.Errorf("default NVIDIADriver %q cannot use nodeSelector", driver.Name)
+	}
+	if _, ok := driver.Spec.NodeSelector[consts.NVIDIADriverOwnerLabel]; ok {
+		return fmt.Errorf("NVIDIADriver %q nodeSelector cannot use reserved label %q", driver.Name, consts.NVIDIADriverOwnerLabel)
+	}
+	return nil
+}
+
+// NodeMatchesSelector returns true when all selector labels are present on the node.
+func NodeMatchesSelector(node *corev1.Node, selector map[string]string) bool {
```
tariq1890: I don't think we need a public method for this. The only property we access in the `node.Labels` which is `map[string]string`. Is there an existing method we could use from the `maps` package instead?

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,143 @@
...
+func AssignOwners(ctx context.Context, c client.Client) (bool, error) {
+	drivers := &nvidiav1alpha1.NVIDIADriverList{}
+	if err := c.List(ctx, drivers); err != nil {
+		return false, fmt.Errorf("failed to list NVIDIADriver CRs: %w", err)
+	}
+
+	var defaultDriver *nvidiav1alpha1.NVIDIADriver
+	defaultDriverNames := []string{}
+	nonDefaultDrivers := make([]nvidiav1alpha1.NVIDIADriver, 0, len(drivers.Items))
+	for i := range drivers.Items {
```
tariq1890: It seems to me that the index variable is not really required? Can't we just use the element variable directly?

internal/validator/validator.go
```diff
@@ -45,25 +47,37 @@ func NewNodeSelectorValidator(c client.Client) Validator {
 // Check returns error when nodes matching with the selector labels of current instance of NVIDIADriver
 // are conflicting with other instances of NVIDIADriver
 func (nsv *nodeSelectorValidator) Validate(ctx context.Context, cr *nvidiav1alpha1.NVIDIADriver) error {
+	if err := nvidiadriverutil.ValidateNodeSelector(cr); err != nil {
```
tariq1890: This would read better IMO. This method can then call the respective `Validate` methods that are tried to the sub-fields of `NVIDIADriver`
```suggestion
	if err := cr.Validate(); err != nil {
```

internal/validator/validator.go
```diff
@@ -45,25 +47,37 @@ func NewNodeSelectorValidator(c client.Client) Validator {
...
 	drivers := &nvidiav1alpha1.NVIDIADriverList{}
 	err := nsv.client.List(ctx, drivers)
 	if err != nil {
 		return err
 	}
 
-	names := map[string]struct{}{}
+	selectedNodeOwners := map[string][]string{}
 	for di := range drivers.Items {
+		if err := nvidiadriverutil.ValidateNodeSelector(&drivers.Items[di]); err != nil {
```
tariq1890: Same comment as before; let's use the element variable instead of the index

internal/nvidiadriver/nvidiadriver.go
rajathagasthya: A user-defined NVIDIADriver CR that covers an entire cluster would conflict with all other NVIDIADriver CR with a non-empty NodeSelector. So I don't think we should allow non-default CRs with nil NodeSelector.

internal/validator/validator.go
rahulait: updated.

internal/validator/validator.go
rahulait: updated after refactoring. It now uses nvidiadriver's respective validation method.

internal/nvidiadriver/nvidiadriver.go
rahulait: Updated.

internal/nvidiadriver/nvidiadriver.go
rahulait: Updated

internal/nvidiadriver/nvidiadriver.go
rahulait: Yes, good point. Updated.

controllers/nvidiadriver_controller.go
rahulait: That was from some optimization but makes sense, updated code to have the two methods merged into one.

api/nvidia/v1alpha1/nvidiadriver_types.go
rahulait: Done.

internal/nvidiadriver/nvidiadriver.go
rahulait: I agree that if a user added any non-default nvidiadriver covering entire cluster, then subsequent user-defined nvidiadrivers will have conflicts. Restricting the user-defined CRs to always have a nodeselector will prevent some conflicts but not all. And we have it even today. So I am fine either way.

internal/nvidiadriver/nvidiadriver.go
rahulait: I think we should. controller-runtime metrics might tell us how much time a reconcile is taking, or if the queue is backing up, but it won't tell us how much time assignOwners is taking or any other call. We should at the very minimum atleast add logs which can be enabled to view time taken during each call. I'll add this requirement to the rate-limit request so that we do both at the same time. Let me know if you think we should add it to current PR and I'll add accordingly.

controllers/nvidiadriver_controller.go
```diff
@@ -276,6 +298,62 @@ func (r *NVIDIADriverReconciler) enqueueAllNVIDIADrivers(ctx context.Context) []
...
+
+	for i := range clusterPolicyList.Items {
+		if !clusterPolicyList.Items[i].Spec.Driver.IsNVIDIADriverCRDEnabled() {
+			continue
+		}
+		if _, err := nvidiadriverutil.AssignOwners(ctx, r.Client); err != nil {
+			log.FromContext(ctx).Error(err, "failed to cleanup NVIDIADriver owner labels")
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
```
tariq1890, line 344:
```suggestion
		break
```

controllers/nvidiadriver_controller.go
```diff
@@ -276,6 +298,62 @@ func (r *NVIDIADriverReconciler) enqueueAllNVIDIADrivers(ctx context.Context) []
...
+// NVIDIADriver matches a GPU node, including deletion of the last NVIDIADriver.
+func (r *NVIDIADriverReconciler) cleanupNVIDIADriverOwnerLabels(ctx context.Context) (reconcile.Result, error) {
+	clusterPolicyList := &gpuv1.ClusterPolicyList{}
+	if err := r.List(ctx, clusterPolicyList); err != nil {
+		wrappedErr := fmt.Errorf("error getting ClusterPolicy list: %w", err)
+		log.FromContext(ctx).Error(wrappedErr, "failed to cleanup NVIDIADriver owner labels")
+		return reconcile.Result{}, wrappedErr
+	}
+
+	for i := range clusterPolicyList.Items {
```
tariq1890: Let's not use the index variable. Also, do we need to range over this? Can't we just get the zeroth index?

internal/nvidiadriver/nvidiadriver.go
rahulait: I don't think there is a way to reliably always remove these labels. For now, I have made sure when nvidiadriver CRs are removed and controller is still running, the labels are removed from nodes. Once the controller is terminated, these labels can remain and can go stale. Once the controller comes back up and if it finds nodes with stale labels, it will clear them. I avoided adding any helm pre-uninstall hook as we have workflows where users can install gpu-operator without using helm (ex openshift) and those workarounds won't work for it. For now, the labels can remain but will be cleaned up or updated to correct value once controller again starts to manage those nodes.

controllers/nvidiadriver_controller.go
```diff
@@ -276,6 +298,62 @@ func (r *NVIDIADriverReconciler) enqueueAllNVIDIADrivers(ctx context.Context) []
 	return reconcileRequests
 }
 
+// enqueueNVIDIADriverReconcilers enqueues the NVIDIADriver that triggered the
+// event and all current NVIDIADriver instances. The triggering object is
+// included even for delete events so the NotFound reconcile path can clear
+// stale node owner labels.
+func (r *NVIDIADriverReconciler) enqueueNVIDIADriverReconcilers(ctx context.Context, driver *nvidiav1alpha1.NVIDIADriver) []reconcile.Request {
+	requests := r.enqueueAllNVIDIADrivers(ctx)
```
tariq1890, line 306: Instead of doing multiple list-alls of NVIDIADriver CR, how about creating a struct field (a map/set) in the `NVIDIADriverReconciler` struct which holds the unique set of all NVIDIADriver CRs? We'd add a new NVIDIADriver CR to the set if it's processed for the first time. We can remove the NVIDIADriver CR from this set when it is marked for deletion.

controllers/nvidiadriver_controller.go
rahulait: For range, one doesn't have to worry if the list is empty. Else, if we rely on index 0, then we have to first check and make sure there is one element in list and then access index 0. Both approaches are ok, range is cleaner and safer.

controllers/nvidiadriver_controller.go
rahulait: This will create a second source of truth beside the controller-runtime cache. We'll have to deal with startup correctness, delete correctness, concurrency, etc. The current r.List(ctx, list) is normally served from the informer cache, not a live API-server request, so it is much cheaper than it looks. It is also much safer because it reflects the cache’s current view after watch/list reconciliation. Are you seeing any performance issues with current approach? Just want to be clear on the issue you might be seeing with current approach.

controllers/nvidiadriver_controller.go
rahulait: Updated.

controllers/nvidiadriver_controller.go
rahulait: updated

controllers/nvidiadriver_controller.go
tariq1890: While it may not be an issue in clusters with a few NVIDIADrivers. This becomes a problem in large clusters where it's also possible to have several NVIDIADriver CRs to loop through. In these scenarios, we are bound to hit scaling problems.

controllers/nvidiadriver_migration.go
```diff
@@ -0,0 +1,110 @@
...
+		if !isDriverUpgradeRequestAllowed(upgradeState) {
+			continue
+		}
+		originalNode := node.DeepCopy()
+		if node.Labels == nil {
+			node.Labels = map[string]string{}
+		}
+		node.Labels[upgradeStateLabel] = upgrade.UpgradeStateUpgradeRequired
+		if err := r.Patch(ctx, node, client.MergeFrom(originalNode)); err != nil {
+			log.FromContext(ctx).Error(err, "failed to label node with orphaned NVIDIA driver pod", "pod", pod.Name, "node", node.Name)
```
tariq1890: Here we are swallowing the patch error. Is that intentional?

internal/validator/validator.go
```diff
@@ -45,25 +46,37 @@ func NewNodeSelectorValidator(c client.Client) Validator {
...
 
-	names := map[string]struct{}{}
-	for di := range drivers.Items {
-		nodeList, err := nsv.getNVIDIADriverSelectedNodes(ctx, &drivers.Items[di])
+	selectedNodeOwners := map[string][]string{}
+	for _, driver := range drivers.Items {
+		if err := driver.ValidateNodeSelector(); err != nil {
+			return err
+		}
+		if driver.IsDefault() {
```
tariq1890, line 64: Can we use the field selector in the List call to filter out the Default NVIDIADriver?

tariq1890: If yes, please apply this change in other places too (where applicable).

controllers/nvidiadriver_controller.go
```diff
@@ -96,6 +96,10 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, wrappedErr
 	}
+	if !instance.GetDeletionTimestamp().IsZero() {
```
tariq1890: Can we create a new method to extract this out? I see three uses of this.
```suggestion
	if instance.hasDeletionTimestamp() {
```

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,126 @@
...
+		matchingDrivers := []string{}
+		for _, driver := range nonDefaultDrivers {
+			if nodeMatchesSelector(node.Labels, driver.GetNodeSelector()) {
+				matchingDrivers = append(matchingDrivers, driver.Name)
+			}
+		}
+		if len(matchingDrivers) > 1 {
+			return false, fmt.Errorf("multiple NVIDIADrivers match the same node %s: %v", node.Name, matchingDrivers)
+		}
+	}
```
tariq1890, line 69: This should be moved to its own private method for better readability.

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,126 @@
...
+
+	changed := false
+	for i := range nodes.Items {
```
tariq1890: Let's not use the index variable here.

internal/nvidiadriver/nvidiadriver.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			delete(node.Labels, consts.NVIDIADriverOwnerLabel)
+			if err := c.Patch(ctx, node, client.MergeFrom(originalNode)); err != nil {
+				return false, fmt.Errorf("failed to remove NVIDIADriver owner label for node %s: %w", node.Name, err)
+			}
+			changed = true
+			continue
+		}
+		if node.Labels != nil && node.Labels[consts.NVIDIADriverOwnerLabel] == owner {
+			continue
```
tariq1890: The number of `continue` statements tell me that this for loop body can be restructured/rearranged.  This will make it a lot easier to grok IMO

controllers/nvidiadriver_migration.go
rahulait: Good catch, no thats not intentional. We should be returning error here. Let me fix it.

rahulait: updated.

controllers/nvidiadriver_controller.go
rahulait: updated.

internal/nvidiadriver/nvidiadriver.go
rahulait: Updated, code has been moved to its own private method.

internal/nvidiadriver/nvidiadriver.go
rahulait: updated.

internal/nvidiadriver/nvidiadriver.go
rahulait: updated.

internal/validator/validator.go
rahulait: Currently, `spec.default` is not registered as a selectable/indexed field. Even if we add that marker and make it selectable/indexed field, we still need to iterate over returned list to filter out CRs which are in currently getting deleted. For example, a user/workflow can add finalizers to their nvidiadriver CRs and that can keep deleting nvidiadriver CRs for some time. Skipping deleting nvidiadriver CRs avoids assigning or retaining ownership for something that is already being removed. Hence, I don't see much value in making this field selectable/indexed field. Current loop is simple enough to loop over all nvidiadrivers and classify them as default or not.

controllers/state_manager.go
```diff
@@ -1058,12 +1058,23 @@ func (n *ClusterPolicyController) step() (gpuv1.State, error) {
...
+			return gpuv1.NotReady, fmt.Errorf("failed to cleanup all NVIDIA driver daemonsets owned by ClusterPolicy: %w", err)
+		}
+		return gpuv1.Disabled, nil
+	}
+	if n.stateNames[n.idx] == "state-vgpu-manager" && n.singleton.Spec.Driver.IsNVIDIADriverCRDEnabled() {
+		n.logger.Info("NVIDIADriver CRD is enabled, cleaning up all NVIDIA driver daemonsets owned by ClusterPolicy")
+		n.idx++
+		// Cleanup all driver daemonsets owned by ClusterPolicy while keeping the
+		// running driver pods available until NVIDIADriver rolls replacements.
+		err := n.cleanupAllDriverDaemonSets(n.ctx, metav1.DeletePropagationOrphan)
```
tariq1890: Why are we introducing the duplication here?

tariq1890: Can't we go back to this?
```
if (n.stateNames[n.idx] == "state-driver" || n.stateNames[n.idx] == "state-vgpu-manager") &&
		n.singleton.Spec.Driver.IsNVIDIADriverCRDEnabled()
```

tariq1890 (approved):
Thank you @rahulait, excellent work on this! This is good to merge. Any other optimisations and improvements can be made in follow-up PRs

controllers/nvidiadriver_controller.go
rahulait: I think we should move the assignOwners() method to its own controller, that way it can reconcile on its own when needed and add ownership label. Then, we don't need to redo those calculations again and again on each reconcile in nvidiadriver CR and it can focus just on driver enablement. I'll add ownership-reconciler in a follow up PR which will improve things a lot in large clusters with lots of nvidiadrivers. I'll open follow up PRs for this.

rajathagasthya (approved):
LGTM!

controllers/nvidiadriver_controller.go
cdesiniotis: @rahulait in addition to moving `assignOwners()` to its own controller, what are your thoughts on moving the other label operations, like `labelNodesWithOrphanDriverPods()`, to this separate controller? Would that be possible / desirable? I am also thinking about whether it makes sense to move all of the node labeling that GPU Operator does (primarily in the clusterpolicy controller) to a dedicated controller.

PR #2524, opened 2026-06-08
add workflow for adding helm chart as oci artifact for commits

rahulait:
## Summary

Adds a reusable workflow to publish the GPU Operator Helm chart as a GHCR OCI artifact.

## Changes

- Adds `.github/workflows/publish-helm-oci-chart.yaml`.
- Defines the workflow as `workflow_call` so it can be triggered from `e2e-tests.yaml`.
- Packages `deployments/gpu-operator` as a Helm chart.
- Uses the operator image version as chart metadata:
  - chart version: `0.0.0-git-<commit_sha>`
  - app version: `<commit_sha>`
- Publishes the chart to `oci://ghcr.io/nvidia/gpu-operator/charts`.

## Notes

The workflow is intended to run after the matching GPU Operator image has been built and pushed, so the published chart points to an image tag that exists.

## Checklist

## Testing

Downloaded and installed the oci artifact built for this PR on a k8s cluster and everything comes up fine as expected.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -62,8 +62,18 @@ jobs:
       operator_image: ${{ inputs.operator_image }}
       operator_version: ${{ inputs.operator_version }}
 
-  e2e-tests-containerd:
+  publish-helm-oci-chart:
     needs: [variables]
+    permissions:
+      contents: read
+      packages: write
+    uses: ./.github/workflows/publish-helm-oci-chart.yaml
+    with:
+      operator_image: ${{ needs.variables.outputs.operator_image }}
+      operator_version: ${{ needs.variables.outputs.operator_version }}
+
+  e2e-tests-containerd:
+    needs: [variables, publish-helm-oci-chart]
     runs-on: linux-amd64-cpu4
     timeout-minutes: 90
     permissions:
@@ -129,7 +139,7 @@ jobs:
           retention-days: 15
 
   e2e-tests-nvidiadriver:
-    needs: [variables]
+    needs: [variables, publish-helm-oci-chart]
     runs-on: linux-amd64-cpu4
     timeout-minutes: 90
     permissions:
--- a/.github/workflows/publish-helm-oci-chart.yaml
+++ b/.github/workflows/publish-helm-oci-chart.yaml
@@ -1,0 +1,101 @@
+# Copyright NVIDIA CORPORATION
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Publish Helm OCI Chart
+
+on:
+  workflow_call:
+    inputs:
+      operator_image:
+        required: true
+        type: string
+      operator_version:
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  packages: write
+
+concurrency:
+  group: publish-helm-oci-chart-${{ inputs.operator_version }}
+  cancel-in-progress: false
+
+jobs:
+  publish-helm-chart:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+        name: Check out code
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v5.0.0
+
+      - name: Set chart image repository
+        env:
+          OPERATOR_IMAGE: ${{ inputs.operator_image }}
+        run: echo "OPERATOR_REPOSITORY=${OPERATOR_IMAGE%/*}" >> "${GITHUB_ENV}"
+
+      - name: Update chart image repository
+        uses: mikefarah/yq@v4
+        with:
+          cmd: yq -i '.operator.repository = env(OPERATOR_REPOSITORY) | .validator.repository = env(OPERATOR_REPOSITORY)' deployments/gpu-operator/values.yaml
+
+      - name: Login to GitHub Container Registry
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo "${GH_TOKEN}" | helm registry login ghcr.io -u "${GITHUB_ACTOR}" --password-stdin
+
+      - name: Package Helm chart
+        id: package
+        env:
+          OPERATOR_VERSION: ${{ inputs.operator_version }}
+        run: |
+          set -euo pipefail
+
+          # Helm chart versions must be SemVer. Prefix the image version so a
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-helm-oci-chart.yaml
```diff
@@ -0,0 +1,101 @@
...
+          OPERATOR_VERSION: ${{ inputs.operator_version }}
+        run: |
+          set -euo pipefail
+
+          # Helm chart versions must be SemVer, so encode the operator image
+          # version as a prerelease while keeping appVersion as the raw image tag.
+          CHART_VERSION="0.0.0-${OPERATOR_VERSION}"
+
+          mkdir -p dist
+          helm lint deployments/gpu-operator
```
tariq1890, line 73: Is this really needed?

.github/workflows/publish-helm-oci-chart.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Publish Helm OCI chart
+        env:
+          GH_REPOSITORY: ${{ github.repository }}
+          CHART_PACKAGE: ${{ steps.package.outputs.chart_package }}
+          CHART_VERSION: ${{ steps.package.outputs.chart_version }}
+        run: |
+          set -euo pipefail
+
+          REPOSITORY="$(echo "${GH_REPOSITORY}" | tr '[:upper:]' '[:lower:]')"
```
tariq1890: For readability, let's use the same approach and variable names when doing this as done in other yml files. See [here](https://github.com/NVIDIA/gpu-operator/blob/4cdd1b2694a013789012c6f89bf088a2483f4b51/.github/workflows/variables.yaml#L99) for an example

.github/workflows/publish-helm-oci-chart.yaml
rahulait: This publish workflow modifies deployments/gpu-operator/values.yaml right before packaging by setting the operator/validator repositories, and it can be invoked manually through e2e-tests.yaml inputs. The lint is cheap and acts as a final guard before publishing an immutable package to GHCR, so I would prefer to keep it.

.github/workflows/publish-helm-oci-chart.yaml
rahulait: Updated.

.github/workflows/publish-helm-oci-chart.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Update chart image repository
+        uses: mikefarah/yq@v4
+        with:
+          cmd: yq -i '.operator.repository = env(OPERATOR_REPOSITORY) | .validator.repository = env(OPERATOR_REPOSITORY)' deployments/gpu-operator/values.yaml
+
+      - name: Login to GitHub Container Registry
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo "${GH_TOKEN}" | helm registry login ghcr.io -u "${GITHUB_ACTOR}" --password-stdin
```
tariq1890, line 59:
```suggestion
        run: helm registry login ghcr.io -u "${GITHUB_ACTOR}" -p "${GH_TOKEN}"
```

.github/workflows/publish-helm-oci-chart.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Set chart image repository
+        env:
+          OPERATOR_IMAGE: ${{ inputs.operator_image }}
+        run: echo "OPERATOR_REPOSITORY=${OPERATOR_IMAGE%/*}" >> "${GITHUB_ENV}"
+
+      - name: Update chart image repository
+        uses: mikefarah/yq@v4
+        with:
+          cmd: yq -i '.operator.repository = env(OPERATOR_REPOSITORY) | .validator.repository = env(OPERATOR_REPOSITORY)' deployments/gpu-operator/values.yaml
```
tariq1890, line 54: Shouldn't we modify the version as well?

.github/workflows/publish-helm-oci-chart.yaml
rahulait: This can expose the token more easily through process arguments, shell tracing, or tool output. GitHub masks secrets in logs, but --password-stdin is still the better pattern and is what Docker/Helm registry login examples generally prefer.

https://docs.docker.com/reference/cli/docker/login/#password-stdin
https://helm.sh/docs/helm/helm_registry_login/

.github/workflows/publish-helm-oci-chart.yaml
rahulait: no, `--version` and `--app-version` are set in `helm package` command and we set the `--app-version` to commit-id. So, when the helm chart is rendered, the operator and validator version defaults to app version we had set for the chart when packaging it.

.github/workflows/publish-helm-oci-chart.yaml
tariq1890: Makes sense, thanks

.github/workflows/e2e-tests.yaml
```diff
@@ -62,8 +72,20 @@ jobs:
...
+      contents: read
+      packages: write
+    uses: ./.github/workflows/publish-helm-oci-chart.yaml
+    with:
+      operator_image: ${{ needs.variables.outputs.operator_image }}
+      operator_version: ${{ needs.variables.outputs.operator_version }}
+
+  e2e-tests-containerd:
+    needs: [variables, publish-helm-oci-chart]
+    if: ${{ always() && needs.variables.result == 'success' && (needs['publish-helm-oci-chart'].result == 'success' || needs['publish-helm-oci-chart'].result == 'skipped') }}
```
tariq1890: Why are we running this on `skipped`?

rahulait: There is an option to skip publishing helm oci chart when manually triggering e2e tests.
```
      publish_helm_oci:
        description: 'Publish the Helm chart as an OCI artifact'
        required: false
        type: boolean
        default: false
```
Maybe I can remove it and always have oci artifact published.

rahulait: Updated. Removed that ability now.

tariq1890 (approved):
Thanks @rahulait ! Let's get one more approve before merging

tariq1890:
As a follow-up, let's update the e2e test scripts to pull these helm charts instead of the project local one.

.github/workflows/publish-helm-oci-chart.yaml
```diff
@@ -0,0 +1,101 @@
...
+          set -euo pipefail
+
+          # Helm chart versions must be SemVer, so encode the operator image
+          # version as a prerelease while keeping appVersion as the raw image tag.
+          CHART_VERSION="0.0.0-${OPERATOR_VERSION}"
+
+          mkdir -p dist
+          helm lint deployments/gpu-operator
+          helm package \
+            --version "${CHART_VERSION}" \
```
karthikvetrivel, line 75: I found that the chart packaging will fail if the chart SHA begins with 0 (due to semver). Will that be problematic in this case?

rahulait: Yeah, thanks for catching that. Updated the versioning to be `0.0.0-git-<commit>` and in that case, it will always be a valid semver.

karthikvetrivel (approved):

PR #2525, opened 2026-06-08
[state-device-plugin] infer feature enablement envars from kernel modules

tariq1890:
### Problem

This PR is linked to https://github.com/NVIDIA/k8s-device-plugin/pull/1837. After PR https://github.com/NVIDIA/k8s-device-plugin/pull/1550 was merged to support `NVIDIADriver` use cases, we observed that default enablement of feature flags especially `MOFED_ENABLED` was leading to disruptive side effects. Users of GPU Operator v26.3.2 (which includes k8s-device-plugin v0.19.2) were observing that GPU workload containers were getting the entire list of `ibverbs` dev nodes injected into them.  This led to failures in RDMA/NCCL-based workloads where the workloads were referencing the wrong NICs as opposed to only picking the appropriate the NICs suited for that workload. See here for an example - https://github.com/NVIDIA/k8s-device-plugin/issues/1692.

### Solution

Instead of unconditionally enabling the feature flags, we enable them dynamically based on the kernel modules that are currently loaded on the node. This dynamic enablement of feature flags will ensure that `ClusterPolicy`, `NVIDIADriver`  and pre-installed driver use cases are supported without depending on a cluster-wide setting in the daemonset spec.

### DIFF
--- a/assets/state-device-plugin/0400_configmap.yaml
+++ b/assets/state-device-plugin/0400_configmap.yaml
@@ -18,6 +18,35 @@ data:
     set -o allexport
     cat /run/nvidia/validations/driver-ready
     . /run/nvidia/validations/driver-ready
+    
+    extra_modules="gdrdrv nvidia_fs"
+    > feature-flags.env
+    for module in ${extra_modules}; do
+        if lsmod | grep -q "^$module "; then
+            case "$module" in
+                "gdrdrv")
+                  if [ -z "$GDRCOPY_ENABLED" ]; then
+                      echo "GDRCOPY_ENABLED=true" >> feature-flags.env
+                  fi
+                ;;
+                "nvidia_fs")
+                  if [ -z "$GDS_ENABLED" ]; then
+                      echo "GDS_ENABLED=true" >> feature-flags.env
+                  fi
+                  if [ -z "$MOFED_ENABLED" ]; then
+                      echo "MOFED_ENABLED=true" >> feature-flags.env
+                  fi
+                ;;
+            esac
+        fi
+    done
+    
+    if [ -s "feature-flags.env" ]; then
+        echo "Enabling nvidia-device-plugin feature flags..."
+        cat feature-flags.env
+        . ./feature-flags.env
+    fi
 
     echo "Starting nvidia-device-plugin"
     exec nvidia-device-plugin
+    


assets/state-device-plugin/0400_configmap.yaml
```diff
@@ -18,6 +18,32 @@ data:
     set -o allexport
     cat /run/nvidia/validations/driver-ready
     . /run/nvidia/validations/driver-ready
+    
+    echo "Printing the status of GDS, MOFED & GDRCopy features..."
+    cat /run/nvidia/validations/.driver-ctr-ready
```
cdesiniotis: Question -- what happens if this file does not exist (e.g. when drivers are preinstalled)?

tariq1890: Good question. I need to think about this a bit more.

tariq1890: I've removed the dependency on the `/run/nvidia/driver/.driver-ctr-ready`. The implementation should now be agnostic to the type of driver installation

assets/state-device-plugin/0400_configmap.yaml
```diff
@@ -18,6 +18,28 @@ data:
...
+                      echo "MOFED_ENABLED=true"  >> feature-flags.env
+                  fi
+                ;;
+            esac
+        fi
+    done
 
     echo "Starting nvidia-device-plugin"
-    exec nvidia-device-plugin
+    exec env $(cat feature-flags.env | xargs) nvidia-device-plugin
```
cdesiniotis: Question -- is sourcing the `feature-flags.env` file not sufficient (like we do above for the `driver-ready` file)?

tariq1890: I went with this because I could not reliably verify if the `. /run/nvidia/validations/driver-ready` line actually led to env vars being passed into the `nvidia-device-plugin` binary

tariq1890: Let me make this change and test again

tariq1890: This works too, and looks cleaner. Thanks @cdesiniotis !

cdesiniotis: >I could not reliably verify if the . /run/nvidia/validations/driver-ready line actually led to env vars being passed into the nvidia-device-plugin binary

I am fairly confident that works since our ability to work with different driver roots depends on it.

cdesiniotis (approved):

rahulait (approved):
LGTM

assets/state-device-plugin/0400_configmap.yaml
```diff
@@ -18,6 +18,35 @@ data:
...
+                  if [ -z "$GDRCOPY_ENABLED" ]; then
+                      echo "GDRCOPY_ENABLED=true" >> feature-flags.env
+                  fi
+                ;;
+                "nvidia_fs")
+                  if [ -z "$GDS_ENABLED" ]; then
+                      echo "GDS_ENABLED=true" >> feature-flags.env
+                  fi
+                  if [ -z "$MOFED_ENABLED" ]; then
+                      echo "MOFED_ENABLED=true"  >> feature-flags.env
```
tariq1890: @rahulait @cdesiniotis  Note that I am pushing this change right now.
```suggestion
                      echo "MOFED_ENABLED=true" >> feature-flags.env
```

PR #2527, opened 2026-06-09
Restart driver pods in place when driver config is unchanged

rajathagasthya:
## Description

A patch chart upgrade can change only cosmetic pod-template metadata (e.g. the `helm.sh/chart` label) without changing the driver itself. The upgrade controller keys on the DaemonSet's controller revision hash, so such a change still evicts running GPU workloads and drains the node, causing disruption for running workloads.

Register a `RestartOnlyPredicate` on the upgrade state manager (from the `UpgradeReconciler`) that compares `DRIVER_CONFIG_DIGEST` — a hash of the install-relevant driver config, already set on the driver pod template — between the running pod and the desired DaemonSet. When the digests match, the node is cordoned and the driver pod restarted in place, with no workload eviction or drain; the driver fast-path keeps the kernel modules loaded across the restart, so running GPU workloads are not disrupted. Cordoning keeps the node unschedulable if the restart fails (same as the full flow), and the node is uncordoned on success. A missing or differing digest falls back to the full upgrade flow.

If the predicate returns an error or the cordon fails, the node stays in `upgrade-required` and is retried on a later reconcile (with a Warning event), rather than being routed to the disruptive flow on an unknown answer.

Known limitation: the first upgrade from a release without restart-only is still disruptive, because the old operator holds the leader-election lease and routes the upgrade before the new operator becomes leader. Steady-state (both sides have the code) is non-disruptive.

Related to https://github.com/NVIDIA/k8s-operator-libs/pull/145

## Checklist

## Testing

Unit tests:
- `internal/config`: `TestDriverConfigDigestFromPodSpec` — digest reader, incl. nil/empty/container-precedence cases.
- `controllers`: `TestDriverPodRestartOnly` — predicate routing, incl. nil pod/DS and missing/equal/differing digests.
- NVIDIA/k8s-operator-libs#145: Ginkgo specs for restart-only routing (cordon + pod-restart), retry-on-error for predicate and cordon failures, orphaned/upgrade-requested/safe-driver-load skips, and the maxParallelUpgrades throttle.

Manual testing (single-node cluster, GPU workload running throughout):

1. **Without helm:** deploy an operator image with this change, then patch the driver DS pod template with a label-only change. Node goes `upgrade-required → pod-restart-required` (cordoned, never `cordon-required`), the driver pod restarts via the fast path, and the GPU workload is not evicted.
2. **Helm, first adoption:** install v26.3.2, then upgrade to a chart built with this change. Full upgrade flow, as expected: the driver version also changed, and the first upgrade is routed by the old operator (see known limitation above).
3. **Helm, patch upgrade:** install a chart built with this change, then upgrade to another chart also built with this change, same driver version. Restart-only flow — the GPU workload is not evicted.
4. **Helm, real driver change:** install a chart built with this change, then upgrade to another chart also built with this change but a different driver version. Full upgrade flow (digest differs).

### DIFF
--- a/cmd/gpu-operator/main.go
+++ b/cmd/gpu-operator/main.go
@@ -51,6 +51,7 @@ import (
 	"github.com/NVIDIA/gpu-operator/controllers/clusterinfo"
 	"github.com/NVIDIA/gpu-operator/internal/consts"
 	"github.com/NVIDIA/gpu-operator/internal/info"
+	"github.com/NVIDIA/gpu-operator/internal/predicates"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -184,7 +185,10 @@ func main() {
 		setupLog.Error(err, "unable to create new ClusterUpdateStateManager", "controller", "Upgrade")
 		os.Exit(1)
 	}
-	clusterUpgradeStateManager = clusterUpgradeStateManager.WithPodDeletionEnabled(gpuPodSpecFilter).WithValidationEnabled("app=nvidia-operator-validator")
+	clusterUpgradeStateManager = clusterUpgradeStateManager.
+		WithPodDeletionEnabled(gpuPodSpecFilter).
+		WithValidationEnabled("app=nvidia-operator-validator").
+		WithRestartOnlyPredicate(predicates.DriverPodRestartOnly(upgradeLogger))
 
 	if err = (&controllers.UpgradeReconciler{
 		Client:          mgr.GetClient(),
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -1064,19 +1064,19 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 	// Set the computed digest in driver-manager initContainer
 	driverManagerContainer := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
 	if driverManagerContainer != nil {
-		setContainerEnv(driverManagerContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+		setContainerEnv(driverManagerContainer, driverconfig.DriverConfigDigestEnvName, configDigest)
 	}
 
 	// Set the computed digest in nvidia-driver container
 	driverContainer := findContainerByName(obj.Spec.Template.Spec.Containers, "nvidia-driver-ctr")
 	if driverContainer != nil {
-		setContainerEnv(driverContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+		setContainerEnv(driverContainer, driverconfig.DriverConfigDigestEnvName, configDigest)
 	}
 
 	// Used by dtk-build-driver to determine if fast path should be used (skip rebuild)
 	driverToolkitContainer := findContainerByName(obj.Spec.Template.Spec.Containers, "openshift-driver-toolkit-ctr")
 	if driverToolkitContainer != nil {
-		setContainerEnv(driverToolkitContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+		setContainerEnv(driverToolkitContainer, driverconfig.DriverConfigDigestEnvName, configDigest)
 	}
 
 	// set hostNetwork for driver if specified
--- a/internal/config/driver_config_digest.go
+++ b/internal/config/driver_config_digest.go
@@ -22,6 +22,38 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+// DriverConfigDigestEnvName is the env var the operator sets on the driver pod
+// template, carrying a hash of the install-relevant driver config (DriverInstallState).
+const DriverConfigDigestEnvName = "DRIVER_CONFIG_DIGEST"
+
+// DriverConfigDigestFromPodSpec returns the DRIVER_CONFIG_DIGEST value from a driver
+// pod spec, or "" if absent. The env is set identically on every driver container, so
+// the first non-empty value (init containers first) is returned.
+func DriverConfigDigestFromPodSpec(spec *corev1.PodSpec) string {
+	if spec == nil {
+		return ""
+	}
+	digestFromEnv := func(env []corev1.EnvVar) string {
+		for _, e := range env {
+			if e.Name == DriverConfigDigestEnvName {
+				return e.Value
+			}
+		}
+		return ""
+	}
+	for _, initCtr := range spec.InitContainers {
+		if v := digestFromEnv(initCtr.Env); v != "" {
+			return v
+		}
+	}
+	for _, ctr := range spec.Containers {
+		if v := digestFromEnv(ctr.Env); v != "" {
+			return v
+		}
+	}
+	return ""
+}
+
 // DriverInstallState lists all fields that affect driver installation.
 // Changes to these fields trigger a driver reinstall.
 //
--- a/vendor/github.com/NVIDIA/k8s-operator-libs/pkg/upgrade/common_manager.go
+++ b/vendor/github.com/NVIDIA/k8s-operator-libs/pkg/upgrade/common_manager.go
@@ -79,6 +79,17 @@ func NewClusterUpgradeState() ClusterUpgradeState {
 	return ClusterUpgradeState{NodeStates: make(map[string][]*NodeUpgradeState)}
 }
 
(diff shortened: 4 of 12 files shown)


vendor/github.com/NVIDIA/k8s-operator-libs/pkg/upgrade/common_manager.go
```diff
@@ -79,6 +79,14 @@ func NewClusterUpgradeState() ClusterUpgradeState {
 	return ClusterUpgradeState{NodeStates: make(map[string][]*NodeUpgradeState)}
 }
 
+// RestartOnlyPredicate is consulted for a node whose driver pod is out-of-sync with
+// its DaemonSet. Returning true means the difference is immaterial to the installed
+// driver, so the pod may be restarted in place without cordon, pod-deletion, or drain;
```
cdesiniotis: Question -- should we skip the cordon in this case? What happens if the `pod-restart` fails for some reason, would it be better to keep the node cordoned until the failures are resolved?

rajathagasthya: Good point. Keeping the node cordoned during failed driver pod restarts is also the current behavior, so I think that's very reasonable.

rajathagasthya: Kept the cordon in case of pod restart failures.

controllers/upgrade_controller.go
```diff
@@ -231,10 +232,38 @@ func (r *UpgradeReconciler) removeNodeUpgradeStateLabels(ctx context.Context) er
 	return nil
 }
 
+// driverPodRestartOnly is the upgrade controller's RestartOnlyPredicate: it allows an
+// out-of-sync driver pod to be restarted in place when the running pod and the desired
+// DaemonSet have the same DRIVER_CONFIG_DIGEST, i.e. the install-relevant config is
+// unchanged (e.g. only a helm.sh/chart label changed). If either digest is missing, it
+// returns false and the node takes the full upgrade flow.
+func (r *UpgradeReconciler) driverPodRestartOnly(_ context.Context, pod *corev1.Pod, ds *appsv1.DaemonSet) (bool, error) {
```
tariq1890: If we are not going to use `ctx` why add it to the function signature of the predicate type?

tariq1890: Instead of `Pod` and `Daemonset` as the arguments, can we have both the arguments of identical types instead? Having `corev1.Pod` on the LHS and the `appsv1.DaemonSet` on the RHS is not too intuitive IMO

controllers/upgrade_controller.go
```diff
@@ -231,10 +232,38 @@ func (r *UpgradeReconciler) removeNodeUpgradeStateLabels(ctx context.Context) er
...
+	if pod == nil || ds == nil {
+		return false, nil
+	}
+	desired := driverconfig.DriverConfigDigestFromPodSpec(&ds.Spec.Template.Spec)
+	running := driverconfig.DriverConfigDigestFromPodSpec(&pod.Spec)
+	if desired == "" || running == "" {
```
tariq1890: Let's do a `len() ==  0` check instead

controllers/upgrade_controller.go
rajathagasthya: I kept the predicate signature for flexibility in case we change how we compare pod and daemonset in the future. But I can convert it to:

```go
type RestartOnlyPredicate func(running, desired *corev1.PodSpec) (bool, error)
```
I don't particularly have a strong opinion.

controllers/upgrade_controller.go
rajathagasthya: Why? Afaik, `if str == ""` is the most common and idiomatic way of checking for an empty string (I see this everywhere in this repo), while `len() == 0` is idiomatic for slices/maps.

controllers/upgrade_controller.go
rajathagasthya: Changed it to the above. Let me know if it looks good.

controllers/upgrade_controller.go
tariq1890: The `len` check in go covers empty and nil cases too; both lead to `0` being the returned value. Not a strong opinion

rajathagasthya: I'll keep it as is since `desired` and `running` are strings, so they cannot be `nil` unlike slices/maps.

cmd/gpu-operator/main.go
```diff
@@ -184,7 +184,10 @@ func main() {
 		setupLog.Error(err, "unable to create new ClusterUpdateStateManager", "controller", "Upgrade")
 		os.Exit(1)
 	}
-	clusterUpgradeStateManager = clusterUpgradeStateManager.WithPodDeletionEnabled(gpuPodSpecFilter).WithValidationEnabled("app=nvidia-operator-validator")
+	clusterUpgradeStateManager = clusterUpgradeStateManager.
+		WithPodDeletionEnabled(gpuPodSpecFilter).
+		WithValidationEnabled("app=nvidia-operator-validator").
+		WithRestartOnlyPredicate(controllers.DriverPodRestartOnly(upgradeLogger))
```
tariq1890: What if we moved the method into a new package called `controllers/predicates`?
```suggestion
		WithRestartOnlyPredicate(predicates.DriverPodRestartOnly(upgradeLogger))
```

controllers/object_controls.go
```diff
@@ -1064,19 +1064,19 @@ func TransformDriver(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpec, n C
 	// Set the computed digest in driver-manager initContainer
 	driverManagerContainer := findContainerByName(obj.Spec.Template.Spec.InitContainers, "k8s-driver-manager")
 	if driverManagerContainer != nil {
-		setContainerEnv(driverManagerContainer, "DRIVER_CONFIG_DIGEST", configDigest)
+		setContainerEnv(driverManagerContainer, driverconfig.DriverConfigDigestEnvName, configDigest)
```
tariq1890, line 1067: Why not move this to a common `const` package that both the current and the `driverconfig` packages can import from? Don't we already have a `const` package like that?

tariq1890: I see this package - `"github.com/NVIDIA/gpu-operator/internal/consts"`

cmd/gpu-operator/main.go
rajathagasthya: Good suggestion! I didn't think `controllers` was a good destination for this predicate either.

rajathagasthya: What about `internal/predicates` instead? This predicate is not used in `controllers/` directly. It just needs to be registered with the upgrade controller in `main.go`.

controllers/object_controls.go
rajathagasthya: My rationale is that this env is specific to the digest feature, which is self contained in that package, and that package is already imported in this file. So I lean towards keeping this in `driverconfig`, but I don't feel strongly about this.

cmd/gpu-operator/main.go
tariq1890: `internal/predicates` is fine too

cdesiniotis (approved):

cdesiniotis:
LGTM, thanks @rajathagasthya. Let's get another approval on this.

internal/config/driver_config_digest.go
```diff
@@ -22,6 +22,38 @@ import (
...
+	}
+	digestFromEnv := func(env []corev1.EnvVar) string {
+		for _, e := range env {
+			if e.Name == DriverConfigDigestEnvName {
+				return e.Value
+			}
+		}
+		return ""
+	}
+	for i := range spec.InitContainers {
```
tariq1890:
```suggestion
	for _, initCtr := range spec.InitContainers {
```

internal/config/driver_config_digest.go
```diff
@@ -22,6 +22,38 @@ import (
...
+		if v := digestFromEnv(spec.InitContainers[i].Env); v != "" {
+			return v
+		}
+	}
+	for i := range spec.Containers {
```
tariq1890:
```suggestion
	for _, ctr := range spec.Containers {
```

controllers/object_controls.go
tariq1890: I'll resolve this comment.

tariq1890 (approved):

PR #2540, opened 2026-06-12
feat(helm): add configurable network policies for gpu-operator components

apsega:
## Summary
- Add global `networkPolicy` configuration in `deployments/gpu-operator/values.yaml` (`enabled`, labels/annotations, global ingress `from` and `ports`).
- Add per-component `networkPolicy` settings under each component key (operator, dcgmExporter, nodeStatusExporter, and NFD master/worker/gc/topologyUpdater) with default metrics port whitelists.
- Add `deployments/gpu-operator/templates/networkpolicy.yaml` to render:
  - component-specific ingress whitelist policies
  - one cluster-wide deny-all-ingress policy (`podSelector: {}`) whenever any component policy is enabled

## Why
This adds a least-privilege network baseline for GPU Operator workloads while preserving flexibility to enable policy gradually and tune per-component ingress without changing templates.

## Test Plan
- [x] `helm lint deployments/gpu-operator`
- [x] `helm template gpu-operator deployments/gpu-operator --set networkPolicy.enabled=true`
- [x] `helm template gpu-operator deployments/gpu-operator --set networkPolicy.enabled=false --set operator.networkPolicy.enabled=true`
- [x] Confirm rendered output includes component whitelist policies and a single `gpu-operator-deny-all-ingress` policy when policy is enabled

## Summary by CodeRabbit

- **New Features**
  - Added optional Kubernetes NetworkPolicy support for GPU Operator components and related exporters.
  - Added configurable ingress sources and allowed ports for each supported component.
  - Added a namespace-wide deny-all-ingress policy when component network policies are enabled.
  - Network policies are disabled by default and created only when relevant ports are configured.

---

<a href="https://nvidia-cloud.devinenterprise.com/review/nvidia/gpu-operator/pull/2540" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>


dmeetreey (approved):

tariq1890:
Thank you for your contribution @apsega! For changes to the dcgm-exporter manifests, they would first need to be made in the [dcgm-exporter standalone helm chart](https://github.com/NVIDIA/dcgm-exporter/tree/main/deployment). The dcgm-exporter manifests that we maintain in gpu-operator need to align with the standalone dcgm-exporter chart.

apsega:
added this to https://github.com/NVIDIA/dcgm-exporter/pull/677

rahulait:
@coderabbitai review

deployments/gpu-operator/values.yaml
```diff
@@ -660,3 +711,11 @@ node-feature-discovery:
       # resourceLabels: ["nvidia.com/feature-1","nvidia.com/feature-2"]
       # enableTaints: false
       # labelWhiteList: "nvidia.com/gpu"
+  topologyUpdater:
+    networkPolicy:
+      enabled: false
+      ingress:
+        from: []
+        ports:
+          - port: 8080
```
rahulait: It looks like each allowlist hard-codes port 8080 instead of deriving the component's existing port value.

deployments/gpu-operator/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,315 @@
...
+{{- $nfdTopologyIngress := $nfdTopologyNP.ingress | default dict -}}
+{{- $nfdTopologyFrom := concat $globalFrom ($nfdTopologyIngress.from | default (list)) -}}
+{{- $nfdTopologyPorts := concat ($nfdTopologyIngress.ports | default (list)) $globalPorts -}}
+{{- if gt (len $nfdTopologyPorts) 0 }}
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: gpu-operator-nfd-topology-updater
+  namespace: {{ .Release.Namespace }}
```
rahulait: The NFD subchart honors node-feature-discovery.namespaceOverride, but these policies always use .Release.Namespace. With an override, the policies select no NFD pods and global NetworkPolicy mode leaves NFD unrestricted.

deployments/gpu-operator/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,315 @@
...
+  {{- end }}
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: {{ $nfdName | quote }}
+      app.kubernetes.io/instance: {{ .Release.Name | quote }}
+      role: "master"
+  policyTypes:
+    - Ingress
+  ingress:
```
kvalliyurnatt: I am seeing an error in helm lint
```
[ERROR] templates/: values don't meet the specifications of the schema(s):
node-feature-discovery:
- at '/worker/networkPolicy/ingress': got object, want array
- at '/gc/networkPolicy/ingress': got object, want array
- at '/topologyUpdater/networkPolicy/ingress': got object, want array
- at '/master/networkPolicy/ingress': got object, want array
Error: 1 chart(s) linted, 1 chart(s) failed
```

The NFD subchart declares ingress as an array, so declaring it as an object here is conflicting, I think this needs to be changed for all NFD related network policy

kvalliyurnatt (changes requested):

deployments/gpu-operator/values.yaml
apsega: The other component ports (operator 8080, dcgm-exporter 9400, node-status-exporter 8000) are
fixed in the operator template

deployments/gpu-operator/templates/networkpolicy.yaml
apsega: Good point, I've dropped the NFD policies from this chart entirely and left them to the subchart.

rahulait:
There are still a couple of blocking issues:
- Preserve helm upgrade --reuse-values compatibility. [networkpolicy.yaml:1](https://github.com/NVIDIA/gpu-operator/pull/2540/files) dereferences .Values.networkPolicy.enabled (and nested ingress fields) before defaulting the parent map. Existing releases upgraded with --reuse-values do not receive newly introduced chart defaults, so they have no networkPolicy map and rendering fails with a nil-pointer error. Default the parent and ingress maps first.
- A component-level toggle must not install a namespace-wide deny policy. [networkpolicy.yaml:124](https://github.com/NVIDIA/gpu-operator/pull/2540/files) emits podSelector: {} whenever any component policy is enabled. Thus operator.networkPolicy.enabled=true also denies ingress to DCGM exporter, node-status exporter, NFD, and unrelated workloads in the release namespace—without emitting their corresponding allow policies. This contradicts the advertised gradual per-component enablement and can break metrics scraping. Emit the namespace-wide deny policy only for global mode, or scope the deny policies to the components whose policies are enabled.

rahulait:
Hi @apsega thank you for the contribution and your patience with this PR.

After internal discussions, it was decided not to add built-in NetworkPolicy management controlled by chart flags at this time.

Network policies are highly environment-specific: permitted traffic depends on the CNI implementation, monitoring setup, namespace layout, enabled operands, custom labels, host networking, and any additional workloads installed alongside GPU Operator. Once we expose and document component-level policy flags, we also take on the ongoing compatibility burden of maintaining those policies for every supported deployment configuration and for future components.

The chart already provides [extraObjects](https://github.com/NVIDIA/gpu-operator/blob/93c0ff560ca71e761d6c1782d904765bb88de3db/deployments/gpu-operator/values.yaml#L640), which is a better fit for this use case. It lets users deploy NetworkPolicy resources alongside the chart while retaining full control over selectors, ports, and allowed sources for their own environment.
For example:
```
extraObjects:
  - |
    apiVersion: networking.k8s.io/v1
    kind: NetworkPolicy
    metadata:
      name: allow-dcgm-exporter-metrics
      namespace: {{ .Release.Namespace }}
    spec:
      podSelector:
        matchLabels:
          app: nvidia-dcgm-exporter
      policyTypes:
        - Ingress
      ingress:
        - from:
            - namespaceSelector:
                matchLabels:
                  kubernetes.io/metadata.name: monitoring
          ports:
            - protocol: TCP
              port: 9400
```
Users can add similar policies for the components they run and tailor the source selectors and ports to their cluster’s security model.

rahulait:
Closing this PR for now as this is not something we want to add by default in gpu-operator. Configurations like this brings additional compatibility burden of maintaining those policies for every supported deployment configuration and for future components. Suggested extraObjects should be used by users who want to bring in their own custom configuration by default.

PR #2569, opened 2026-06-23
Create new controller for all node labeling operations

cdesiniotis:
This PR adds a new controller, named `node-labeling-controller`, which is responsible for labeling k8s nodes with GPU Operator-related state. Currently, both the clusterpolicy and nvidiadriver controller's label nodes. When DRA is integrated, we will have yet another controller that will need to label nodes as well. The intent is for the `node-labeling-controller` to reconcile any node labels required by any of these controllers. The only exception here is the `nvidia.com/gpu-driver-upgrade-state` label which will still be managed by the `driver-upgrade` controller for facilitating driver daemonset upgrades.

### Testing

I tested the following scenarios manually:

1. Install GPU Operator with default values. Verify relevant `nvidia.com/*` node labels are added and all GPU Operator pods come up successfully.
2. Remove `nvidia.com/gpu.present` label and verify it gets re-added to the node without disruption to operands.
3. Disable a single operand via node label. I labeled my GPU node with `nvidia.com/gpu.deploy.mig-manager=false` and verified only the mig-manager pod got removed. I re-set this label to true and verified the mig-manager pod got rescheduled.
4. Migrate from ClusterPolicy-owned driver daemonset to NVIDIADriver-owned driver daemonset. I set `driver.useNvidiaDriverCRD=true` in ClusterPolicy and created a default NVIDIADriver CR. Verified node got labeled with `nvidia.com/gpu-operator.driver.owner=default`. Verified old driver pod get replaced with new one.
5. Deploy a non-default NVIDIADriver CR that targets my GPU node. Once the CR was created, I verified that the `nvidia.com/gpu-operator.driver.owner` was updated from `default` to the name of the CR I just created. Verified that the new driver pod comes up successfully.
6. Disable all operands by setting `nvidia.com/gpu.deploy.operands=false` label.  
7. Re-enable all operands by setting `nvidia.com/gpu.deploy.operands=true` label.
8. Add new GPU worker node to cluster. Verify appropriate `nvidia.com/*` node labels are added. Verify default NVIDIADriver gets deployed (`nvidia.com/gpu-operator.driver.owner` gets set to `default` on the new node) and all operands come up successfully.


controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		return reconcile.Result{}, fmt.Errorf("failed to list ClusterPolicy: %w", err)
+	}
+
+	// (cdesiniotis) Return early if a ClusterPolicy CR does not exist.
+	// This means that nodes will not get labeled unless a ClusterPolicy
+	// CR has been created. This may be relaxed in the future when the
+	// NVIDIA DRA Driver for GPUs is integrated with the GPU Operator
+	// and new CRDs are introduced.
+	if len(clusterPolicyList.Items) == 0 {
+		return reconcile.Result{}, fmt.Errorf("no ClusterPolicy object exists")
```
rahulait: Instead of returning error here, maybe return nil so that we don't keep retrying.

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,426 @@
...
+
+		if node.Annotations == nil {
+			node.Annotations = map[string]string{}
+		}
+		if autoUpgradeEnabled {
+			node.Annotations[driverAutoUpgradeAnnotationKey] = "true"
+		} else {
+			delete(node.Annotations, driverAutoUpgradeAnnotationKey)
+		}
+		if err := nlc.client.Update(ctx, &node); err != nil {
```
rahulait: Lets use patch instead of update.

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		CreateFunc: func(e event.TypedCreateEvent[*corev1.Node]) bool {
+			labels := e.Object.GetLabels()
+			return hasGPULabels(labels) || hasCommonGPULabel(labels)
+		},
+		UpdateFunc: func(e event.TypedUpdateEvent[*corev1.Node]) bool {
+			oldL, newL := e.ObjectOld.GetLabels(), e.ObjectNew.GetLabels()
+			relevant := hasGPULabels(oldL) || hasGPULabels(newL) ||
+				hasCommonGPULabel(oldL) || hasCommonGPULabel(newL)
+			return relevant && !maps.Equal(oldL, newL)
+		},
```
rahulait, line 463: This will still trigger reconcile on every node label update. To avoid always calculating new owners on every node label update, I was thinking we should reconcile only for selected set of labels:
1. GPU labels and GPU present label
2. Labels in nodeSelectors of all nvidiadrivers
3. Ownership label itself

So like:
```
func (r *NVIDIADriverOwnershipReconciler) nodeOwnershipLabelsChanged(ctx context.Context, oldLabels, newLabels map[string]string) bool {
	if oldLabels[consts.GPUPresentLabel] != newLabels[consts.GPUPresentLabel] {
		return true
	}
	if oldLabels[consts.NVIDIADriverOwnerLabel] != newLabels[consts.NVIDIADriverOwnerLabel] {
		return true
	}
	if !hasGPULabels(oldLabels) && !hasGPULabels(newLabels) {
		return false
	}

	nvidiaDrivers := &nvidiav1alpha1.NVIDIADriverList{}
	if err := r.List(ctx, nvidiaDrivers); err != nil {
		log.FromContext(ctx).Error(err, "failed to list NVIDIADrivers while checking node label update")
		return true
	}

	for _, nvidiaDriver := range nvidiaDrivers.Items {
		for key := range nvidiaDriver.Spec.NodeSelector {
			if oldLabels[key] != newLabels[key] {
				return true
			}
		}
	}
	return false
}
```

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,426 @@
...
+			nlc.logger.Error(err, "failed to get node for orphaned driver pod",
+				"pod", pod.Name, "node", pod.Spec.NodeName)
+			continue
+		}
+		if !nodeOwnedByNVIDIADriver(node, nvidiaDrivers.Items) {
+			continue
+		}
+
+		upgradeStateLabel := upgrade.GetUpgradeStateLabelKey()
+		upgradeState := ""
```
rahulait: We should set this to:
```
upgradeState := upgrade.UpgradeStateUnknown
```

This is because isDriverUpgradeRequestAllowed() relies on this state.

controllers/nodelabeling_controller.go
karthikvetrivel: Question: should this be `Update` or `Patch`? 

Other node writers in this file use `Patch` so just curious here.

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,426 @@
...
+		cp.Spec.Driver.UpgradePolicy.AutoUpgrade &&
+		!cp.Spec.SandboxWorkloads.IsEnabled()
+
+	list := &corev1.NodeList{}
+	if err := nlc.client.List(ctx, list); err != nil {
+		return fmt.Errorf("unable to list nodes: %w", err)
+	}
+
+	for _, node := range list.Items {
+		node := node
```
karthikvetrivel: nit: why do we need this copy? Correct me if I'm wrong, but as of Go 12.2, each iteration of a `for` loop gets it own copy of the iterating variable.

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+func (nlc *nodeLabelingController) labelGPUNodes(ctx context.Context) error {
+	list := &corev1.NodeList{}
+	if err := nlc.client.List(ctx, list); err != nil {
+		return fmt.Errorf("unable to list nodes: %w", err)
+	}
+
+	for _, node := range list.Items {
+		node := node
```
karthikvetrivel: Same question here as above.

controllers/nodelabeling_controller.go
cdesiniotis: Makes sense. Updated to not return an error.

controllers/nodelabeling_controller.go
cdesiniotis: Updated!

controllers/nodelabeling_controller.go
cdesiniotis: Good catch. Updated!

controllers/nodelabeling_controller.go
cdesiniotis: Updated to `Patch`.

cdesiniotis: Updated to `Patch`.

controllers/nodelabeling_controller.go
cdesiniotis: Thanks for catching this! Updated as suggested. Also reviewed this method again and verified there are no other discrepancies after moving it here from `nvidiadriver_migration.go`.

controllers/nodelabeling_controller.go
cdesiniotis: I have updated this significantly now to align with the predicates used in ClusterPolicy previously. This should cover the three events you just listed, PTAL.

cdesiniotis: Question -- I notice the NVIDIADriver controller reconciled on every node label update. Do we want to update this in a follow-up PR? https://github.com/NVIDIA/gpu-operator/blob/9e35d5d4d2b30ca123aae53176ad9b8dfa6342f7/controllers/nvidiadriver_controller.go#L440

karthikvetrivel (approved):

a-mccarthy:
@cdesiniotis, looking through the PR and description, it seems like this will be a behind the scenes change for users. Is there anything that we should plan to document here? Or just a release note? 

Is this planned to be included in the 26.7.0 release? its not currently added to a milestone.

cdesiniotis:
> @cdesiniotis, looking through the PR and description, it seems like this will be a behind the scenes change for users. Is there anything that we should plan to document here? Or just a release note?
> 
> Is this planned to be included in the 26.7.0 release? its not currently added to a milestone.

@a-mccarthy this is planned for 26.7.0, just added it to the milestone. I would say this is an implementation detail (with regards to how we label nodes in gpu-operator) that does not require documentation. We could consider mentioning this as an "improvement" in our release notes but I haven't put much thought into it.

config/rbac/role.yaml
```diff
@@ -27,6 +26,16 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
tariq1890: This reduces the available permissions of the role. While I am okay with this, I think it should be done in a separate PR.

controllers/clusterpolicy_controller.go
```diff
@@ -294,19 +295,17 @@ func addWatchNewGPUNode(r *ClusterPolicyReconciler, c controller.Controller, mgr
...
 				gpuWorkloadConfigLabelChanged ||
 				osTreeLabelChanged
 
 			if needsUpdate {
 				r.Log.Info("Node needs an update",
 					"name", nodeName,
-					"gpuCommonLabelMissing", gpuCommonLabelMissing,
+					"gpuCommonLabelAdded", gpuCommonLabelAdded,
 					"gpuCommonLabelOutdated", gpuCommonLabelOutdated,
-					"migManagerLabelMissing", migManagerLabelMissing,
```
tariq1890, line 308: why remove this?

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,475 @@
...
+		return reconcile.Result{}, fmt.Errorf("failed to list ClusterPolicy: %w", err)
+	}
+
+	// (cdesiniotis) Return early if a ClusterPolicy CR does not exist.
+	// This means that nodes will not get labeled unless a ClusterPolicy
+	// CR has been created. This may be relaxed in the future when the
+	// NVIDIA DRA Driver for GPUs is integrated with the GPU Operator
+	// and new CRDs are introduced.
+	if len(clusterPolicyList.Items) == 0 {
+		return reconcile.Result{}, nil
```
tariq1890, line 82: Should we add a logging statement here?

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,475 @@
...
+		if err := nlc.labelNodesWithOrphanedDriverPods(ctx); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to label nodes with orphaned NVIDIA driver pods: %w", err)
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (nlc *nodeLabelingController) labelGPUNodes(ctx context.Context) error {
+	list := &corev1.NodeList{}
```
tariq1890:
```suggestion
	nodeList := &corev1.NodeList{}
```

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,475 @@
...
+
+// applyDriverAutoUpgradeAnnotation sets or clears the driver auto-upgrade annotation on GPU nodes.
+func (nlc *nodeLabelingController) applyDriverAutoUpgradeAnnotation(ctx context.Context) error {
+	cp := nlc.clusterPolicy
+	autoUpgradeEnabled := (cp.Spec.Driver.IsEnabled() || cp.Spec.Driver.UseNvidiaDriverCRDType()) &&
+		cp.Spec.Driver.UpgradePolicy != nil &&
+		cp.Spec.Driver.UpgradePolicy.AutoUpgrade &&
+		!cp.Spec.SandboxWorkloads.IsEnabled()
+
+	list := &corev1.NodeList{}
```
tariq1890:
```suggestion
	nodeList := &corev1.NodeList{}
```

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,475 @@
...
+		}
+	}
+	return modified
+}
+
+// applyDriverAutoUpgradeAnnotation sets or clears the driver auto-upgrade annotation on GPU nodes.
+func (nlc *nodeLabelingController) applyDriverAutoUpgradeAnnotation(ctx context.Context) error {
+	cp := nlc.clusterPolicy
+	autoUpgradeEnabled := (cp.Spec.Driver.IsEnabled() || cp.Spec.Driver.UseNvidiaDriverCRDType()) &&
+		cp.Spec.Driver.UpgradePolicy != nil &&
```
tariq1890: Let's factor these two lines out into a new helper method.

controllers/clusterpolicy_controller.go
cdesiniotis: The node-labeling-controller is now responsible for reconciling the `nvidia.com/mig.capable` label, which is why I removed it from here -- the ClusterPolicy controller does not need to reconcile on this event anymore. And now that I think about it, I can also remove `gpuCommonLabelOutdated` here as well since that is only relevant for the node-labeling-controller.

config/rbac/role.yaml
cdesiniotis: Sure, I can remove this and track it for a follow-up PR.

cdesiniotis: Reverted this change.

controllers/nodelabeling_controller.go
cdesiniotis: Added one.

controllers/nodelabeling_controller.go
cdesiniotis: Updated.

controllers/nodelabeling_controller.go
cdesiniotis: Done.

controllers/nodelabeling_controller.go
cdesiniotis: Updated.

controllers/nodelabeling_controller.go
cdesiniotis: I have also updated this method now to account for when the NVIDIADriver CRD is used. Since each NVIDIADriver has it's own upgradePolicy spec, we must reconcile the annotation per-NVIDIADriver CR instance.

controllers/clusterpolicy_controller.go
tariq1890: Sounds good, thanks for the context!

controllers/nodelabeling_controller.go
```diff
@@ -0,0 +1,510 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	gpuv1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1"
+	nvidiav1alpha1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1alpha1"
+	"github.com/NVIDIA/gpu-operator/internal/consts"
```
tariq1890, line 40:
```suggestion
```

cdesiniotis: Updated.

cdesiniotis:
I have added `nodelabeling_controller_test.go` which tests some of the private methods in `nodelabeling_controller.go`

tariq1890 (approved):

PR #2571, opened 2026-06-23
Add GPUCluster CRD and controller for DRA-based stack

karthikvetrivel:
## 1. Overview

We introduce a new CRD named `GPUCluster` and a new controller for reconciling it. Like `ClusterPolicy` today, it is a singleton, cluster-scoped CRD that configures the operands needed to enable GPUs in Kubernetes. `GPUCluster` represents the new DRA-based software-enablement stack; it is an evolution of `ClusterPolicy`. Unlike `ClusterPolicy`, it does not manage the driver or a device plugin: the driver is either preinstalled on the host or managed by `NVIDIADriver` CRs, and GPUs are surfaced to workloads through DRA. 

A `GPUCluster` may coexist with a `ClusterPolicy` in the same cluster, with every GPU node served by exactly one of the two stacks.

## 2. Testing

Unit tests cover the new API, operand rendering, controller status handling, and the node-labeling logic. I also tested this end-to-end on a 2-node heterogeneous cluster (A100 + T4, Kubernetes v1.34):

- GPUCluster-only install via Helm: all operands reconcile to ready. DRA allocation was validated for full GPUs, MIG, and adminAccess. The dra-driver-validator was verified against both host-installed and NVIDIADriver-managed drivers.
- Coexistence: ClusterPolicy and GPUCluster run in one cluster, with one node per plane. Each node runs only its plane's operands, with extended resources on the device-plugin node and ResourceSlices on the DRA node.
- Mode switching: flipping nvidia.com/gpu-operator.mode in either direction swaps the operand set. A shared NVIDIADriver-managed driver survives the switch without a reload. 
- Label lifecycle: driver-manager pause values survive reconciles. A node's previous-plane labels are cleaned up, and unrecognized gpu.deploy.* labels are left untouched.

For compute domains, I created a ComputeDomain CR with `numNodes: 2` and a channel ResourceClaimTemplate, then ran a 2-pod indexed Job with pod anti-affinity so one pod landed on each GB200 node. I gave each pod a resource claim against the channel template, and inside each container I listed `/dev/nvidia-caps-imex-channels/` to confirm containerd injected channel0 via CDI. Note that compute domains functionality requires containerd 2.x+.

## 3. Misc Design Notes

The driver daemonset deployed by the NVIDIADriver CR is not gated on the deploy `mode` so that it is not evicted and re-scheduled when the user switches modes.  

[...]

PR #2572, opened 2026-06-23
Support NVIDIADriver reconciliation and driver upgrades without a ClusterPolicy

karthikvetrivel:
> Stacked on #2571 (`kv-gpuclusterconfig-crd`). This PR's diff is just the two commits below. GitHub will retarget it onto `main` once #2571 lands.

## Summary

This PR lets the NVIDIADriver controller reconcile without a ClusterPolicy, so a GPUCluster (the DRA stack) can use operator-managed drivers standalone.

- The driver host root is now sourced via the info catalog instead of the ClusterPolicy CR. The driver renderer consumes a new `InfoTypeHostRoot` catalog entry rather than reading the ClusterPolicy directly, which decouples it from the `gpuv1` types. When both CRs exist, the ClusterPolicy's `hostPaths.rootFS` takes precedence. Otherwise the GPUCluster's value is used.
- Enabling DRA on a cluster requires that all driver management in that cluster is done through NVIDIADriver CRs.  A ClusterPolicy may coexist with a GPUCluster CR, but only if it delegates driver management by setting driver.useNvidiaDriverCRD=true.  A ClusterPolicy that owns driver daemonsets directly (useNvidiaDriverCRD=false) alongside a GPUCluster CR is an invalid configuration.  There is no supported topology where ClusterPolicy-owned drivers and NVIDIADriver-owned drivers run in the same cluster.

## Testing

There is unit coverage for the host-root sourcing and the gate transitions, plus two live validations on a single-node Tesla T4 cluster (Kubernetes 1.34, containerd 2.2).

- Standalone end-to-end: with no ClusterPolicy, a GPUCluster plus a default NVIDIADriver rolled out the driver DaemonSet. The DRA kubelet-plugin came up against the operator-installed driver root (`devRoot=/run/nvidia/driver`) and published its DeviceClasses and ResourceSlice. A pod with a DRA claim ran `nvidia-smi` against the operator-installed driver.
- Gate matrix, observed via `.status.state` transitions on a cluster converted from a released operator install. A ClusterPolicy that is not delegating with no GPUCluster produced `disabled`. A GPUCluster alone produced `ready`. With neither CR, the NVIDIADriver went `notReady` with the expected message. The ClusterPolicy plus GPUCluster coexistence path is covered by the unit tests.
- All transitions were watch-driven. Deleting or creating a GPUCluster flipped the NVIDIADriver state without manual intervention, and a single GPUCluster creation transitioned two NVIDIADriver instances at once.


controllers/nvidiadriver_controller.go
```diff
@@ -145,6 +127,17 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			}
 			return reconcile.Result{}, nil
 		}
+		hostRoot = clusterPolicy.Spec.HostPaths.RootFS
+	case gpuCluster != nil:
+		hostRoot = gpuCluster.Spec.HostPaths.RootFS
+	default:
+		err := fmt.Errorf("no ClusterPolicy or GPUCluster object found in the cluster")
+		logger.Error(err, "failed to get a cluster-wide configuration object")
```
tariq1890:
```suggestion
		logger.Error(err, "failed to retrieve hostPaths information. No ClusterPolicy or GPUCluster object found in the cluster")
```

controllers/upgrade_controller.go
```diff
@@ -87,32 +90,29 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-			// Return and don't requeue
-			return reconcile.Result{}, nil
-		}
-		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
 
+	if clusterPolicy == nil {
```
tariq1890: We can merge two if blocks into one with this change
```suggestion
	if clusterPolicy == nil || clusterPolicy.Spec.Driver.UseNvidiaDriverCRDType() {
```

controllers/nvidiadriver_controller.go
```diff
@@ -145,6 +127,17 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			}
 			return reconcile.Result{}, nil
 		}
+		hostRoot = clusterPolicy.Spec.HostPaths.RootFS
+	case gpuCluster != nil:
+		hostRoot = gpuCluster.Spec.HostPaths.RootFS
```
tariq1890: Correct me if I am wrong, but in the scenario where both ClusterPolicy and GPUCluster exist, will the `.spec.hostPaths.rootFS` from GPUCluster  have the higher order of precedence?

tariq1890:
Thanks @karthikvetrivel! Can you also squash your commit history here?

controllers/upgrade_controller.go
cdesiniotis: +1

cdesiniotis: Actually, isn't it possible now to have both ClusterPolicy-backed drivers and NVIDIADriver-backed drivers (we can continue to debate whether this is actually desired or not in a separate thread)? IIUC the upgrade-controller, as currently implemented, does not account for that use case -- it only calls `reconcileNVIDIADriverUpgrades()` OR `reconcileClusterPolicyDriverUpgrades`.

cdesiniotis: The  problematic scenario -- ClusterPolicy and GPUCluster exist; `clusterpolicy.spec.driver.useNvidiaDriverCRD=false`; NVIDIADriver CRs exist which target DRA nodes. In this case, the upgrade-controller will not be able to manage the upgrades for both the ClusterPolicy-owned driver daemonsets and the NVIDIADriver-owned driver daemonsets.

karthikvetrivel: Good catch. You are right that the upgrade controller does not handle that. 

As we discussed offline, we decided not to support it. To summaize, enabling DRA now requires that all driver management in the cluster is done through NVIDIADriver CRs. A ClusterPolicy with `useNvidiaDriverCRD=false` alongside a GPUCluster is treated as an invalid configuration.

controllers/nvidiadriver_controller.go
karthikvetrivel: ClusterPolicy takes precedence today. Now, I believe this combination is only valid when the ClusterPolicy sets `useNvidiaDriverCRD=true`, and in that case ClusterPolicy's `hostPaths.rootFS` is used.

controllers/nvidiadriver_controller.go
karthikvetrivel: Done, applied in the squashed commit.

controllers/nvidiadriver_controller.go
tariq1890: I see that this `hostRoot` value gets used [here](https://github.com/NVIDIA/gpu-operator/blob/f048f1e78c8e0457a368d879a05b7e0405ec946b/manifests/state-driver/0500_daemonset.yaml#L682)
```
- name: host-root
          hostPath:
            path: {{ .HostRoot | default "/" }}
```

Thinking about this more, it doesn't make much sense to make this global value and we are better off getting this value from the `NVIDIADriver` CR. For this, we would need to add a new field called `hostRoot` to the NVIDIADriver CR. This is the behaviour I envision in the following scenarios:

- If the `GPUCluster` resource is present (whether by itself or along with `ClusterPolicy`, then we get the hostRoot from the `NVIDIADriver` CR, else we fallback to the GPUCluster hostRoot. Here we will never get the hostRoot from `ClusterPolicy`. This way we ensure zero dependency on `ClusterPolicy` in a `GPUCluster` flow
- If only `ClusterPolicy` is present, then we either get the hostRoot from the `NVIDIADriver CR` (the newly added field), else we fallback to the `ClusterPolicy.Spec.HostPaths.RootFS`

This can be done in a follow-up PR

tariq1890: Note that when we add `hostRoot` to the NVIDIADriver CR, it'll be an additive change and it will be an optional field. If this is empty, it will be defaulted to `"/"` as we currently do today

cdesiniotis: Note, the host root is needed for many of our daemonsets, not just the driver. So we still need this field to be present in the `GPUCluster` CR. What is the benefit in having users configure this field in multiple places (GPUCluster and NVIDIADriver CRs)?

tariq1890: Yes, the rest of the `HostPaths` fields are used correctly in operands whether they come from GPUCluster or ClusterPolicy. For the driver daemonset however, it's just the `.HostPaths.RootFS` field that is relevant

rajathagasthya: I'm in favor of moving `hostRoot` to `NVIDIADriver` and removing it entirely from `GPUCluster`. 

In the legacy `ClusterPolicy` path, we can get the `hostRoot` from either `NVIDIADriver` CR (if specified) or from `ClusterPolicy.Spec.HostPaths.RootFS` when `useNvidiaDriverCRD=true`, and from `ClusterPolicy.Spec.HostPaths.RootFS` only when `useNvidiaDriverCRD=false`.

tariq1890: After discussing internally, we have come to a consensus on how to move forward with this. The main issue we that we want to address (in a follow-up PR) is that there should be no cross-CR or circular dependency between the custom resources.

- The hostRoot will be hard-coded as `/` for now.
- We had between two approaches to fix this problem. Either adding a new hostRoot field to the `NVIDIADriver` CRD or adding a new environment variable to determine the hostRoot to be set in both the driver daemonset and the DRA driver kubelet plugin daemonset. This is still ongoing.

controllers/gpucluster_controller.go
```diff
@@ -162,6 +178,22 @@ func (r *GPUClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 }
 
+// validatePrerequisites checks the cross-CR rules that gate DRA enablement, returning
+// a message describing the first unmet prerequisite or an empty string when all are met.
+func (r *GPUClusterReconciler) validatePrerequisites(ctx context.Context) (string, error) {
+	clusterPolicies := &gpuv1.ClusterPolicyList{}
+	if err := r.List(ctx, clusterPolicies); err != nil {
+		return "", fmt.Errorf("error listing ClusterPolicy objects: %w", err)
+	}
+	for i := range clusterPolicies.Items {
```
tariq1890: Does it make sense to loop through every possible clusterPolicy? Don't we only need to care about the  active/singleton clusterPolicy?

karthikvetrivel: I was thinking about this as well, but we don't ever enforce a singleton ClusterPolicy. You're right — there is only one active ClusterPolicy and extra CPs get status.state = Ignored, but only asynchronously after the CP controller reconciles them so filtering on status could lead to a potential races.

karthikvetrivel: This is somewhat related to the TODO above `resolveActiveConfig`, which we can address in a follow up.

controllers/nvidiadriver_controller.go
karthikvetrivel: Done.  The host root is now hard-coded to `/` whenever a GPUCluster is present, including when a ClusterPolicy coexists with it, so NVIDIADriver never reads it from another CR. Only the ClusterPolicy-only legacy path still sources `spec.hostPaths.rootFS`. Also added a ClusterPolicy-only test case for that path.

controllers/gpucluster_controller.go
tariq1890: Can you add the TODO here too?

tariq1890:
```suggestion
	for _, clusterPolicy := range clusterPolicies.Items {
```

controllers/nvidiadriver_controller.go
```diff
@@ -98,75 +99,67 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-		logger.Error(err, "failed to get ClusterPolicy object")
+	var hostRoot string
+	switch {
+	case gpuCluster != nil:
+		hostRoot = "/"
+	case clusterPolicy != nil:
+		hostRoot = clusterPolicy.Spec.HostPaths.RootFS
+	default:
+		err := fmt.Errorf("no ClusterPolicy or GPUCluster object found in the cluster")
+		logger.Error(err, "failed to retrieve hostPaths information. No ClusterPolicy or GPUCluster object found in the cluster")
```
cdesiniotis: Question -- should we not default to `/` in this case too? That way, one could deploy NVIDIADriver CRs without having to deploy either a ClusterPolicy or GPUCluster first?

controllers/gpucluster_controller.go
karthikvetrivel: Done.

karthikvetrivel: Done.

controllers/nvidiadriver_controller.go
karthikvetrivel: Yes, I think so. I added it.

cdesiniotis (approved):

controllers/nvidiadriver_controller.go
```diff
@@ -98,53 +99,37 @@ func (r *NVIDIADriverReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-				logger.Error(condErr, "failed to set condition")
-			}
-			return reconcile.Result{}, nil
-		}
+
+	// The host root is hard-coded to "/" to avoid a cross-CR dependency; only the
+	// legacy ClusterPolicy-only path reads spec.hostPaths.rootFS.
+	// TODO: support a custom host root via a new NVIDIADriver API field or an
+	// operator environment variable.
+	hostRoot := "/"
```
tariq1890: please move this to a constant so that this assumption is more visible. You can annotate that constant with the comments you have written here.

tariq1890: Create a new constant named `defaultHostRoot`

karthikvetrivel: Yep, good callout. Done.

rajathagasthya (approved):

cdesiniotis (approved):

tariq1890 (approved):

PR #2573, opened 2026-06-23
Add Cluster Autoscaler integration example with NPD and NRC

rajathagasthya:
## Description

When the Cluster Autoscaler adds a GPU node, the node reports Ready long before the driver, container toolkit, and device plugin are installed. Workloads scheduled in that window fail or occupy the node so the autoscaler considers the scale-up satisfied.

This adds a reference guide and example manifests that gate scheduling on GPU readiness using upstream components: the node pool template applies a startup taint, Node Problem Detector publishes a `nvidia.com/GPUReady` node condition from an nvidia-smi probe, and the Node Readiness Controller removes the taint once the condition is True. The GPU Operator itself is unchanged; its operands only need tolerations via the existing `daemonsets.tolerations` and NFD worker toleration values.

The example includes a kind-based simulation for clusters without GPUs: workers join pre-tainted via kubelet registration (matching node pool template semantics), a marker file stands in for the nvidia-smi probe, and a node-scaler script vendored from the NRC repo simulates a scale-up by adding a fresh worker to the running cluster.

Known limitation documented in the guide: NRC requires the `readiness.k8s.io/` taint prefix while managed autoscalers on GKE and AKS only support the reserved startup-taint key prefix, so the pattern currently requires a self-managed Cluster Autoscaler; kubernetes-sigs/node-readiness-controller#279 tracks this issue.

## Checklist

- [ ] Lint checks passing (`make lint`) — N/A, documentation and example manifests only
- [ ] Generated assets in-sync (`make validate-generated-assets`) — N/A, no generated assets affected
- [ ] Go mod artifacts in-sync (`make validate-modules`) — N/A, no Go changes
- [ ] Test cases are added for new code paths — N/A, no code paths added

## Testing

Manually verified on a kind cluster:

- Walkthrough A (simulation): NPD publishes `nvidia.com/GPUReady`, NRC holds the startup taint while the condition is False and removes it once it turns True, and the gated pod then schedules. `reset.sh` re-arms the node.
- New workers, including one added via `kindscaler.sh`, join with the startup taint and `nvidia.com/gpu.present` label already applied.

### DIFF
--- a/examples/cluster-autoscaler/README.md
+++ b/examples/cluster-autoscaler/README.md
@@ -1,0 +1,659 @@
+# Cluster Autoscaler Integration with GPU Operator
+
+This guide shows how to keep workloads off autoscaled GPU nodes until the GPU
+stack is actually ready, using a startup taint that is removed when the node
+passes a GPU readiness probe.
+
+When the Cluster Autoscaler adds a GPU node, the node reports `Ready` long
+before the GPU Operator has finished installing the driver, container toolkit,
+and device plugin. Workloads scheduled during that window fail, or occupy the
+node so the autoscaler considers the scale-up satisfied. Scaling a GPU pool
+up from zero can stall for a separate reason, covered in the Scale-from-zero
+section below.
+
+The integration adds three components; the GPU Operator itself is unchanged:
+
+| Component | Role |
+|---|---|
+| Node pool template (cloud provider) | Applies the startup taint to every new GPU node |
+| [Node Problem Detector (NPD)](https://github.com/kubernetes/node-problem-detector) | Runs a GPU readiness probe on each GPU node and publishes the `nvidia.com/GPUReady` node condition |
+| [Node Readiness Controller (NRC)](https://github.com/kubernetes-sigs/node-readiness-controller) | Removes the startup taint once the condition is `True` |
+| GPU Operator | Unchanged; its operands tolerate the startup taint via existing toleration settings |
+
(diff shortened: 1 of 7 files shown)


rajathagasthya:
Should a short version of this be in the documentation? I think this detailed guide under `docs/` is valuable. cc @a-mccarthy @cdesiniotis @tariq1890

examples/cluster-autoscaler/README.md
```diff
@@ -0,0 +1,484 @@
...
+On a cluster that already has GPU nodes, preview the rule's effect first.
+NRC adds the taint to any matching node whose condition is not `True` — in
+both enforcement modes — so applying the rule before NPD reports readiness
+on every existing node makes those nodes unschedulable for new pods. Set
+`dryRun: true` in the rule spec; the controller then reports intended taint
+changes in the rule's `status.dryRunResults` without modifying nodes:
+
+```sh
+kubectl apply -f node-readiness-rule.yaml
+kubectl get nodereadinessrule nvidia-gpu-readiness -o jsonpath='{.status.dryRunResults}'
```
rahulait, line 193: nit: should this example be updated to have dryrun in it?

rahulait (approved):
Overall, LGTM

examples/cluster-autoscaler/README.md
rajathagasthya: Not sure I follow. Do you mean dry run for applying the NRR?

PR #2582, opened 2026-06-25
Add upgradePolicy to NVIDIADriver CRD

cdesiniotis:
The upgradePolicy influences how the driver-upgrade controller upgrades GPU driver daemonsets. Adding this field to the NVIDIADriver CRD allows users to define different upgrade policies for different NVIDIADriver CRs. If nil or empty, the driver-upgrade controller will fallback to using a default upgradePolicy defined in the code which aligns with the defaults in our helm values.

### Testing

I tested the following scenarios on a k8s cluster with two GPUs:

1. Install gpu operator with default options. Verify entire stack comes up.
2. Upgrade ClusterPolicy-managed driver daemonset. Verify upgrade controller uprades driver pods in rolling fashion based on configured upgrade policy.
3. Migrate to NVIDIADriver-managed driver daemonset by a) setting `driver.useNvidiaDriverCRD=true` in ClusterPolicy; b) creating a default NVIDIADriver CR. Verify ClusterPolicy-managed pods gets orphaned then deleted by upgrade controller in a rolling fashion.
4. Upgrade the default NVIDIADriver-managed driver daemonset by changing `spec.version`. Verify upgrade controller upgrades pods in a rolling fashion.
5. Repeat step 4 with a different upgradePolicy configured in the NVIDIADriver CR. Verify upgrade controller upgrades pods according to the new upgrade policy.


api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,86 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
 	}
 	return *l.NLSEnabled
 }
+
```
rahulait, line 791: Should we move the constants to the top of this file?

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
```diff
@@ -13,6 +13,28 @@ spec:
...
+    # a percentage of total nodes at the start of upgrade (ex:
+    # 10%). Absolute number is calculated from percentage by rounding
+    # up. By default, a fixed value of 25% is used.'
+    maxUnavailable: 25%
+    # options for waiting on pod(job) completions
+    waitForCompletion:
+      timeoutSeconds: 0
+      podSelector: ""
+    # options for gpu pod deletion
+    gpuPodDeletion:
```
rahulait: should this be `podDeletion`?

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,86 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+			Enable:         defaultDrainEnable,
+			Force:          defaultDrainForce,
+			PodSelector:    defaultDrainPodSelector,
+			TimeoutSecond:  defaultDrainTimeoutSecond,
+			DeleteEmptyDir: defaultDrainDeleteEmptyDir,
+		}
+	}
+
+	return policy
+}
```
tariq1890: Consider using a library to merge structs instead. It addresses merging of default values at higher granular levels. https://github.com/darccio/mergo

controllers/upgrade_controller.go
```diff
@@ -202,6 +212,139 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	)
+
+	nvidiaDriverList := &nvidiav1alpha1.NVIDIADriverList{}
+	if err := r.List(ctx, nvidiaDriverList); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Check if any NVIDIADriver has auto upgrade enabled
+	anyAutoUpgradeEnabled := false
+	for i := range nvidiaDriverList.Items {
```
tariq1890:
```suggestion
	for _, nvd := range nvidiaDriverList.Items {
```

controllers/upgrade_controller.go
```diff
@@ -202,6 +212,139 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		nvd := &nvidiaDriverList.Items[i]
+		upgradePolicy := nvd.Spec.GetUpgradePolicyWithDefaults()
+		if upgradePolicy.AutoUpgrade {
+			anyAutoUpgradeEnabled = true
+			break
+		}
+	}
+
+	if !anyAutoUpgradeEnabled {
```
tariq1890: nit (non-blocking): can you change the `!anyAutoUpgradeEnabled` or `noAutoUpgradesEnabled`? Easier to read than an boolean flag inverted if check.

controllers/upgrade_controller.go
```diff
@@ -202,6 +212,139 @@ func (r *UpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			if statesByNVD[ownerName] == nil {
+				s := upgrade.NewClusterUpgradeState()
+				statesByNVD[ownerName] = &s
+			}
+			statesByNVD[ownerName].NodeStates[stateKey] = append(statesByNVD[ownerName].NodeStates[stateKey], nodeState)
+		}
+	}
+
+	// Apply the upgrade policy for each NVIDIADriver instance using its partitioned cluster upgrade state
+	for i := range nvidiaDriverList.Items {
```
tariq1890:
```suggestion
	for _, nvd := range nvidiaDriverList.Items {
```

controllers/upgrade_controller.go
```diff
@@ -231,6 +374,33 @@ func (r *UpgradeReconciler) removeNodeUpgradeStateLabels(ctx context.Context) er
...
+
+	nodeList := &corev1.NodeList{}
+	if err := r.List(ctx, nodeList, client.MatchingLabels{gpuconsts.NVIDIADriverOwnerLabel: nvdName}); err != nil {
+		r.Log.Error(err, "Failed to list nodes for NVIDIADriver", "name", nvdName)
+		return err
+	}
+
+	upgradeStateLabel := upgrade.GetUpgradeStateLabelKey()
+
+	for i := range nodeList.Items {
```
tariq1890:
```suggestion
	for _, node := range nodeList.Items {
```

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: I tried this out and am not sure if this will work for our use case. For our scalar fields, like `autoUpgrade` and `maxParallelUpgrades` our default values are not the zero-values for their respective data types, e.g. `autoUpgrade=true` and `maxParallelUpgrades=1` by default. When merging the user-supplied config and the default config structs, `mergo` is not able to address the scenario where a user explicitly sets `autoUpgrade` to `false` or `maxParallalUpgrades` to `0` (the zero-values for these fields).

For example, the below code:
```
cfg = getDefaultCfg()
src = &upgrade_v1alpha1.DriverUpgradePolicySpec{AutoUpgrade: false, MaxParallelUpgrades: 0}
mergo.Merge(&cfg, src, mergo.WithOverride, mergo.WithoutDereference)
spew.Dump(cfg)
```
prints out
```
(*v1alpha1.DriverUpgradePolicySpec)(0x20fb78ae2ed0)({
 AutoUpgrade: (bool) true,
 MaxParallelUpgrades: (int) 1,
 MaxUnavailable: (*intstr.IntOrString)(0x20fb789fa580)(25%),
 PodDeletion: (*v1alpha1.PodDeletionSpec)(0x20fb7898a3d8)({
  Force: (bool) false,
  TimeoutSecond: (int) 300,
  DeleteEmptyDir: (bool) false
 }),
 WaitForCompletion: (*v1alpha1.WaitForCompletionSpec)(0x20fb78980630)({
  PodSelector: (string) "",
  TimeoutSecond: (int) 0
 }),
 DrainSpec: (*v1alpha1.DrainSpec)(0x20fb78ae2f00)({
  Enable: (bool) false,
  Force: (bool) false,
  PodSelector: (string) "",
  TimeoutSecond: (int) 300,
  DeleteEmptyDir: (bool) false
 })
})
```

controllers/upgrade_controller.go
rahulait, line 387: Since now upgrade related config is in nvidiadriver CR, should the upgrade_controller now watch NVIDIADriver objects as well? Since the new behavior is controlled by NVIDIADriver.spec.upgradePolicy, changing or enabling that policy will not enqueue the upgrade controller unless an unrelated event happens.

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,86 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+	defaultDrainDeleteEmptyDir               = false
+)
+
+// GetUpgradePolicyWithDefaults returns the upgrade policy for this driver with default
+// values applies for any missing fields.
+func (s *NVIDIADriverSpec) GetUpgradePolicyWithDefaults() *upgrade_v1alpha1.DriverUpgradePolicySpec {
+	if s.UpgradePolicy == nil {
+		return getDefaultUpgradePolicy()
+	}
+
```
rahulait, line 827: Default value for autoUpgrade is false: https://github.com/NVIDIA/k8s-operator-libs/blob/fa6a3643c44164619e5a941c53d653fe967d39ef/api/upgrade/v1alpha1/upgrade_spec.go#L31-L32

In our defaults, we are setting it to true. So if there is no upgradePolicy, we default to true. However, if there is upgradePolicy but default is not set, its getting set to false. Should it be also set to true?

cdesiniotis: I had a similar question around this as well.

The reason why I decided to set `autoUpgrade` to `true` by default in our code is because this is what we set in our helm [values](https://github.com/NVIDIA/gpu-operator/blob/197a7dd03e1b37ac665ea1769adff94906f31597/deployments/gpu-operator/values.yaml#L167) and [sample CR](https://github.com/NVIDIA/gpu-operator/blob/197a7dd03e1b37ac665ea1769adff94906f31597/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L48) in OLM by default. I want to maintain this default experience (which is that `autoUpgrade` is enabled unless users explicitly opt-out of it).

As you said, if users omit setting `upgradePolicy` entirely, our code will set `upgradePolicy.autoUpgrade=true` (this is desirable imo). However, if a user does set `upgradePolicy` but forgets to set the `autoUpgrade` field, the default value from the CRD will be applied (which is false). I am not sure we can do anything about this. We could consider changing the default value in the CRD, but I would assume that is a breaking change?

rahulait: Ah, rest of the fields we are setting are all pointers and this one is not. So we cannot say if its false, its by default or the user has set it to false. Yeah, kind of tricky to decide when to set it to true.

rahulait: If a user does:
```
spec:
  upgradePolicy:
    maxUnavailable: 10%
```
They might think that autoUpgrade will be true but since the upgrade controller's CRD defaults to false, it will be disabled by default. I think if we really want to fix this gap, one option is to have our own copy of upgradepolicy spec like NVIDIADriverUpgradePolicySpec and in that default to true for autoUpgrade.

cdesiniotis: >one option is to have our own copy of upgradepolicy spec like NVIDIADriverUpgradePolicySpec and in that default to true for autoUpgrade.

I am not opposed to this. It just means that if / when this spec is updated in `k8s-operator-libs`, we'd have to keep this up-to-date (updates to this spec have been very infrequent since we first introduced it). @tariq1890 thoughts?

controllers/upgrade_controller.go
cdesiniotis: I considered this. The reason why I opted to exclude this was that a change in upgrade-policy will get picked up whenever an actual upgrade is being reconciled. So IMO it is not critical to reconcile immediately following a change to an NVIDIADriver CR's upgrade policy. But I don't have a strong opinion here.

controllers/upgrade_controller.go
cdesiniotis: Updated.

controllers/upgrade_controller.go
cdesiniotis: Updated to `noAutoUpgradesEnabled`.

controllers/upgrade_controller.go
cdesiniotis: Updated.

controllers/upgrade_controller.go
cdesiniotis: Updated.

config/samples/nvidia_v1alpha1_nvidiadriver.yaml
cdesiniotis: Good catch! Updated.

api/nvidia/v1alpha1/nvidiadriver_types.go
tariq1890: Got it. Thanks for checking.

controllers/upgrade_controller.go
rahulait: Yeah, sounds good.

api/nvidia/v1alpha1/nvidiadriver_types.go
tariq1890: > I think if we really want to fix this gap, one option is to have our own copy of upgradepolicy spec like NVIDIADriverUpgradePolicySpec and in that default to true for autoUpgrade.

I am okay with this.

cdesiniotis: I have updated this now. We now define our own local `DriverUpgradePolicySpec` type where the default value for `autoUpgrade` is `true` (instead of `false`). I also added some logic to convert between `DriverUpgradePolicySpec` and `upgrade_v1alpha1.DriverUpgradePolicySpec` (defined in k8s-operator-libs).

There is precedent already for doing this. Network Operator also follows this approach, see:
https://github.com/Mellanox/network-operator/blob/747df8b563011cdd909e166362407843745d69da/api/v1alpha1/nicclusterpolicy_types.go#L156
https://github.com/Mellanox/network-operator/blob/747df8b563011cdd909e166362407843745d69da/controllers/upgrade_controller.go#L147-L159

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: I have removed the constants altogether in the latest push.

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,164 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+	// +optional
+	// +kubebuilder:default="25%"
+	MaxUnavailable    *intstr.IntOrString    `json:"maxUnavailable,omitempty"`
+	PodDeletion       *PodDeletionSpec       `json:"podDeletion,omitempty"`
+	WaitForCompletion *WaitForCompletionSpec `json:"waitForCompletion,omitempty"`
+	DrainSpec         *DrainSpec             `json:"drain,omitempty"`
+}
+
+// WaitForCompletionSpec describes the configuration for waiting on job completions.
+type WaitForCompletionSpec struct {
```
rahulait: Can we do like:

```
type PodDeletionSpec = upgradev1alpha1.PodDeletionSpec
type WaitForCompletionSpec = upgradev1alpha1.WaitForCompletionSpec
type DrainSpec = upgradev1alpha1.DrainSpec

type DriverUpgradePolicySpec struct {
    // AutoUpgrade is a global switch for automatic upgrade feature.
    // If set to false all other options are ignored.
    // +optional
    // +kubebuilder:default=true
    AutoUpgrade bool `json:"autoUpgrade,omitempty"`

    // +optional
    // +kubebuilder:default=1
    // +kubebuilder:validation:Minimum=0
    MaxParallelUpgrades int `json:"maxParallelUpgrades,omitempty"`

    // +optional
    // +kubebuilder:default="25%"
    MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`

    PodDeletion       *PodDeletionSpec       `json:"podDeletion,omitempty"`
    WaitForCompletion *WaitForCompletionSpec `json:"waitForCompletion,omitempty"`
    DrainSpec         *DrainSpec             `json:"drain,omitempty"`
}
```

cdesiniotis: Yeah makes sense. Updated as you have suggested.

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,112 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+		DrainSpec:           getDefaultDrainSpec(),
+	}
+}
+
+func getDefaultMaxUnavailable() *intstr.IntOrString {
+	defaultMaxUnavailable := intstr.FromString("25%")
+	return &defaultMaxUnavailable
+}
+
+func getDefaultPodCompletionSpec() *upgrade_v1alpha1.PodDeletionSpec {
```
tariq1890:
```suggestion
func getDefaultPodCompletionSpec() *PodDeletionSpec {
```

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,112 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+	return &upgrade_v1alpha1.PodDeletionSpec{
+		Force:          false,
+		TimeoutSecond:  300,
+		DeleteEmptyDir: false,
+	}
+}
+
+func getDefaultWaitForCompletionSpec() *upgrade_v1alpha1.WaitForCompletionSpec {
```
tariq1890:
```suggestion
func getDefaultWaitForCompletionSpec() *WaitForCompletionSpec {
```

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,112 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+	return &upgrade_v1alpha1.WaitForCompletionSpec{
+		PodSelector:   "",
+		TimeoutSecond: 0,
+	}
+}
+
+func getDefaultDrainSpec() *upgrade_v1alpha1.DrainSpec {
```
tariq1890:
```suggestion
func getDefaultDrainSpec() *DrainSpec {
```

api/nvidia/v1alpha1/nvidiadriver_types.go
tariq1890: Not a strong opinion, but what if you used this as the alias for the shared UpgradePolicy struct?

```
type UpgradePolicySpec = upgrade_v1alpha1.DriverUpgradePolicySpec
```

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,112 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+func getDefaultPodCompletionSpec() *upgrade_v1alpha1.PodDeletionSpec {
+	return &upgrade_v1alpha1.PodDeletionSpec{
+		Force:          false,
+		TimeoutSecond:  300,
+		DeleteEmptyDir: false,
+	}
+}
+
+func getDefaultWaitForCompletionSpec() *upgrade_v1alpha1.WaitForCompletionSpec {
+	return &upgrade_v1alpha1.WaitForCompletionSpec{
```
tariq1890: let's apply similar changes in the other places.

```suggestion
	return &WaitForCompletionSpec{
```

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: Not sure I follow how that is possible, considering we are updating the default value of the `autoUpgrade` field.

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: Updated

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: Updated

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: Updated

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: Updated throughout

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,112 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
 	}
 	return *l.NLSEnabled
 }
+
+// DriverUpgradePolicySpec describes policy configuration for automatic upgrades of the driver.
+type DriverUpgradePolicySpec struct {
+	// AutoUpgrade is a global switch for automatic upgrade feature.
```
tariq1890: Let's drop the use of `global` here.

cdesiniotis: Updated.

tariq1890 (approved):

rahulait:
LGTM. I am wondering about backward compatibility. Existing user-defined nvidiadriver CRs won't have upgrade configuration, and previously they were defaulting to the config in clusterpolicy. If custom upgrade config was used in clusterpolicy previously and user specified nvidiadriver was using it and now user upgrades to newer version, nvidiadriver will start using default config we specify. Would we be documenting this as a breaking change or something for users to be aware of when they are using nvidiadrivers and they jump to v26.7.0?

rahulait (approved):

api/nvidia/v1alpha1/nvidiadriver_types.go
```diff
@@ -780,3 +788,112 @@ func (l *DriverLicensingConfigSpec) IsNLSEnabled() bool {
...
+		DrainSpec:           getDefaultDrainSpec(),
+	}
+}
+
+func getDefaultMaxUnavailable() *intstr.IntOrString {
+	defaultMaxUnavailable := intstr.FromString("25%")
+	return &defaultMaxUnavailable
+}
+
+func getDefaultPodCompletionSpec() *PodDeletionSpec {
```
rajathagasthya:
```suggestion
func getDefaultPodDeletionSpec() *PodDeletionSpec {
```

rajathagasthya: optional: my style preference is to remove the `get` from function names. `Get/Set` reminds me of Java :)

controllers/upgrade_controller.go
```diff
@@ -219,6 +353,32 @@ func (r *UpgradeReconciler) removeNodeUpgradeStateLabels(ctx context.Context) er
...
+	}
+
+	upgradeStateLabel := upgrade.GetUpgradeStateLabelKey()
+
+	for _, node := range nodeList.Items {
+		if _, present := node.Labels[upgradeStateLabel]; !present {
+			continue
+		}
+		delete(node.Labels, upgradeStateLabel)
+		if err := r.Update(ctx, &node); err != nil {
```
rajathagasthya: Any reason why we're not using `Patch` here?

rajathagasthya (commented):
Just a couple of minor comments. Overall LGTM.

cdesiniotis:
> LGTM. I am wondering about backward compatibility. Existing user-defined nvidiadriver CRs won't have upgrade configuration, and previously they were defaulting to the config in clusterpolicy. If custom upgrade config was used in clusterpolicy previously and user specified nvidiadriver was using it and now user upgrades to newer version, nvidiadriver will start using default config we specify. Would we be documenting this as a breaking change or something for users to be aware of when they are using nvidiadrivers and they jump to v26.7.0?

Good observation. If users are depending on a custom upgradePolicy from ClusterPolicy, then yes, this would be a breaking change for user-created NVIDIADriver CRs. We will need to call this out in our release notes.

api/nvidia/v1alpha1/nvidiadriver_types.go
cdesiniotis: Good catch! Updated.

controllers/upgrade_controller.go
cdesiniotis: No particular reason. This was copied from the removeNodeUpgradeStateLabels() method. I have updated both to use `Patch` instead.

rajathagasthya (approved):

PR #2584, opened 2026-06-26
Decouple OperatorMetrics object from ClusterPolicyController object

cdesiniotis:
With this commit, any controller in the GPU Operator that needs to set operator metrics can do so without relying on state from the ClusterPolicy controller object. This will be especially relevant when we integrate the NVIDIA DRA Driver for GPUs since we will be introducing a new CRD that serves as an alternative to the ClusterPolicy CRD.

### DIFF
--- a/cmd/gpu-operator/main.go
+++ b/cmd/gpu-operator/main.go
@@ -154,11 +154,16 @@ func main() {
 	}
 
 	ctx := ctrl.SetupSignalHandler()
+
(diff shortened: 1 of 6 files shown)


rajathagasthya (approved):

karthikvetrivel (approved):
LGTM!

rahulait (approved):
LGTM

tariq1890 (approved):

PR #2599, opened 2026-07-04
fix(vgpu-device-manager): wait for host-installed vGPU Manager readiness

lexfrei:
## Description

The `vgpu-manager-validation` init container of the `nvidia-vgpu-device-manager` DaemonSet waited only for the `vgpu-manager-ready` status file, which the validator writes when the vGPU Manager is deployed as a container. When the vGPU Manager driver is pre-installed on the host (`driver.enabled=false`), the validator instead writes `host-vgpu-manager-ready` (and deletes both files at the start of each run), so the init container blocked indefinitely on "waiting for NVIDIA vGPU Manager to be setup" and the vGPU Device Manager never started.

`host-vgpu-manager-ready` is written by the validator but read nowhere else in the tree, so the host-installed-driver path never had a working readiness gate. There was already a `TODO` next to the gate acknowledging this.

This makes the init container wait for either status file, so the operand starts in both the container-managed and host-installed driver modes, and resolves that `TODO`.

## Checklist

## Testing

Added `TestVGPUDeviceManagerReadinessGate` (controllers package) which decodes the daemonset asset and asserts the init container's readiness gate waits for both `vgpu-manager-ready` and `host-vgpu-manager-ready`, combined with `||` (OR) so it passes when either file is present. Verified red/green: the test fails on the previous single-file gate and on an accidental `&&`, and passes with the fix. `go test ./controllers/`, `go vet ./controllers/`, and `golangci-lint run` on the changed package all pass. This changes only a static asset and a unit test, so `api/`, `config/`, `bundle/`, `deployments/`, and the module files are untouched.

### DIFF
--- a/assets/state-vgpu-device-manager/0600_daemonset.yaml
+++ b/assets/state-vgpu-device-manager/0600_daemonset.yaml
@@ -23,9 +23,11 @@ spec:
         - name: vgpu-manager-validation
           image: "FILLED BY THE OPERATOR"
           command: ['sh', '-c']
-          # TODO: Account for pre-installed vGPU Manager. Currently validator
-          # creates a different status file when driver is pre-installed.
-          args: ["until [ -f /run/nvidia/validations/vgpu-manager-ready ]; do echo waiting for NVIDIA vGPU Manager to be setup; sleep 5; done"]
+          # The validator writes vgpu-manager-ready when the vGPU Manager is
+          # deployed as a container, and host-vgpu-manager-ready when the vGPU
+          # Manager driver is pre-installed on the host. Wait for either so this
+          # operand starts in both cases.
+          args: ["until [ -f /run/nvidia/validations/vgpu-manager-ready ] || [ -f /run/nvidia/validations/host-vgpu-manager-ready ]; do echo waiting for NVIDIA vGPU Manager to be setup; sleep 5; done"]
           securityContext:
             privileged: true
           volumeMounts:
--- a/controllers/object_controls_test.go
+++ b/controllers/object_controls_test.go
@@ -65,6 +65,7 @@ const (
 	devicePluginAssetsPath        = "assets/state-device-plugin/"
 	dcgmExporterAssetsPath        = "assets/state-dcgm-exporter/"
 	migManagerAssetsPath          = "assets/state-mig-manager/"
+	vGPUDeviceManagerAssetsPath   = "assets/state-vgpu-device-manager/"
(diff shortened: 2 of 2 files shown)


lexfrei:
The failing `e2e-tests / e2e-tests-nvidiadriver` job did not execute any gpu-operator code. It failed while Holodeck was provisioning the EC2 GPU runner, before the cluster came up:

> InsufficientInstanceCapacity: We currently do not have sufficient g4dn.xlarge capacity in the Availability Zone you requested (us-west-1a).

The test VM was never created, so the operator was never installed and the vGPU validator this PR touches never ran. The `e2e-tests-containerd` job on the same commit passed. Could a maintainer re-run the nvidiadriver job? Thanks.

lexfrei:
Update, superseding my earlier note: the re-run got past Holodeck provisioning and stood up the cluster and operator, but `e2e-tests-nvidiadriver` still fails — and it does not look related to this PR.

What the log shows: the operator Helm-installs cleanly (`STATUS: deployed`), then the driver pod is never created. The test polls the `test-operator` namespace for ~50 min ("No resources found in test-operator namespace", 542×) and times out. The operator itself is healthy — `e2e-tests-containerd` passes on the same operator image, and no reconcile/parse error (or any reference to this PR's asset) appears in the log.

The same signature shows up on unrelated changes:
- #2601 (validator SR-IOV VFs, a different diff): nvidiadriver fails identically, containerd passes.
- #2571 (GPUCluster golden render tests, not mine): same — nvidiadriver fails, containerd passes.
- `main` ran nvidiadriver green earlier the same day (run 29751380127, ~15:33 UTC); every PR run after ~16:46 UTC fails the same way.

So this looks like an environmental break in the nvidiadriver e2e — the driver DaemonSet is never created — that appeared after main was last green, not a regression from this change. The operator-side logs weren't archived in the run ("No files were found ... ./logs/"), so I can't yet pin the exact upstream cause; glad to dig further if the captured cluster logs are available somewhere.

tariq1890:
@lexfrei Can you rebase your PR? I think the e2e test failure may be related to a bug introduced in our `main` branch which we have since fixed.

lexfrei:
Rebased onto current main (also rebased #2601, which hit the same nvidiadriver e2e failure). Thanks for the pointer!

tariq1890:
Thanks @lexfrei, this looks good to me. I'd like to get an approve from @cdesiniotis before merging this.

cdesiniotis (approved):
Thanks for the contribution @lexfrei

Issue #2600, opened 2026-07-04
[Feature]: Persist MIG mode and SR-IOV vGPU VFs across reboot on the sandbox/vGPU path (operator-driven GPU reset)


lexfrei:
### Motivation

On the sandbox / vGPU workload path, changing MIG mode and creating SR-IOV vGPU VFs does not survive a node reboot, and the operator does not re-establish this state on boot. After a reboot of an NVSwitch node configured for MIG-backed vGPU:

- MIG mode / MIG instances are not present until re-applied;
- SR-IOV VFs (created via `sriov-manage -e`) are gone — VFs are runtime state and are never persisted by SR-IOV;
- the vGPU devices created on those VFs are gone.

The operator's MIG manager can enable MIG mode, but committing a MIG-mode change on these GPUs requires a GPU reset. The only reset mechanism the Kubernetes mig-manager has is `WITH_REBOOT` ("reboot the node if changing the MIG mode fails for any reason") — it never performs a targeted `nvidia-smi --gpu-reset`. And VF creation (`sriov-manage`) is not performed by the operator at all — `sriov-manage` appears nowhere in `gpu-operator`. So on the vGPU path, after a reboot, MIG + VFs + vGPUs have to be re-established out-of-band (host units that enable MIG, run `--gpu-reset`, run `sriov-manage -e` per PF, then release the operand readiness gate).

By contrast, the **systemd** deployment of `mig-parted` does handle reboot persistence: `nvidia-mig-manager.service` persists the selected config across reboot (`persist_config_across_reboot`) and orders itself via `nvidia-gpu-reset.target` (`After=nvidia-fabricmanager.service`, `Before=nvidia-gpu-reset.target`) so MIG reconfiguration, GPU reset, and driver-service ordering are coordinated at boot. The Kubernetes / operator path has no equivalent.

### Proposal / discussion

1. Give the operator a way to commit MIG-mode changes via a targeted GPU reset instead of only a full node reboot (`WITH_REBOOT`). This likely needs coordination with `nvidia-persistenced` (see #118), which holds a handle that blocks `--gpu-reset`.
2. On the vGPU / sandbox path, have the operator own re-establishing SR-IOV VFs on boot (the equivalent of `sriov-manage -e` per PF) and re-applying the MIG + vGPU device configuration, so a reboot recovers hands-off without host units. This is the smallest useful seam and could be a first change on its own (VF re-enablement in the vGPU device-manager operand before it applies the vGPU config).
[...]

PR #2601, opened 2026-07-04
feat(validator): re-enable SR-IOV vGPU VFs before waiting for them

lexfrei:
## Description

On the vGPU (sandbox) workload path, SR-IOV Virtual Functions (VFs) are runtime state that does not survive a node reboot. The vGPU Manager validator already waits for VFs to appear (`waitForVFs`) before the vGPU device-manager applies its config, but nothing re-creates them. After a node reboot the VFs are gone, so validation blocks until they are re-established out-of-band (e.g. host systemd units running `sriov-manage -e`) — the operator does not recover this on its own.

This adds the smallest useful part of the reboot-recovery seam discussed in #2600: the operator re-enables the SR-IOV VFs itself. Right before the existing wait, `VGPUManager.validate()` runs `sriov-manage -e ALL` inside the resolved driver root (the driver container root, or the host root when the vGPU Manager driver is pre-installed on the host).

The call is:

- gated to the `vm-vgpu` workload path — the function already early-returns for other workloads;
- idempotent — skipped when every SR-IOV-capable GPU already has its full VF count, which is the normal steady state, so it is a no-op except right after a reboot when no VMs are running yet;
- best-effort — on failure it logs and falls through to `waitForVFs`, preserving the current behavior for setups that create VFs out-of-band.

The PF-counting loop is extracted into a shared `countVFs` helper so the new idempotency guard and `waitForVFs` agree on the accounting.

Scope is intentionally limited to VF re-enablement. GPU reset, MIG-mode commit, and `nvidia-persistenced` coordination (the rest of #2600) are out of scope here.

Refs #2600.

## Checklist

## Testing

- `go vet` and `golangci-lint run` on the changed package, cross-compiled `GOOS=linux` (the validator is a Linux-only binary): clean, 0 issues. No asset or `go.mod`/`go.sum` changes.
- New unit test `TestCountVFs` covers the shared VF-accounting helper across the boundary cases (no SR-IOV GPUs, VFs missing after reboot, fully enabled, partially enabled across multiple PFs, and VFs not miscounted as PFs); it passes in a Linux container.
- Not yet exercised on GPU hardware; the runtime `sriov-manage`/chroot path relies on CI and cluster testing.

### DIFF
--- a/cmd/nvidia-validator/main.go
+++ b/cmd/nvidia-validator/main.go
@@ -239,6 +239,10 @@ const (
 	shell = "sh"
 	// defaultVFWaitTimeout is the default timeout for waiting for VFs to be created
 	defaultVFWaitTimeout = 5 * time.Minute
+	// sriovManageBinaryPath is the path to NVIDIA's sriov-manage script inside the
+	// driver root. It ships with the vGPU Manager (host driver) and enables
+	// SR-IOV Virtual Functions on the NVIDIA GPUs.
+	sriovManageBinaryPath = "/usr/lib/nvidia/sriov-manage"
 	// constants for driver components
 	GDRCOPY       = "gdrcopy"
 	NVIDIAFS      = "nvidia-fs"
@@ -1747,6 +1751,14 @@ func (v *VGPUManager) validate() error {
 		return err
 	}
 
+	// SR-IOV VFs are runtime state and do not survive a node reboot, so
+	// re-establish them before waiting. This is best-effort: on failure we still
+	// fall through to waitForVFs, which preserves the prior behavior on setups
+	// where the VFs are created out-of-band.
+	if err := enableVFs(hostDriver); err != nil {
+		log.Warnf("Unable to enable SR-IOV VFs, will wait for them to appear: %v", err)
+	}
+
 	log.Info("Waiting for VFs to be available...")
 	if err := waitForVFs(ctx, defaultVFWaitTimeout); err != nil {
 		return fmt.Errorf("vGPU Manager VFs not ready: %w", err)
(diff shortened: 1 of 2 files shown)


rajathagasthya:
Thanks @lexfrei! I will defer to @cdesiniotis to assess this PR since he has more knowledge on SR-IOV parts.

karthikvetrivel:
Hey @lexfrei! Sorry for the delay on the response here. For some context, the validator infers the GPU's vGPU mode from side effects: `totalVFs > 0` and `numVFs > 0`.

`enableVFs` gates on the PCIe capability bit (IsPF() && TotalVFs > 0) rather than the driver's vGPU mode, so a Turing T4, which advertises `sriov_totalvfs=16` but runs PF-based mdev and never creates VFs, reads as 0/16 "missing" and gets sriov-manage -e ALL.  The `sriov-manage` still unbinds the PF from nvidia, binds pci-pf-stub, and rebinds on a PF that may have live mdev vGPUs. 

We can't reuse `isDriverUsingSRIOV` because it requires NumVFs > 0.  It could be worth switching to using `nvmlDeviceGetHostVgpuMode()` to get a signal from the driver directly.

karthikvetrivel:
Another high level comment: 

Every other check in this pod just reads state and reports, while this one detaches GPU from its driver and reattaches it. This could be out of scope of a "validator." Given the vGPU Manager container already calls `sriov-manage` at startup,
is the validator the right home for this, or should VF creation live with whichever component owns the driver lifecycle? 

@cdesiniotis interested to hear your thoughts on this thread.

cdesiniotis:
> Another high level comment:
> 
> Every other check in this pod just reads state and reports, while this one detaches GPU from its driver and reattaches it. This could be out of scope of a "validator." Given the vGPU Manager container already calls `sriov-manage` at startup, is the validator the right home for this, or should VF creation live with whichever component owns the driver lifecycle?
> 
> @cdesiniotis interested to hear your thoughts on this thread.

Agreed. SRIOV creation is out-of-scope for the validator. 

If the vGPU Manager is deployed by the operator, then the VFs will be created by the vGPU Manager pod on a reboot. If the vGPU Manager is installed directly on the host (not managed by the operator), then some host-side component (i.e. a systemd service, a udev rule) should be responsible for ensuring the VFs are recreated on a reboot.

Closing this PR as out-of-scope.

PR #2603, opened 2026-07-04
feat(validator): commit pending MIG-mode change via targeted GPU reset

lexfrei:
## Description

> **RFC / DRAFT**: opening for design feedback before the runtime path is exercised on GPU hardware. See the "Destructive GPU reset, why it is safe here" section below.

On Ampere (A100, A30) enabling MIG mode needs a GPU reset, and the driver tries that reset itself. It doesn't always work: the reset is refused while the GPU is in use, and under passthrough the hypervisor can disallow it. The GPU then sits with `mig.mode.pending=Enabled` and `mig.mode.current` not `Enabled`, so MIG-backed vGPU devices can't be created until something resets it later. Right now that means a hand-rolled host-side step.

This adds an opt-in step to vGPU Manager validation that commits such a pending enable with a targeted `nvidia-smi --gpu-reset`. It runs before the wait for SR-IOV VFs, and it skips any GPU that already has VFs.

What this is not: it does not recreate MIG state lost to a reboot, and it isn't needed for that. MIG mode survives a reboot on Ampere via an InfoROM status bit, and from Hopper on, enabling it no longer needs a reset. MIG instances don't survive on any generation, but recreating those is `mig-parted`'s job.

## Destructive GPU reset, why it is safe here

`nvidia-smi --gpu-reset` clears GPU hardware and software state; the manual describes it as a way to avoid a machine reboot. It does not tear down running work: nvidia-smi refuses the reset while any application is using the device, and the MIG guide says to kill those processes and retry. The step is off by default and gated because a validator should not do a privileged device-level operation implicitly, and because clearing device state breaks anything holding the GPU. It resets a GPU only when all of these hold:

- Opt-in. New flag `--commit-mig-mode-with-gpu-reset` (env `COMMIT_MIG_MODE_WITH_GPU_RESET`), default `false`. Unset, behavior is identical to today: no reset, `validate()` unchanged.
- vGPU path only. `commitMIGMode` is reached only from `VGPUManager.validate()`, which early-returns for non-`vm-vgpu` workloads.
- MIG mode requested but not committed: `mig.mode.pending == Enabled` and `mig.mode.current != Enabled`. Something else already requested the change; this only commits it, never decides it.
- No SR-IOV VFs enabled on that GPU.
- No running compute processes. The probe is `--query-compute-apps`, so it only sees compute clients. `nvidia-smi --gpu-reset` refuses to reset a GPU while any application is using it, including the graphics and monitoring clients this query misses. The manual also says an individual reset may not be feasible while compute apps run on other GPUs; that comes back as a logged failure.

## Enabling it

The flag reads its env var from the `vgpu-manager` validation init container, so the generic `validator.env` does not reach it: that one is applied to the main validator container. The key that works is `validator.vgpuManager.env`, which the chart templates and the operator forwards on `case "vgpu-manager"`:

```
--set validator.vgpuManager.env[0].name=COMMIT_MIG_MODE_WITH_GPU_RESET \
--set validator.vgpuManager.env[0].value=true
```

That path works but is not discoverable. `validator.vgpuManager` is absent from `values.yaml`, which declares only `env`, `args` and `plugin.env` under `validator`, so the key never shows up in `helm show values` and an operator can only find it by reading the chart template. `validator.vfioPCI` and `validator.vgpuDevices` have the same gap. Happy to send a separate PR declaring all three if you want them discoverable.

In the steady state, MIG already committed or VFs present, it does nothing. Best-effort: reset failures are logged and validation continues, so setups that commit MIG mode out-of-band keep working as before.

Two open questions I'd rather raise than paper over, both discussed in #2600:

- The VF guard is stated here as a precondition, but I can't source a rule that a GPU reset requires SR-IOV to be disabled first. It is not among the preconditions NVIDIA documents for `nvidia-smi -r`, and NVIDIA's own MIG recipes enable VFs before setting MIG mode. So the guard is observed behavior without a verified rationale.
- On NVSwitch / Fabric Manager systems a single-GPU reset can interact with fabric partition state. This change only commits MIG mode and doesn't manage fabric partitions.

## Checklist

- [ ] Lint checks passing (`make lint`): the changed package is clean, `GOOS=linux golangci-lint run ./cmd/nvidia-validator/` reports 0 issues. Repo-wide `make lint` was not run locally.
- [ ] Generated assets in-sync (`make validate-generated-assets`): this change adds no API/CRD changes, so no generated assets change; the full target was not run.

## Testing

- `gofmt`, `go build`, `go vet`, and `golangci-lint run` on the changed package, cross-compiled `GOOS=linux` (the validator is a Linux-only binary): clean, 0 issues. No asset or `go.mod`/`go.sum` changes.
- New unit tests run green in a Linux container (`go test ./cmd/nvidia-validator/`): `TestNormalizePCIAddress` (nvidia-smi 8-digit vs go-nvlib 4-digit PCI domains normalize to the same key), `TestParseMIGModes` (CSV parsing, `[N/A]`, malformed rows, the pending-vs-current guard), `TestMIGModeNeedsCommit` (the enable-only commit predicate across every current/pending combination, including case-insensitivity), and `TestShouldResetForMIGCommit` (the reset-decision guard matrix over uncommitted-enable, VFs-present and workload-running).
- `commitMIGMode` itself has no test, including its off-by-default gate. That is the gap I'd fix first if this design survives the discussion in #2600.
- Not yet exercised on GPU hardware; the runtime chroot / `nvidia-smi --gpu-reset` path relies on CI and cluster testing. This is why the PR is opened as a draft/RFC.

Refs #2600.

### DIFF
--- a/cmd/nvidia-validator/main.go
+++ b/cmd/nvidia-validator/main.go
@@ -18,12 +18,14 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -136,6 +138,7 @@ var (
 	hostRootFlag                  string
 	driverInstallDirFlag          string
 	driverInstallDirCtrPathFlag   string
+	commitMIGModeWithResetFlag    bool
 )
 
 // defaultGPUWorkloadConfig is "vm-passthrough" unless
@@ -239,6 +242,10 @@ const (
 	shell = "sh"
 	// defaultVFWaitTimeout is the default timeout for waiting for VFs to be created
 	defaultVFWaitTimeout = 5 * time.Minute
+	// migModeEnabled is the value nvidia-smi reports for mig.mode.current and
+	// mig.mode.pending when MIG mode is enabled (the other values being
+	// "Disabled" and "[N/A]" for GPUs that do not support MIG).
+	migModeEnabled = "Enabled"
 	// constants for driver components
 	GDRCOPY       = "gdrcopy"
 	NVIDIAFS      = "nvidia-fs"
@@ -375,6 +382,17 @@ func main() {
 			Usage:       "the path where the NVIDIA driver install dir is mounted in the container",
 			Destination: &driverInstallDirCtrPathFlag,
 			Sources:     cli.EnvVars("DRIVER_INSTALL_DIR_CTR_PATH"),
+		},
+		&cli.BoolFlag{
+			Name:  "commit-mig-mode-with-gpu-reset",
+			Value: false,
+			Usage: "on the vGPU (sandbox) workload path, commit a pending-but-uncommitted MIG-mode enable " +
+				"by performing a targeted GPU reset (nvidia-smi --gpu-reset) during vGPU Manager validation. " +
+				"This is destructive: it resets the GPU. It only acts on GPUs that have an uncommitted MIG-mode " +
+				"enable, no SR-IOV VFs enabled, and no running compute processes, so it is a no-op except right " +
+				"after a reboot before any VM is running. Disabled by default.",
+			Destination: &commitMIGModeWithResetFlag,
+			Sources:     cli.EnvVars("COMMIT_MIG_MODE_WITH_GPU_RESET"),
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


rajathagasthya:
Just linking back to the ongoing discussion at https://github.com/NVIDIA/gpu-operator/issues/2600#issuecomment-4960267693 (thanks for the additional context there!). Let's revisit this PR once we decide on a direction for #2600.

PR #2607, opened 2026-07-07
add new deploy label to manage third-party gpu client pods

tariq1890:
This PR introduces a new node label `"nvidia.com/gpu.deploy.client": "true"` to designate pods that are gpu clients (like our operands `dcgm`, `mig-manager`, `device-plugin`, `gpu-feature-discovery`) which are not first-class components of the gpu-operator.

### Motivation

Advanced users bring in their own gpu client pods that require management/admin access to the GPUs so they too need to bypass the device plugin for gpu allocation much like the gpu-operator's operands. Today, users who bring their own gpu client pods will face issues during driver upgrades as the gpu-operator is unaware of the existence of these pods which actively hold gpu device handles thus interrupting a driver upgrade. Users today will have to manually evict these pods during these gpu client pods so that the upgrade can proceed. This scales poorly and also entails heavy operational overhead as it involves human intervention.

### Solution

Users who bring their own gpu client pods can now take advantage of this feature by adding the  `"nvidia.com/gpu.deploy.client": "true"` label to the **nodeSelector** of their gpu client's resource manifest (whether Daemonset/Deployment/StatefulSet/Job). With this new label added to the nodeSelector, the gpu-operator will then be able to bounce gpu-client pods automatically during a driver upgrade without any human intervention.

Related PRs:

- https://github.com/NVIDIA/k8s-driver-manager/pull/205
- https://github.com/NVIDIA/mig-parted/pull/426

### DIFF
--- a/controllers/state_manager.go
+++ b/controllers/state_manager.go
@@ -94,6 +94,7 @@ var gpuStateLabels = map[string]map[string]string{
 		"nvidia.com/gpu.deploy.dcgm-exporter":         "true",
 		"nvidia.com/gpu.deploy.node-status-exporter":  "true",
 		"nvidia.com/gpu.deploy.operator-validator":    "true",
+		"nvidia.com/gpu.deploy.client":                "true",
 	},
 	gpuWorkloadConfigVMPassthrough: {
 		"nvidia.com/gpu.deploy.sandbox-device-plugin": "true",
@@ -101,6 +102,7 @@ var gpuStateLabels = map[string]map[string]string{
 		"nvidia.com/gpu.deploy.vfio-manager":          "true",
 		"nvidia.com/gpu.deploy.kata-manager":          "true",
 		"nvidia.com/gpu.deploy.cc-manager":            "true",
+		"nvidia.com/gpu.deploy.client":                "true",
 	},
 	gpuWorkloadConfigVMVgpu: {
 		"nvidia.com/gpu.deploy.sandbox-device-plugin": "true",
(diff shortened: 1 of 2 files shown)


controllers/state_manager.go
```diff
@@ -95,6 +95,7 @@ var gpuStateLabels = map[string]map[string]string{
 		"nvidia.com/gpu.deploy.dcgm-exporter":         "true",
 		"nvidia.com/gpu.deploy.node-status-exporter":  "true",
 		"nvidia.com/gpu.deploy.operator-validator":    "true",
+		"nvidia.com/gpu.deploy.client":                "true",
```
cdesiniotis, line 97: question -- should we be adding this label on all GPU nodes (regardless of the sandbox mode)?

tariq1890: Good question, I think we can

tariq1890: This is done.

rahulait (approved):
LGTM

tariq1890:
Thanks for the review @rahulait . Looks like I missed a change. Seeing the errors below when testing mig-manager

```
time="2026-07-10T22:52:57Z" level=error msg="Error: failed to wait for pods to be deleted: third-party gpu client pods did not shutdown: failed to list pods on node ipp1-1618: pods is forbidden: User \"system:serviceaccount:test-operator :nvidia-mig-manager\" cannot list resource \"pods\" in API group \"\" at the cluster scope"
```

tariq1890:
This PR has been updated based on review comments provided in:
https://github.com/NVIDIA/k8s-driver-manager/pull/205
https://github.com/NVIDIA/mig-parted/pull/426

rajathagasthya (approved):

PR #2610, opened 2026-07-08
trigger nodeLabelingReconciler when nvidia.com/mig.capable is changed

tariq1890:
This PR fixes a regression where the gpu-operator no longer applies the `nvidia.com/deploy.mig-manager="true"` label on the node following the apply of the `nvidia.com/mig.capable: "true"` node label by gpu-feature-discovery


controllers/nodelabeling_controller.go
cdesiniotis, line 461: Can you also update this log message to contain `migCapableLabelChanged`

tariq1890: Done, please check now

cdesiniotis (approved):

PR #2612, opened 2026-07-08
Fix NVIDIADriver owner assignment after GPU label updates

rahulait:
## Description

This change fixes a race in NVIDIADriver CRD mode where driver DaemonSets may never be rendered after a fresh install or migration.

NodeLabeling now defers dependent node-label operations after it patches GPU labels, and the node update predicate now explicitly reconciles when `nvidia.com/gpu.present` changes. This lets the follow-up reconcile run after the informer observes the GPU label update, so NVIDIADriver owner labels can be assigned reliably.

Code is drafted with the help of codex.

 

## Checklist

### To reproduce the issue
a. Had a clean node with no nvidia labels and no gpu-operator running.
b. Installed gpu-operator which had the bug
```
helm install gpu-operator -n gpu-operator --create-namespace oci://ghcr.io/nvidia/gpu-operator/charts/gpu-operator:0.0.0-git-16638954 --set driver.nvidiaDriverCRD.enabled=true --set driver.nvidiaDriverCRD.deployDefaultCR=true
```
c. Waited for driver pod to not show up. If it did, re-ran steps a-c.

I was able to reproduce this issue after running for 2-3 times.

### To test the fix:
Did the same steps as mentioned to reproduce the issue but with the oci helm chart built for this PR. Re-ran them atleast 15-20 times and couldn't reproduce it again.

### DIFF
--- a/controllers/nodelabeling_controller.go
+++ b/controllers/nodelabeling_controller.go
@@ -61,6 +61,56 @@ type nodeLabelingController struct {
 	logger        logr.Logger
 }
 
+// gpuNodeLabelsUpdateResult reports total node patches and the subset where GPU
+// discovery state changed. The discovery state is stored in nvidia.com/gpu.present,
+// which AssignOwners uses to find GPU nodes, so dependent operations are deferred
+// until the informer cache observes those node updates.
+type gpuNodeLabelsUpdateResult struct {
+	totalPatchedNodeCount             int
+	gpuDiscoveryStateChangedNodeCount int
+}
+
+// nodeLabelUpdateReasons captures why a node update event should trigger node-label reconciliation.
+type nodeLabelUpdateReasons struct {
+	gpuCommonLabelMissing        bool
+	gpuCommonLabelOutdated       bool
+	gpuCommonLabelChanged        bool
+	commonOperandsLabelChanged   bool
+	gpuWorkloadConfigChanged     bool
+	migCapableLabelChanged       bool
+	osTreeLabelChanged           bool
+	nvidiaDriverOwnerLabelChange bool
+}
+
+// needsUpdate reports whether any tracked node-label change requires reconciliation.
+func (r nodeLabelUpdateReasons) needsUpdate() bool {
+	return r.gpuCommonLabelMissing ||
+		r.gpuCommonLabelOutdated ||
+		r.gpuCommonLabelChanged ||
+		r.commonOperandsLabelChanged ||
+		r.gpuWorkloadConfigChanged ||
+		r.migCapableLabelChanged ||
+		r.osTreeLabelChanged ||
+		r.nvidiaDriverOwnerLabelChange
+}
+
+// getNodeLabelUpdateReasons compares old and new node labels for changes that affect GPU Operator labels.
+func getNodeLabelUpdateReasons(oldLabels, newLabels map[string]string) nodeLabelUpdateReasons {
+	oldGPUWorkloadConfig, _ := getWorkloadConfig(oldLabels, true)
+	newGPUWorkloadConfig, _ := getWorkloadConfig(newLabels, true)
+
+	return nodeLabelUpdateReasons{
+		gpuCommonLabelMissing:        hasGPULabels(newLabels) && !hasCommonGPULabel(newLabels),
+		gpuCommonLabelOutdated:       !hasGPULabels(newLabels) && hasCommonGPULabel(newLabels),
+		gpuCommonLabelChanged:        oldLabels[commonGPULabelKey] != newLabels[commonGPULabelKey],
+		commonOperandsLabelChanged:   hasOperandsDisabled(oldLabels) != hasOperandsDisabled(newLabels),
+		gpuWorkloadConfigChanged:     oldGPUWorkloadConfig != newGPUWorkloadConfig,
+		migCapableLabelChanged:       hasMIGCapableGPU(oldLabels) != hasMIGCapableGPU(newLabels),
+		osTreeLabelChanged:           oldLabels[nfdOSTreeVersionLabelKey] != newLabels[nfdOSTreeVersionLabelKey],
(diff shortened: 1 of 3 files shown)


cdesiniotis (approved):
lgtm. let's try to get another pair of eyes on this

karthikvetrivel:
Thanks, Rahul! Did you do any e2e testing manually or only unit tests?

rahulait:
> Thanks, Rahul! Did you do any e2e testing manually or only unit tests?

Yeah, I did test the change locally on my single node setup. I specifically did:
**To reproduce the issue**
a. Had a clean node with no nvidia labels and no gpu-operator running.
b. Installed gpu-operator which had the bug
```
helm install gpu-operator -n gpu-operator --create-namespace oci://ghcr.io/nvidia/gpu-operator/charts/gpu-operator:0.0.0-git-16638954 --set driver.nvidiaDriverCRD.enabled=true --set driver.nvidiaDriverCRD.deployDefaultCR=true
```
c. Waited for driver pod to not show up. If it did, re-ran steps a-c.

I was able to reproduce this issue after running for 2-3 times.

**To test the fix:**
Did the same steps as mentioned to reproduce the issue but with the oci helm chart built for this PR. Re-ran them atleast 15-20 times and couldn't reproduce it again.

karthikvetrivel (approved):
LGTM!

controllers/nodelabeling_controller.go
```diff
@@ -90,9 +140,23 @@ func (r *NodeLabelingReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		if gpuLabelUpdateResult.totalPatchedNodeCount > 0 {
+			r.Log.Error(err, "GPU node label update failed after partially updating nodes",
+				"totalPatchedNodeCount", gpuLabelUpdateResult.totalPatchedNodeCount,
+				"gpuPresentPatchedNodeCount", gpuLabelUpdateResult.gpuPresentPatchedNodeCount,
+			)
+		}
 		return reconcile.Result{}, err
 	}
+	if gpuLabelUpdateResult.gpuPresentPatchedNodeCount > 0 {
+		r.Log.Info("GPU present labels used by owner assignment updated; dependent node label operations will run after the node update event",
```
tariq1890: The gpu-operator controller logs are already quite chatty. So I am questioning if we need to add this as an Info log. Can we make it debug instead?

controllers/nodelabeling_controller.go
```diff
@@ -61,6 +61,56 @@ type nodeLabelingController struct {
 	logger        logr.Logger
 }
 
+// gpuNodeLabelUpdateResult reports total node patches and the subset that changed
+// nvidia.com/gpu.present, the label AssignOwners uses to find GPU nodes. When
+// this label changes, owner assignment is deferred until the informer cache
+// observes the node update.
+type gpuNodeLabelUpdateResult struct {
```
tariq1890: I see that `totalPatchedNodeCount` no longer factors into the decision making of the deferred reconciliation. Do we need it?

controllers/nodelabeling_controller.go
```diff
@@ -61,6 +61,56 @@ type nodeLabelingController struct {
...
+	totalPatchedNodeCount      int
+	gpuPresentPatchedNodeCount int
```
tariq1890: The `gpuNodeLabelUpdateResult` already does some heavylifting in terms of the naming
```suggestion
	stateLabelsModified int
```

controllers/nodelabeling_controller.go
tariq1890:
```suggestion
type gpuNodeLabelsUpdateResult struct {
```

rahulait: This is for logging purposes only in case we need to troubleshoot things.

controllers/nodelabeling_controller.go
rahulait: Yeah, makes sense. Let me make it debug then.

rahulait: updated.

controllers/nodelabeling_controller.go
rahulait: Updated.

controllers/nodelabeling_controller.go
rahulait: This field is counting for node patches for `nvidia.com/gpu.present` label. I would prefer to keep the current name as its more clear.

tariq1890: `gpuPresentPatchedCount` is confusing as it suggests the count of `nvidia.com/gpu.present` labels that are changed whereas it counts the number of updates to either `gpu.present` or any of the stateLabels. At least, let's use a prefix other than `gpuPresent`.

controllers/nodelabeling_controller.go
tariq1890: thanks

controllers/nodelabeling_controller.go
tariq1890: I am comparing these two this way:

i) `gpuNodeLabelUpdateResult.gpuPresentPatchedNodeCount`
vs
ii) `gpuNodeLabelUpdateResult.stateLabelsModified`

rahulait: Correct me if I am wrong, but we are updating `totalPatchedNodeCount` in case of `gpu.present` or stateLabels change. `gpuPresentPatchedNodeCount` is updated only if `gpu.present` label is modified. So its a subset of the larger count.
```
		modified := gpuPresentLabelModified || stateLabelsModified
		if modified {
			if err := nlc.client.Patch(ctx, &node, client.MergeFrom(original)); err != nil {
				return result, fmt.Errorf("unable to label node %s: %w", node.Name, err)
			}
			result.totalPatchedNodeCount++
			if gpuPresentLabelModified {
				result.gpuPresentPatchedNodeCount++
			}
		}
```

tariq1890 (approved):
Thanks very much for fixing this, @rahulait !

PR #2642, opened 2026-07-15
[helm] validate devicePlugin custom configMap settings

### LINKED ISSUE #2641: [Bug]: devicePlugin config volume is added even when create=true and config.data is empty, so no ConfigMap is created
_**Important Note:  NVIDIA AI Enterprise customers can get support from NVIDIA Enterprise support. Please open a case [here](https://enterprise-support.nvidia.com/s/create-case)**._

**Describe the bug**
For the device plugin, the chart/controller behavior is inconsistent when a user asks the chart to create the config ConfigMap, but provides empty config data.

If I set:

```yaml
devicePlugin:
  config:
    name: plugin-config
    create: true
    data: {}
    default: default
```

then the chart does **not** create the ConfigMap, because `plugin_config.yaml` only renders when `create` is true **and** `data` is non-empty.

However, `clusterpolicy.yaml` still sets `spec.devicePlugin.config.name`, and the controller later treats that as sufficient to configure the device-plugin DaemonSet with:
- `CONFIG_FILE=/config/config.yaml`
- a ConfigMap-backed volume for `plugin-config`
- volume mounts for `/available-configs` and `/config`

This means a user can explicitly request ConfigMap creation with `create: true`, but because `data` is empty, no ConfigMap is created, while the DaemonSet still references it.

That can leave the device-plugin pod referencing a missing ConfigMap and failing to start.

This is specifically confusing because from the user's perspective:
- they asked the chart to create the ConfigMap
- the chart silently does not create it because `data` is empty
- the operator still injects the volume mount because `config.name` is set

**To Reproduce**
1. Install the chart from `deployments/gpu-operator`.
2. Use values like:

```yaml
devicePlugin:
  enabled: true
  config:
    name: plugin-config
    create: true
    data: {}
    default: default
```

3. Install or upgrade the release.
4. Observe that:
   - `deployments/gpu-operator/templates/plugin_config.yaml` does not render a ConfigMap
   - `deployments/gpu-operator/templates/clusterpolicy.yaml` still sets `spec.devicePlugin.config.name`
   - the controller adds ConfigMap volume mounts to the device-plugin DaemonSet based on `config.name`

**Expected behavior**
If `devicePlugin.config.create: true` is set, but `devicePlugin.config.data` is empty, one of the following should happen consistently:

1. The chart/operator should reject the configuration as invalid, because the user asked to create a ConfigMap but provided no data.
2. Or the chart should avoid setting `spec.devicePlugin.config.name` in this case, so the DaemonSet does not reference a ConfigMap that was not created.
3. At minimum, this should be validated/documented clearly so the resulting behavior is not surprising.

**Relevant code**
Chart only creates the ConfigMap when `create == true` and `data` is non-empty:
- `deployments/gpu-operator/templates/plugin_config.yaml`

`ClusterPolicy` still gets config name when `devicePlugin.config.name` is set:
- `deployments/gpu-operator/templates/clusterpolicy.yaml`

Controller injects config mounts whenever `spec.devicePlugin.config.name` is non-empty:
- `controllers/object_controls.go`
- `isCustomPluginConfigSet()`
- `handleDevicePluginConfig()`

 - GPU Operator Version: main branch behavior
 - OS: N/A
 - Kernel Version: N/A
 - Container Runtime Version: N/A
 - Kubernetes Distro and Version: N/A

**Information to attach** (optional if deemed irrelevant)

 - [ ] kubernetes pods status: `kubectl get pods -n OPERATOR_NAMESPACE`
 - [ ] kubernetes daemonset status: `kubectl get ds -n OPERATOR_NAMESPACE`
 - [ ] If a pod/ds is in an error state or pending state `kubectl describe pod -n OPERATOR_NAMESPACE POD_NAME`
 - [ ] If a pod/ds is in an error state or pending state `kubectl logs -n OPERATOR_NAMESPACE POD_NAME --all-containers`
 - [ ] Output from running `nvidia-smi` from the driver container: `kubectl exec DRIVER_POD_NAME -n OPERATOR_NAMESPACE -c nvidia-driver-ctr -- nvidia-smi`
 - [ ] containerd logs `journalctl -u containerd > containerd.log`

tariq1890:
This fixes a bug where the rendered device-plugin daemonset ends up in a bad state when the user enables custom configmap creation but provides empty configmap data as input.

Closes #2641

### DIFF
--- a/deployments/gpu-operator/templates/validations.yaml
+++ b/deployments/gpu-operator/templates/validations.yaml
@@ -5,3 +5,11 @@
 {{- if and (eq .Values.cdi.nriPluginEnabled true) (eq .Values.toolkit.enabled false)  }}
 {{ fail "the NRI Plugin cannot be enabled when the Container Toolkit is disabled" }}
 {{- end }}
+
+{{- if and (.Values.devicePlugin.config.create) (empty .Values.devicePlugin.config.data) }}
+{{ fail "devicePlugin.config.data cannot be empty when devicePlugin.config.create is set to true" }}
+{{- end }}
+
+{{- if and (.Values.devicePlugin.config.create) (empty .Values.devicePlugin.config.name) }}
+{{ fail "devicePlugin.config.name cannot be empty when devicePlugin.config.create is set to true" }}
+{{- end }}


rajathagasthya (approved):

deployments/gpu-operator/templates/validations.yaml
```diff
@@ -5,3 +5,7 @@
 {{- if and (eq .Values.cdi.nriPluginEnabled true) (eq .Values.toolkit.enabled false)  }}
 {{ fail "the NRI Plugin cannot be enabled when the Container Toolkit is disabled" }}
 {{- end }}
+
+{{- if and (.Values.devicePlugin.config.create) (empty .Values.devicePlugin.config.data) }}
+{{ fail "devicePlugin.config.data cannot be empty when device.Plugin.config.create is set to true" }}
```
rahulait:
```suggestion
{{ fail "devicePlugin.config.data cannot be empty when devicePlugin.config.create is set to true" }}
```

deployments/gpu-operator/templates/validations.yaml
```diff
@@ -5,3 +5,7 @@
 {{- if and (eq .Values.cdi.nriPluginEnabled true) (eq .Values.toolkit.enabled false)  }}
 {{ fail "the NRI Plugin cannot be enabled when the Container Toolkit is disabled" }}
 {{- end }}
+
```
rahulait, line 8: We should add this check as well:

```
{{- if and (.Values.devicePlugin.config.create) (empty .Values.devicePlugin.config.name) }}
{{ fail "devicePlugin.config.name cannot be empty when devicePlugin.config.create is set to true" }}
{{- end }}
```

deployments/gpu-operator/templates/validations.yaml
tariq1890: great catch, thanks!

tariq1890:
@rahulait Thanks for the review! Please check now

rahulait (approved):
LGTM

PR #2656, opened 2026-07-21
Add unit tests for internal packages

abrarshivani:
## Description

Adds unit tests to raise coverage across the `internal/` packages. This PR is **tests-only** — it contains no production changes. Each package was driven to full statement coverage of every reachable path it owns (both success and error paths), verified with `go test -covermode=count`.

The branch is organized as **one commit per package**, so each can be reviewed independently.

### Packages covered

| Commit | Package | Coverage |
| --- | --- | --- |
| `Add unit tests for internal/info` | `internal/info` | **100%** |
| `Add unit tests for internal/conditions` | `internal/conditions` | `clusterpolicy.go` **100%**, `nvidiadriver.go` **100%** |
| `Add unit tests for internal/nodeinfo` | `internal/nodeinfo` | **100%** |
| `Add unit tests for internal/nvidiadriver ...` | `internal/nvidiadriver` | **98.6%** |
| `Add unit tests for internal/utils` | `internal/utils` | `GetFilesWithSuffix` + object-hash helpers **100%** |
| `Add unit tests for internal/validator ...` | `internal/validator` | **100%** |

- **`internal/info`** — `GetVersionParts` / `GetVersionString`: with/without `gitCommit`, empty/whitespace values, variadic joining, fresh-slice / no-mutation guarantees.
- **`internal/conditions`** — ClusterPolicy and NVIDIADriver updaters: Ready/Error success, type-assertion guard, failed-Get, unknown-status-type default branch, retry-on-conflict, non-conflict update-error propagation, and `status.state` defaulting/preservation. Uses the controller-runtime fake client + interceptors; condition assertions use `cmp.Diff` (ignoring server-set `LastTransitionTime`/`ObservedGeneration`); schemes are built per-test rather than mutating the global client-go scheme. This promotes `github.com/google/go-cmp` to a direct dependency (a test dependency) and vendors `cmp/cmpopts`.
- **`internal/nodeinfo`** — the previously untested `Reset` methods on both filter builders.
- **`internal/nvidiadriver`** — `AssignOwners` error paths (driver-list, node-list, owner-label patch add/remove). The one uncovered line is a defensive nil-`Labels` guard, unreachable because nodes are listed by the GPU-present label and therefore always carry labels.
- **`internal/utils`** — `GetFilesWithSuffix` and the object-hash helpers.
- **`internal/validator`** — `Validate` error paths (driver-list failure, invalid listed selector, node-list failure).

### Out of scope

Two packages touched here also gained new, untested code on `main` after this branch was opened. Covering it is left to follow-up PRs, so the package-level numbers are below 100%:

- `internal/conditions` is **64.6%** overall — `gpucluster.go` (added by #2571) is untested.
- `internal/utils` is **67.2%** overall — `PrependPathListEnvvar`, `SetEnvVar`, and `WriteFileAtomically` are untested. `GetStringHash` is at 75%: its `panic` on a `hasher.Write` error is a deliberate assertion of an invariant and is unreachable by design, so it is left as-is.

The `internal/image` tests that previously lived in this PR have been **dropped**: the `ImagePath` work was split into #2682, which has since merged and already brings `internal/image` to 100%.

### Production changes split out

The small `internal/utils` production cleanups that previously lived here (the `GetFilesWithSuffix` dedupe `break` and the `GetObjectHashIgnoreEmptyKeys` precondition doc) have been **moved to their own PR, #2690**, so this PR stays tests-only. `utils.go` here is unchanged from `main`.

## Checklist

## Testing

Every commit was verified to build and pass independently (`git rebase --exec`), so the branch is bisect-clean:

```
go build ./internal/... ./api/... ./controllers/...
go test ./internal/...
```

Full-branch checks:

```
GOOS=linux go build ./...                    # pathrs-lite is Linux-only
go test ./internal/... ./api/... ./controllers/...
go mod tidy && go mod vendor                 # no diff
GOOS=linux golangci-lint run ./internal/...  # 0 issues
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/go-logr/logr v1.4.4
 	github.com/go-logr/zapr v1.3.0
+	github.com/google/go-cmp v0.7.0
 	github.com/onsi/ginkgo/v2 v2.32.0
 	github.com/onsi/gomega v1.42.1
 	github.com/openshift/api v0.0.0-20260727141720-967cc4c36c9b
@@ -72,7 +73,6 @@ require (
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
 	github.com/google/btree v1.1.3 // indirect
 	github.com/google/gnostic-models v0.7.1 // indirect
-	github.com/google/go-cmp v0.7.0 // indirect
 	github.com/google/pprof v0.0.0-20260402051712-545e8a4df936 // indirect
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/huandu/xstrings v1.5.0 // indirect
--- a/internal/conditions/clusterpolicy_test.go
+++ b/internal/conditions/clusterpolicy_test.go
@@ -1,0 +1,232 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package conditions
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	nvidiav1 "github.com/NVIDIA/gpu-operator/api/nvidia/v1"
+)
+
+func clusterPolicyScheme(t *testing.T) *runtime.Scheme {
+	t.Helper()
+	s := runtime.NewScheme()
+	require.NoError(t, nvidiav1.AddToScheme(s))
+	return s
+}
+
+func newClusterPolicyClient(t *testing.T, objs ...client.Object) client.Client {
+	t.Helper()
+	b := fake.NewClientBuilder().WithScheme(clusterPolicyScheme(t))
+	if len(objs) > 0 {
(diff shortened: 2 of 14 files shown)


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -2061,6 +2061,12 @@ func imagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		// A partial CR spec (repository or image missing) can only produce a
+		// malformed reference such as "/image:tag" or "repo/:tag", so fail fast
+		// with a clear error instead of emitting an invalid image path.
+		if repository == "" || image == "" {
```
tariq1890: This is an update in the validation logic. It should be its own PR so it can be tracked as a update to the validation to the gpu-operator. The test cases covering this change should also be a part of that PR

abrarshivani: Updated. Moved the changes to separate PR here https://github.com/NVIDIA/gpu-operator/pull/2682

tariq1890: Thanks!

internal/utils/utils.go
```diff
@@ -121,60 +124,7 @@ func hashNonZeroFields(h hash.Hash32, v reflect.Value) {
 
 func GetStringHash(s string) string {
 	hasher := fnv.New32a()
-	if _, err := hasher.Write([]byte(s)); err != nil {
-		panic(err)
-	}
+	// hash.Hash's Write never returns an error, per its documented contract.
+	_, _ = hasher.Write([]byte(s))
```
tariq1890: I would revert this change. We know that the `hasher.Write` will never return an error which is why we `panic` should `err != nil` ever evaluate to true. One of the purposes of `panic` is to throw an error when a fundamental assumption is broken or when we operate under an inconceivable scenario.

tariq1890:
Can you condense the commit history and split it by the package instead? For eg:- If you have two or more commits writing tests for code in the same package, those commits can be merged to one.

internal/utils/utils.go
abrarshivani: Makes sense. Will update this.

abrarshivani:
@tariq1890 Sure, will segregate each package tests into its own PR.

tariq1890:
Thanks @abrarshivani. To clarify, it doesn't have to be separate PRs. I just meant that we reorganize this into commits per package. If you want to make it a separate PR for each commit, that's fine too. I'll leave that for you to decide.

abrarshivani:
Oh, ok. I was thinking about this from a reviewer's perspective. I can also group the commits by package to make it easier to review.

internal/utils/utils.go
```diff
@@ -53,6 +53,9 @@ func GetFilesWithSuffix(baseDir string, suffixes ...string) ([]string, error) {
 		for _, s := range suffixes {
 			if strings.HasSuffix(base, s) {
 				files = append(files, path)
+				// Stop at the first matching suffix so a file that matches
+				// more than one of the provided suffixes is returned only once.
+				break
```
tariq1890: Are we sure that this is correct? Won't this lead to a bug as we need to collect multiple yaml files under a `state-` directory for template rendering?

tariq1890: @rahulait Can you check this one too?

tariq1890: Also, even if this change is valid, it warrants its own PR IMO. We shouldn't slip in any behavioural change in a PR that claims addition of unit tests.

rahulait: Agreed, even though it looks safe, this behavioral change should be in a separate PR

abrarshivani: Let me move the production code-related changes to another PR.

tariq1890 (approved):

PR #2665, opened 2026-07-23
fix clusterpolicy fluctuation when nvidiadriver upgrade is happening

rahulait:
## Description

Fixes ClusterPolicy status fluctuation during NVIDIADriver rolling upgrades.

When an NVIDIADriver upgrade is in progress (nodes transitioning through cordon/drain/restart/validation states), ClusterPolicy now explicitly reports NotReady with a clear reason, instead of oscillating between ready/not-ready as individual node states change.

The approach:
* After reconciling all operand states, check if any NVIDIADriver-owned node has an incomplete upgrade state (pending, in-progress, or failed).
* If so, hold overallStatus = NotReady and append descriptive reasons to the condition message.
* Refactor the not-ready message into a structured "; "-joined format that combines operand states and upgrade reasons.
## Checklist

## Testing

Tested this change on a 3 node cluster. 

Status when the upgrade starts:
```
== Node owner and upgrade labels ==
NODE            GPU   OWNER    UPGRADE_STATE         CLUSTERPOLICY  NVIDIADRIVER
a1u1g-mil-0369  true  default  upgrade-required      notReady       notReady
ipp2-1815       true  default  pod-restart-required  notReady       notReady
ipp2-2152       true  default  upgrade-required      notReady       notReady

k get clusterpolicy cluster-policy -o yaml | yq .status
conditions:
  - lastTransitionTime: "2026-07-24T18:05:33Z"
    message: ""
    reason: Error
    status: "False"
    type: Ready
  - lastTransitionTime: "2026-07-24T18:05:33Z"
    message: 'ClusterPolicy is not ready; states not ready: [state-operator-validation state-dcgm-exporter]; NVIDIADriver upgrade has not completed; one or more NVIDIADriver-owned Nodes are marked pending, in-progress, or failed'
    reason: OperandNotReady
    status: "True"
    type: Error
namespace: gpu-operator
state: notReady
```
Status when all operands are ready but driver upgrade is still in progress:
```
== Node owner and upgrade labels ==
NODE            GPU   OWNER    UPGRADE_STATE         CLUSTERPOLICY  NVIDIADRIVER
a1u1g-mil-0369  true  default  upgrade-required      notReady       notReady
ipp2-1815       true  default  upgrade-done          notReady       notReady
ipp2-2152       true  default  upgrade-required      notReady       notReady

k get clusterpolicy cluster-policy -o yaml | yq .status
conditions:
  - lastTransitionTime: "2026-07-24T18:05:33Z"
    message: ""
    reason: Error
    status: "False"
    type: Ready
  - lastTransitionTime: "2026-07-24T18:05:33Z"
    message: ClusterPolicy is not ready; NVIDIADriver upgrade has not completed; one or more NVIDIADriver-owned Nodes are marked pending, in-progress, or failed
    reason: OperandNotReady
    status: "True"
    type: Error
namespace: gpu-operator
state: notReady
```
Status when the upgrade finishes:
```
== Node owner and upgrade labels ==
NODE            GPU   OWNER    UPGRADE_STATE  CLUSTERPOLICY  NVIDIADRIVER
a1u1g-mil-0369  true  default  upgrade-done   ready          ready
ipp2-1815       true  default  upgrade-done   ready          ready
ipp2-2152       true  default  upgrade-done   ready          ready

k get clusterpolicy cluster-policy -o yaml | yq .status
conditions:
  - lastTransitionTime: "2026-07-24T18:20:17Z"
    message: ClusterPolicy is ready as all resources have been successfully reconciled
    reason: Reconciled
    status: "True"
    type: Ready
  - lastTransitionTime: "2026-07-24T18:20:17Z"
    message: ""
    reason: Ready
    status: "False"
    type: Error
namespace: gpu-operator
state: ready
```

### DIFF
--- a/controllers/clusterpolicy_controller.go
+++ b/controllers/clusterpolicy_controller.go
@@ -19,7 +19,9 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/NVIDIA/k8s-operator-libs/pkg/upgrade"
 	"github.com/go-logr/logr"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -146,6 +148,7 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	clusterPolicyCtrl.operatorMetrics.reconciliationTotal.Inc()
 	overallStatus := gpuv1.Ready
 	statesNotReady := []string{}
+	notReadyReasons := []string{}
 	for {
 		status, statusError := clusterPolicyCtrl.step()
 		if statusError != nil {
@@ -171,12 +174,31 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	if clusterPolicyCtrl.singleton.Spec.Driver.UseNvidiaDriverCRDType() {
+		upgradeInProgress, err := r.nvidiaDriverUpgradeIncomplete(ctx)
+		if err != nil {
+			clusterPolicyCtrl.operatorMetrics.reconciliationStatus.Set(reconciliationStatusNotReady)
+			clusterPolicyCtrl.operatorMetrics.reconciliationFailed.Inc()
+			if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, fmt.Sprintf("Failed to determine NVIDIADriver upgrade state: %s", err)); condErr != nil {
+				r.Log.Error(condErr, "failed to set condition")
+			}
+			return ctrl.Result{}, err
+		}
+		if upgradeInProgress {
+			overallStatus = gpuv1.NotReady
+			notReadyReasons = append(notReadyReasons,
+				"NVIDIADriver upgrade has not completed",
+				"one or more NVIDIADriver-owned Nodes are marked pending, in-progress, or failed",
+			)
+		}
+	}
+
 	// if any state is not ready, requeue for reconcile after 5 seconds
 	if overallStatus != gpuv1.Ready {
 		clusterPolicyCtrl.operatorMetrics.reconciliationStatus.Set(reconciliationStatusNotReady)
 		clusterPolicyCtrl.operatorMetrics.reconciliationFailed.Inc()
 
-		err := fmt.Errorf("ClusterPolicy is not ready, states not ready: %v", statesNotReady)
+		err := fmt.Errorf("%s", clusterPolicyNotReadyMessage(statesNotReady, notReadyReasons))
 		r.Log.Error(err, "ClusterPolicy not yet ready")
 		updateCRState(ctx, r, req.NamespacedName, gpuv1.NotReady)
 		if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.OperandNotReady, err.Error()); condErr != nil {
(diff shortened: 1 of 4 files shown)


controllers/clusterpolicy_controller.go
```diff
@@ -171,12 +174,31 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			clusterPolicyCtrl.operatorMetrics.reconciliationStatus.Set(reconciliationStatusNotReady)
+			clusterPolicyCtrl.operatorMetrics.reconciliationFailed.Inc()
+			if condErr := r.conditionUpdater.SetConditionsError(ctx, instance, conditions.ReconcileFailed, fmt.Sprintf("Failed to determine NVIDIADriver upgrade state: %s", err)); condErr != nil {
+				r.Log.Error(condErr, "failed to set condition")
+			}
+			return ctrl.Result{}, err
+		}
+		if upgradeInProgress {
+			overallStatus = gpuv1.NotReady
+			notReadyReasons = append(notReadyReasons,
```
tariq1890, line 189: I am not sure I follow why `notReadyReasons` needs to be a string slice.

controllers/clusterpolicy_controller.go
```diff
@@ -227,6 +249,52 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
+// clusterPolicyNotReadyMessage formats a condition message with operand states and additional not-ready reasons.
+func clusterPolicyNotReadyMessage(statesNotReady, notReadyReasons []string) string {
+	messageParts := []string{"ClusterPolicy is not ready"}
```
tariq1890, line 259: Can we use stringBuilder in this method?

controllers/clusterpolicy_controller.go
```diff
@@ -227,6 +249,52 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		}
+	}
+
+	return false, nil
+}
+
+// isIncompleteDriverUpgradeState reports whether a Node upgrade state keeps the aggregate driver rollout incomplete.
+func isIncompleteDriverUpgradeState(state string) bool {
+	switch state {
+	case upgrade.UpgradeStateUpgradeRequired,
```
rajathagasthya: Can we invert this condition and only check for `!(state == done || state == unknown)`? That way we don't have to list all the states here.

controllers/clusterpolicy_controller.go
rahulait: It doesn't need to be a string slice. The final message is composed of two things, statesNotReady and notReadyReasons
```
clusterPolicyNotReadyMessage(statesNotReady, notReadyReasons)
```
statesNotReady is rendered as `states not ready: [...]`
notReadyReasons is going to be rendered as `reason A; reason B; ...`

Right now, we are adding reason for nvidiadriver. If we need to add more reasons in future, it can be added to same slice. Keeping it to just a string for now is also fine. Let me know your thoughts and I can update accordingly.

controllers/clusterpolicy_controller.go
rahulait: I think its cleaner this way if we are using slices and strings.Join(). With stringBuilder, we'll have to add delimiter handling which we are avoiding here using slices and strings.Join().

controllers/clusterpolicy_controller.go
```diff
@@ -171,12 +174,31 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	if clusterPolicyCtrl.singleton.Spec.Driver.UseNvidiaDriverCRDType() {
+		upgradeInProgress, err := r.nvidiaDriverUpgradeIncomplete(ctx)
```
rajathagasthya: Claude is flagging this issue:

> A node labeled `nvidia.com/gpu-driver-upgrade.skip=true` stays in `upgrade-required` permanently. The PR counts `upgrade-required` as incomplete, so one skipped node means ClusterPolicy reports `notReady` forever and reconciles every 5 seconds forever, with all operands healthy. The check needs to exclude skipped nodes.

This sounds like a possibility to me. Is there something we can do?

controllers/clusterpolicy_controller.go
```diff
@@ -227,6 +249,52 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	messageParts := []string{"ClusterPolicy is not ready"}
+	if len(statesNotReady) > 0 {
+		messageParts = append(messageParts, fmt.Sprintf("states not ready: %v", statesNotReady))
+	}
+	messageParts = append(messageParts, notReadyReasons...)
+	return strings.Join(messageParts, "; ")
+}
+
+// nvidiaDriverUpgradeIncomplete reports whether any NVIDIADriver-owned Node has a pending, active, or failed upgrade.
+func (r *ClusterPolicyReconciler) nvidiaDriverUpgradeIncomplete(ctx context.Context) (bool, error) {
```
cdesiniotis, line 268: Question -- if an NVIDIADriver-owned daemonset is undergoing an upgrade (or at least is incomplete), is this reflected in the NVIDIADriver CR `status.state` (e.g. it is marked NotReady until the upgrade is fully rolled out)? If yes, what are your thoughts on checking the NVIDIADriver CR(s) `status.state` instead of iterating over all nodes and checking the driver upgrade label? Just an alternative solution to consider.

rahulait: I was trying to avoid adding dependency in clusterpolicy to fetch nvidiadriver CRs directly. Also, nvidiadrivers can go into notReady state if there is nodeSelector conflict with other nvidiadrivers, we don't want the clusterpolicy status to change to notReady if nvidiaDriver is marked as notReady. Since we are talking about fluctuation during upgrades, checking node upgrade labels felt more precise than relying on nvidiadriver status field.

rahulait: It might look like making clusterpolicy notReady whenever nvidiaDriver is notReady would be the right approach, however, that also brings in the additional challenge to have clusterpolicy keep a watch on nvidiadriver CRs so that it can reconcile and update status, or have nvidiadriver trigger reconcile for clusterpolicy whenever status changes. With my approach, I am trying to avoid all that and just stop the clusterpolicy fluctuation during driver upgrades.

cdesiniotis: Makes sense! Let's stick with the current implementation.

cdesiniotis (approved):

controllers/clusterpolicy_controller.go
tariq1890: Understood. It can remain as a string slice. Resolving...

controllers/clusterpolicy_controller.go
```diff
@@ -171,12 +174,31 @@ func (r *ClusterPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	if clusterPolicyCtrl.singleton.Spec.Driver.UseNvidiaDriverCRDType() {
```
rajathagasthya, line 177: My one concern is that this extends the 5s requeue until all nodes are done upgrading. Previously we would incorrectly set status to `Ready` in the middle of the upgrade (the problem being fixed here), but we weren't requeuing so the next reconcile would only happen when a watched object was updated.

We're doing the right thing here, but in each 5s reconcile we render all the templates, which is unnecessary CPU usage. If possible, we should requeue at a longer interval if we're only waiting for incomplete nodes, but I'm not sure it's straightforward.

controllers/clusterpolicy_controller.go
rahulait: Good point, updated

controllers/clusterpolicy_controller.go
rahulait: Yeah, makes sense. Updated.

controllers/clusterpolicy_controller.go
rahulait: Changed logic to rely on upgrade node label changes for reconciling clusterpolicy when driver is still upgrading. With this approach, we'll not continue retrying every 5 secs.

tariq1890 (approved):

rajathagasthya (approved):
LGTM.

PR #2666, opened 2026-07-24
[DRA Driver] sync operator clusterrole with that of state-dra-driver

tariq1890:
This bug should address the following error thrown when deploying a GPUCluster resource.

```
  status:                                                                                                                                                                                                                                        
    conditions:                                                                                                                                                                                                                                  
    - lastTransitionTime: "2026-07-24T00:18:36Z"                                                                                                                                                                                                 
      message: ""                                                                                                                                                                                                                                
      reason: Error                                                                                                                                                                                                                              
      status: "False"                                                                                                                                                                                                                            
      type: Ready                                                                                                                                                                                                                                
    - lastTransitionTime: "2026-07-24T00:18:36Z"                                                                                                                                                                                                 
      message: |-                                                                                                                                                                                                                                
        Error syncing state state-dra-driver: failed to create/update objects: clusterroles.rbac.authorization.k8s.io "compute-domain-daemon" is forbidden: user "system:serviceaccount:gpu-operator:gpu-operator" (groups=["system:serviceacco  
  nts" "system:serviceaccounts:gpu-operator" "system:authenticated"[]) is attempting to grant RBAC permissions not currently held:                                                                                                               
        {APIGroups:["resource.nvidia.com"], Resources:["computedomains"], Verbs:["patch"]}
```

This change aligns the ClusterRole used by the gpu-operator controller with the ClusterRole directly provisioned through the `state-dra-driver` templates.

### DIFF
--- a/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+++ b/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
@@ -876,14 +876,6 @@ spec:
           - resource.nvidia.com
           resources:
           - computedomains
-          verbs:
-          - get
-          - list
-          - watch
-          - update
-        - apiGroups:
-          - resource.nvidia.com
-          resources:
           - computedomains/status
           verbs:
           - get
--- a/deployments/gpu-operator/templates/clusterrole.yaml
+++ b/deployments/gpu-operator/templates/clusterrole.yaml
@@ -171,14 +171,6 @@ rules:
   - resource.nvidia.com
   resources:
   - computedomains
-  verbs:
-  - get
-  - list
-  - watch
-  - update
-- apiGroups:
-  - resource.nvidia.com
-  resources:
   - computedomains/status
   verbs:
   - get


internal/state/testdata/golden/gpucluster-dra-driver-full-spec.yaml
cdesiniotis, line 1: I see you are only updating a test file. However, the PR descriptions indicates you meant to update the main clusterrole object?

tariq1890: That was done by mistake. I've reverted that.

I've now updated the PR diff, the commit message and PR title to reflect my intent

cdesiniotis: Thanks @tariq1890. Looks good now. 

As an aside, is our ClusterRole aligned with the upstream helm chart? https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/main/deployments/helm/dra-driver-nvidia-gpu/templates/rbac-controller.yaml

It looks the upstream chart only specifies the `update` verb for the `computedomains/status` resource.

tariq1890: Good point. I agree that we should keep it aligned with the standalone helm chart. I will make that change

cdesiniotis: Sorry, I was looking at the wrong ClusterRole object. It is this one https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/main/deployments/helm/dra-driver-nvidia-gpu/templates/rbac-compute-domain-daemon.yaml

cdesiniotis: It looks like our original ClusterRole object for the `compute-domain-daemon` was correct. I think we need to update the GPU Operator ClusterRole so that it also has `patch` permissions for `computedomain`objects: https://github.com/NVIDIA/gpu-operator/blob/2a07dc4d7d33bb890a4084007cf7374ca7d943c5/deployments/gpu-operator/templates/clusterrole.yaml#L173

tariq1890: Got it. So the controller gpu-operator clusterrole should be updated as the `state-dra-driver` is already consistent with the RBAC in the standalone DRA driver helm chart.

tariq1890: The combination of `computedomains` and `computedomains/status` can definitely be called into question, but I guess that's going to make having consistent RBAC manifests across projects harder

tariq1890: Thanks @cdesiniotis ! I've updated both the helm chart and OLM. Please check now

rahulait (approved):
LGTM. I see Karthik has similar change in https://github.com/NVIDIA/gpu-operator/pull/2572/changes#diff-d1b5d7fab4bc7b5048008a78a2ad1956380805f1850ae208ef6a6dc6da107522R172-R183, but we can review them once that PR is ready for review after rebasing.

PR #2673, opened 2026-07-24
add finalizer directly to the GPUCluster manifest

tariq1890:
This PR adds the finaliser inline to the GPUCluster resource manifests. This is done for readability reasons and it also shaves off an kube client UPDATE call

### DIFF
--- a/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
+++ b/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
@@ -158,7 +158,10 @@ metadata:
           "apiVersion": "nvidia.com/v1alpha1",
           "kind": "GPUCluster",
           "metadata": {
-            "name": "gpu-cluster"
+            "name": "gpu-cluster",
+            "finalizers": [
+              "gpucluster.nvidia.com/dra-resourceclaim"
+            ]
           },
           "spec": {
             "draDriver": {
--- a/deployments/gpu-operator/templates/gpucluster.yaml
+++ b/deployments/gpu-operator/templates/gpucluster.yaml
@@ -10,6 +10,8 @@ metadata:
   # keep helm from also deleting it.
   annotations:
     "helm.sh/resource-policy": keep
+  finalizers:
+  - gpucluster.nvidia.com/dra-resourceclaim
 spec:
   draDriver:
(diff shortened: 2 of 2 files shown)


controllers/gpucluster_controller.go
```diff
@@ -97,12 +97,6 @@ func (r *GPUClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if !instance.DeletionTimestamp.IsZero() {
 		return r.reconcileDelete(ctx, instance)
 	}
-	if !controllerutil.ContainsFinalizer(instance, gpuClusterFinalizer) {
```
rahulait, line 100: If a user doesn't specify finalizer by default, we should add it. So we should keep this logic as it is and just add finalizer by default in our templates.

tariq1890: Yeah, that's fair. I was thinking about this just after pushing the commit

tariq1890: Addressed your comment. Please check now

bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml
```diff
@@ -158,7 +158,10 @@ metadata:
           "apiVersion": "nvidia.com/v1alpha1",
           "kind": "GPUCluster",
           "metadata": {
-            "name": "gpu-cluster"
+            "name": "gpu-cluster",
+            "finalizers": [
+              "gpucluster.nvidia.com/dra-resourceclaim"
```
rajathagasthya, line 163: This is somewhat of a behavior change. We're pre-adding the finalizer instead of adding it after the operator's first reconcile. If the operator never comes up during `helm install`, you cannot delete this CR without manually removing the finalizer. I'd rather have one way to add and remove the finalizer. Not a strong opinion, but I'm not entirely sure this change is worth it.

This also means we need to maintain this finalizer in multiple places instead of a single constant in the controller.

rahulait:
I agree with @rajathagasthya here. If we are testing things with DRA and operator fails to come up, cleaning of gpucluster resource would require manually removing the finalizer, which is not great.

tariq1890:
Thanks for the reviews. I agree with the comments. Closing...

PR #2676, opened 2026-07-24
add finaliser to GPUCluster CR using PATCH

tariq1890:
This PR adds a new util method to add a finaliser to the GPUCluster custom resource. It also uses `PATCH` to add the finaliser as it's safer and more robuse to race conditions around object updates.

h/t to @rajathagasthya for pointing me to the [method](https://github.com/kubernetes-sigs/cluster-api/blob/a51856a032639cebcfed15fcb0ec9f907c442e1e/util/finalizers/finalizers.go#L31) in CAPI that inspired this change

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -17,6 +17,7 @@
 package utils
 
 import (
+	"context"
 	"fmt"
 	"hash"
 	"hash/fnv"
@@ -29,6 +30,8 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // GetFilesWithSuffix returns all files under a given base directory that have a specific suffix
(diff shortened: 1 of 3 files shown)


internal/utils/utils.go
```diff
@@ -163,3 +166,15 @@ func WriteFileAtomically(path, content string) error {
 	}
 	return nil
 }
+
+// EnsureFinalizer adds a finalizer if an object marked for deletion does not have a finalizer already set.
```
rahulait: This comment seems wrong. Aren't we skipping adding finalizer if object is marked for deletion? The comment says other way.

tariq1890: Thanks for catching that :)

tariq1890: Fixed.

rahulait (approved):

tariq1890:
Thanks for the review @rahulait. I updated the PR to address a review comment from Claude and also added unit tests

```
 Missing optimistic lock (internal/utils/utils.go:179): client.MergeFrom generates a patch with no resourceVersion, so the API server applies it without conflict detection. Two concurrent reconcilers can silently clobber each other's finalizers. The old r.Update included resourceVersion and would have 409'd. Fix: replace client.MergeFrom(original) with client.MergeFromWithOptions(original, client.MergeFromWithOptimisticLock{}).
 ```

rajathagasthya (approved):
Thanks!

PR #2678, opened 2026-07-25
fix nvidiadriver status for multiple default nvidiadrivers

rahulait:
Detect multiple default NVIDIADrivers during validation, mark the affected CRs NotReady with a ReconcileFailed condition, and retain the fail-closed owner-assignment behavior.

Add unit and E2E coverage for duplicate defaults and recovery after the duplicate is removed.

## Bug
In PR #2569, we moved AssignOwners() from NVIDIADriverReconciler into the new NodeLabelingReconciler to centralize node-label writes.

Before that change, the NVIDIADriver reconciler called AssignOwners() directly. When multiple NVIDIADriver CRs had `spec.default: true`, `AssignOwners()` returned an error and the reconciler:
- set the reconciled NVIDIADriver’s status.state to notReady
- set a ReconcileFailed condition containing the duplicate-default error
- returned the error to retry reconciliation

After PR #2569, `AssignOwners()` continued to detect duplicate defaults and fail closed, but it ran only in NodeLabelingReconciler. That controller returned the error without updating any NVIDIADriver status. At the same time, NVIDIADriverReconciler's selector validator intentionally skips default drivers while checking node-selector overlap, so duplicate defaults passed validation and could be marked ready.

As a result, duplicate defaults were still prevented from changing node owner labels, but the invalid CRs no longer reported notReady or exposed the error in their status conditions.

## Fix
The fix restores status reporting without moving owner assignment back into NVIDIADriverReconciler.

- Extend NVIDIADriver validation to reject multiple default NVIDIADrivers.
- Preserve the existing owner-assignment behavior: NodeLabelingReconciler remains the sole writer of node owner labels, and AssignOwners() still fails closed before changing labels.
- When duplicate defaults are detected, NVIDIADriverReconciler sets `status.state=notReady`, adds a ReconcileFailed condition with the duplicate-default message, and returns the error for retry.
- Ignore NVIDIADrivers being deleted when checking duplicate defaults, consistent with owner assignment behavior.
- Keep the existing selector-conflict behavior unchanged: normal overlapping non-default selectors continue to use the ConflictingNodeSelector condition reason.

The NVIDIADriver controller already requeues all NVIDIADrivers after a NVIDIADriver create, update, or delete. Therefore, both duplicate default CRs converge to notReady; after one is deleted or changed to non-default, the remaining default reconciles normally again.

## Checklist

## Testing

Deployed a cluster with multiple nodes and drivers managed by nvidiadriver.

### Before the fix
If 2 default nvidiadrivers were created, their status was showing up as ready even though they conflicted with one another.

### After the fix
Created a k8s cluster and added multiple default nvidiadrivers. They correctly went into notReady state on conflict.

### DIFF
--- a/tests/scripts/update-nvidiadriver.sh
+++ b/tests/scripts/update-nvidiadriver.sh
@@ -13,6 +13,7 @@ source ${SCRIPT_DIR}/checks.sh
 
 NVIDIA_DRIVER_NAME="${NVIDIA_DRIVER_NAME:-e2e-driver}"
 DEFAULT_NVIDIA_DRIVER_NAME="${DEFAULT_NVIDIA_DRIVER_NAME:-e2e-default-driver}"
+DUPLICATE_DEFAULT_NVIDIA_DRIVER_NAME="${DUPLICATE_DEFAULT_NVIDIA_DRIVER_NAME:-e2e-duplicate-default-driver}"
 
 get_default_nvidiadriver_name() {
     kubectl get nvidiadriver -o json |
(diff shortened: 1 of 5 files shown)


tests/scripts/update-nvidiadriver.sh
```diff
@@ -308,6 +309,18 @@ test_removed_default_field_conflict_preserves_owners() {
     wait_for_nvidiadriver_owner "${NVIDIA_DRIVER_NAME}"
 }
 
+test_multiple_default_drivers_are_not_ready() {
```
tariq1890, line 337: I am still wondering if this needs to be added to the e2e suite. Shouldn't the unit tests suffice for this?

rahulait: Yes, e2e is not really required. Unit test covers it. I'll remove it then.

rahulait: Looking more closely, this e2e test is checking that when the duplicate default nvidiadriver is deleted, the controller automatically reconciles and the cluster recovers on its own.
A unit test could delete the duplicate and then call reconcile directly, but that skips the watch predicate and queueing path entirely. It would confirm the recovery logic once reconciliation runs, but not that the delete event actually causes reconciliation in a real cluster.
Since recovery now relies on that watch event instead of retrying failed reconciliations, keeping the e2e test makes sense.

tariq1890 (approved):

PR #2682, opened 2026-07-27
Reject incomplete image specs in ImagePath

abrarshivani:
## Description

`ImagePath` (`internal/image`) and its duplicate `imagePath` (`api/nvidia/v1`, used by `v1.ImagePath`) build a component's image reference from the CR's `repository`/`image`/`version` fields. When `repository` or `version` is set (so the CR branch is taken) but `repository` or `image` is empty, they previously concatenated the parts into a **structurally invalid reference** — `"/image:tag"`, `"repo/:tag"`, or `"/@sha256:..."` — which then failed much later at image-pull time with an opaque error.

This change makes both functions **fail fast with a clear error** when `repository` or `image` is empty in that branch:

```
invalid image specification: both repository and image must be set (repository=%q, image=%q, version=%q)
```

The kbld/carvel passthrough (`repository` **and** `version` both empty → `image` used verbatim) and the env-var fallback are unchanged.

### Tests
- `internal/image`: `TestImagePath_ValidSpec` (fully-specified specs still resolve) and `TestImagePath_InvalidSpec` (incomplete specs are rejected, and an env value does not rescue a partial CR spec).
- `api/nvidia/v1`: `TestImagePath` covering valid, kbld, incomplete-spec and unsupported-type paths.

Split out of #2656 (test-coverage PR) per review, so the validation logic is tracked on its own. Affected suites (`api`, `controllers`, `internal/state`) pass unchanged — no caller relied on the malformed-path behavior.

## Checklist

## Testing

```
go test ./internal/image/... ./api/nvidia/... ./internal/state/... ./controllers/...
golangci-lint run ./internal/image/... ./api/nvidia/v1/...   # 0 issues (GOOS=linux)
```

### DIFF
--- a/api/nvidia/v1/clusterpolicy_types.go
+++ b/api/nvidia/v1/clusterpolicy_types.go
@@ -2061,6 +2061,9 @@ func imagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		if repository == "" || image == "" || version == "" {
+			return "", fmt.Errorf("invalid image specification: repository, image and version must all be set (repository=%s, image=%s, version=%s)", repository, image, version)
+		}
 		// use @ if image digest is specified instead of tag
 		if strings.HasPrefix(version, "sha256:") {
 			crdImagePath = repository + "/" + image + "@" + version
--- a/internal/image/image.go
+++ b/internal/image/image.go
@@ -32,6 +32,9 @@ func ImagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		if repository == "" || image == "" || version == "" {
+			return "", fmt.Errorf("invalid image specification: repository, image and version must all be set (repository=%s, image=%s, version=%s)", repository, image, version)
+		}
 		// use @ if image digest is specified instead of tag
 		if strings.HasPrefix(version, "sha256:") {
 			crdImagePath = repository + "/" + image + "@" + version
--- a/api/nvidia/v1/clusterpolicy_types_test.go
+++ b/api/nvidia/v1/clusterpolicy_types_test.go
@@ -1,0 +1,74 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package v1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestImagePath(t *testing.T) {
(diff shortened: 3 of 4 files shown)


api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -2061,6 +2061,12 @@ func imagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		// A partial CR spec (repository or image missing) can only produce a
+		// malformed reference such as "/image:tag" or "repo/:tag", so fail fast
+		// with a clear error instead of emitting an invalid image path.
```
tariq1890: This comment block is not necessary. It is already understood from the code IMO

tariq1890: If we want to prove this context, it can be a part of the commit description instead. Readers can find this through a `git log|blame` lookup.

abrarshivani: I was planning to fix this before marking the PR as ready for review. Thanks for pointing this out.

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -2061,6 +2061,9 @@ func imagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		if repository == "" || image == "" {
+			return "", fmt.Errorf("invalid image specification: both repository and image must be set (repository=%q, image=%q, version=%q)", repository, image, version)
```
```suggestion
			return "", fmt.Errorf("invalid image specification: both repository and image must be set (repository=%s, image=%s, version=%s)", repository, image, version)
```

internal/image/image.go
```diff
@@ -32,6 +32,9 @@ func ImagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		if repository == "" || image == "" {
+			return "", fmt.Errorf("invalid image specification: both repository and image must be set (repository=%q, image=%q, version=%q)", repository, image, version)
```
tariq1890:
```suggestion
			return "", fmt.Errorf("invalid image specification: both repository and image must be set (repository=%s, image=%s, version=%s)", repository, image, version)
```

api/nvidia/v1/clusterpolicy_types.go
tariq1890: In gpu-operator, the logs are emitted as JSON. Since `%q` quotes the strings, the strings end up  being printed as `\"<string>\"`. `%s` prevents the escaping, so it makes it more readable

abrarshivani: Makes sense. Will update this.

internal/image/image.go
abrarshivani: ditto

tariq1890:
Can you squash the commit history?

tariq1890 (approved):

api/nvidia/v1/clusterpolicy_types.go
```diff
@@ -2061,6 +2061,9 @@ func imagePath(repository string, image string, version string, imagePathEnvName
 			crdImagePath = image
 		}
 	} else {
+		if repository == "" || image == "" {
```
rahulait: Should we also check for version here?
```
 if repository == "" || image == "" || version == "" {
```

tariq1890: Good point. I think, yes

abrarshivani: Good catch! Will address this case as well.

abrarshivani:
@rahulait PTAL. I have addressed the comments.

rahulait (approved):

tariq1890 (approved):

PR #2689, opened 2026-07-28
Enforce singleton GPUCluster name and select ClusterPolicy by creation time

karthikvetrivel:
With multiple CRs of a kind, controllers resolved "the active one" inconsistently: the owning controllers kept a first-reconciled-wins claim in memory (unstable across operator restarts, invisible to other controllers), while `resolveActiveConfig` picked the first list item. This PR makes the singleton deterministic for both kinds:

- **GPUCluster**: a CEL rule on the CRD pins `metadata.name` to `gpu-cluster` (the name the Helm chart creates), enforcing the singleton at admission. That makes the controller's runtime singleton handling dead code, so it is removed: the in-memory first-wins claim and the `ignored` status value are gone, and the sample is renamed to `gpu-cluster`. GPUCluster has not shipped, so there is no upgrade path to consider.
- **ClusterPolicy**: nothing enforces a singleton at admission, so the instance with the oldest `creationTimestamp` (name as tie-breaker) is treated as the singleton, via a new `getSingletonClusterPolicy` helper used by the ClusterPolicy controller, `resolveActiveConfig`, and the GPUCluster prerequisite check (which now inspects only the singleton ClusterPolicy). This selection is a heuristic: ideally the operator would hold a reference to the singleton ClusterPolicy globally throughout its lifetime, and until it does, picking the oldest is the next-best approach.

Deleting the singleton ClusterPolicy promotes the next-oldest lazily (on its next reconcile), same as today.

### DIFF
--- a/api/nvidia/v1alpha1/gpucluster_types.go
+++ b/api/nvidia/v1alpha1/gpucluster_types.go
@@ -26,11 +26,6 @@ const (
 	GPUClusterCRDName = "GPUCluster"
 )
 
-const (
-	// Ignored marks a duplicate GPUCluster that the singleton controller does not reconcile.
-	Ignored State = "ignored"
-)
-
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // GPUClusterSpec defines the desired state of GPUCluster, the DRA-based
@@ -170,7 +165,7 @@ type HostPathsSpec struct {
 
 // GPUClusterStatus defines the observed state of GPUCluster
 type GPUClusterStatus struct {
-	// +kubebuilder:validation:Enum=ignored;ready;notReady;disabled
+	// +kubebuilder:validation:Enum=ready;notReady;disabled
 	// State indicates the status of the GPUCluster instance
 	State State `json:"state"`
 	// Namespace indicates the namespace in which the operator and operands are installed
@@ -186,6 +181,7 @@ type GPUClusterStatus struct {
 //+kubebuilder:resource:scope=Cluster,shortName={"gc"}
 //+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.state`,priority=0
 //+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,priority=0
+//+kubebuilder:validation:XValidation:rule="self.metadata.name == 'gpu-cluster'",message="GPUCluster is a singleton, metadata.name must be 'gpu-cluster'"
 
 // GPUCluster is the Schema for the gpuclusters API
 type GPUCluster struct {
--- a/controllers/active_config.go
+++ b/controllers/active_config.go
@@ -27,16 +27,38 @@ import (
 	"github.com/NVIDIA/gpu-operator/internal/consts"
 )
 
-// TODO: with multiple CRs of a kind, the tie-breaker is list order, which is not
-// guaranteed. Resolve the singleton the reconcilers actually selected instead.
-func resolveActiveConfig(ctx context.Context, c client.Client) (*gpuv1.ClusterPolicy, *nvidiav1alpha1.GPUCluster, error) {
-	var clusterPolicy *gpuv1.ClusterPolicy
+// getSingletonClusterPolicy returns the ClusterPolicy treated as the cluster-wide
+// singleton: oldest creationTimestamp first, lowest name as the tie-breaker. This
+// selection is a heuristic: ideally the operator would hold a reference to the
+// singleton ClusterPolicy globally throughout its lifetime, and until it does,
+// picking the oldest is the next-best approach. Unlike a first-reconciled-wins claim
+// held in controller memory, the result is stable across operator restarts and
+// derivable by every controller. Returns nil for an empty list.
+// GPUCluster needs no selection: its CRD pins metadata.name, enforcing the singleton
+// at admission.
+func getSingletonClusterPolicy(items []gpuv1.ClusterPolicy) *gpuv1.ClusterPolicy {
+	var active *gpuv1.ClusterPolicy
+	for i := range items {
+		candidate := &items[i]
+		if active == nil {
+			active = candidate
+			continue
+		}
+		candidateCreated, activeCreated := candidate.CreationTimestamp, active.CreationTimestamp
+		if candidateCreated.Before(&activeCreated) ||
+			(candidateCreated.Equal(&activeCreated) && candidate.Name < active.Name) {
+			active = candidate
+		}
+	}
+	return active
+}
+
+// resolveActiveConfig returns the active ClusterPolicy, selected with the same
+// getSingletonClusterPolicy rule the ClusterPolicy controller uses, and the GPUCluster.
+func resolveActiveConfig(ctx context.Context, c client.Reader) (*gpuv1.ClusterPolicy, *nvidiav1alpha1.GPUCluster, error) {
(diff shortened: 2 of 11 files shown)


controllers/active_config.go
```diff
@@ -20,35 +20,52 @@ import (
...
 	gpuClusters := &nvidiav1alpha1.GPUClusterList{}
 	if err := c.List(ctx, gpuClusters); err != nil {
 		return nil, nil, fmt.Errorf("failed to list GPUCluster: %w", err)
 	}
-	if len(gpuClusters.Items) > 0 {
-		gpuCluster = &gpuClusters.Items[0]
-	}
 
-	return clusterPolicy, gpuCluster, nil
+	return electSingleton(clusterPolicies.Items), electSingleton(gpuClusters.Items), nil
```
rajathagasthya: Do we need `electSingleton(gpuClusters.Items)` if singleton is enforced by the CEL expression?

karthikvetrivel: No, we don't. Removed.

I also had to deprecate a few fields too, like the `Ignored` state.

cdesiniotis (approved):

rajathagasthya (approved):

controllers/active_config.go
```diff
@@ -27,17 +27,36 @@ import (
...
-// guaranteed. Resolve the singleton the reconcilers actually selected instead.
-func resolveActiveConfig(ctx context.Context, c client.Client) (*gpuv1.ClusterPolicy, *nvidiav1alpha1.GPUCluster, error) {
-	var clusterPolicy *gpuv1.ClusterPolicy
+// electSingleton returns the ClusterPolicy that wins the cluster-wide singleton
+// election: oldest creationTimestamp first, lowest name as the tie-breaker. Unlike a
+// first-reconciled-wins claim held in controller memory, the result is stable across
+// operator restarts and derivable by every controller. Returns nil for an empty list.
+// GPUCluster needs no election: its CRD pins metadata.name, enforcing the singleton
+// at admission.
+func electSingleton(items []gpuv1.ClusterPolicy) *gpuv1.ClusterPolicy {
```
tariq1890: I think "select" is a more appropriate term here.

api/nvidia/v1alpha1/gpucluster_types.go
```diff
@@ -186,6 +181,7 @@ type GPUClusterStatus struct {
 //+kubebuilder:resource:scope=Cluster,shortName={"gc"}
 //+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.state`,priority=0
 //+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,priority=0
+//+kubebuilder:validation:XValidation:rule="self.metadata.name == 'gpu-cluster'",message="GPUCluster is a singleton, metadata.name must be 'gpu-cluster'"
```
tariq1890, line 184: I want to make sure we have completely thought through the name for our singleton resource here.

controllers/active_config.go
tariq1890: Or "getSingleton"

karthikvetrivel: `getSingleton` is fine by me.

api/nvidia/v1alpha1/gpucluster_types.go
cdesiniotis: I am fine with `gpu-cluster`. Did you have any alternatives in mind?

controllers/active_config.go
tariq1890: Thanks

api/nvidia/v1alpha1/gpucluster_types.go
tariq1890: I was thinking if it should be `nvidia-gpu-cluster` instead? I don't have a strong preference here I suppose. If the others prefer `gpu-cluster` as the name, we will just proceed with that

tariq1890:
@karthikvetrivel Can we amend the PR title and commit as well?

Note, the reason I am being particular here is that we don't really `elect` a ClusterPolicy in the way that distributed systems elect a "leader" or "primary actor". Secondly, let'd mention in the comments that the selection of the oldest clusterPolicy as the singleton is a heuristic. Ideally, we should have a way of holding a reference to the singleton `ClusterPolicy` globally throughout the operator's lifetime. Since we don't have this right now, selection of the ClusterPolicy by the oldestTimestamp is the next best approach.

controllers/active_config.go
```diff
@@ -27,17 +27,36 @@ import (
 	"github.com/NVIDIA/gpu-operator/internal/consts"
 )
 
-// TODO: with multiple CRs of a kind, the tie-breaker is list order, which is not
-// guaranteed. Resolve the singleton the reconcilers actually selected instead.
-func resolveActiveConfig(ctx context.Context, c client.Client) (*gpuv1.ClusterPolicy, *nvidiav1alpha1.GPUCluster, error) {
-	var clusterPolicy *gpuv1.ClusterPolicy
+// getSingleton returns the ClusterPolicy that wins the cluster-wide singleton
+// election: oldest creationTimestamp first, lowest name as the tie-breaker. Unlike a
```
tariq1890: Let's remove the general usage of "election" in this PR please

controllers/active_config.go
```diff
@@ -27,17 +27,36 @@ import (
...
+		candidateCreated, activeCreated := candidate.CreationTimestamp, active.CreationTimestamp
+		if candidateCreated.Before(&activeCreated) ||
+			(candidateCreated.Equal(&activeCreated) && candidate.Name < active.Name) {
+			active = candidate
+		}
+	}
+	return active
+}
+
+// resolveActiveConfig returns the active ClusterPolicy, elected with the same
```
tariq1890: here too.

controllers/active_config.go
```diff
@@ -27,17 +27,36 @@ import (
...
-// guaranteed. Resolve the singleton the reconcilers actually selected instead.
-func resolveActiveConfig(ctx context.Context, c client.Client) (*gpuv1.ClusterPolicy, *nvidiav1alpha1.GPUCluster, error) {
-	var clusterPolicy *gpuv1.ClusterPolicy
+// getSingleton returns the ClusterPolicy that wins the cluster-wide singleton
+// election: oldest creationTimestamp first, lowest name as the tie-breaker. Unlike a
+// first-reconciled-wins claim held in controller memory, the result is stable across
+// operator restarts and derivable by every controller. Returns nil for an empty list.
+// GPUCluster needs no election: its CRD pins metadata.name, enforcing the singleton
+// at admission.
+func getSingleton(items []gpuv1.ClusterPolicy) *gpuv1.ClusterPolicy {
```
tariq1890:
```suggestion
func getSingletonClusterPolicy(items []gpuv1.ClusterPolicy) *gpuv1.ClusterPolicy {
```

controllers/active_config.go
karthikvetrivel: Updated.

controllers/active_config.go
karthikvetrivel: Updated.

controllers/active_config.go
karthikvetrivel: Updated.

api/nvidia/v1alpha1/gpucluster_types.go
tariq1890: Resolving as the consensus is to go with `gpu-cluster`

tariq1890 (approved):

PR #2697, opened 2026-07-30
Enable DCGM exporter DRA attribution when pod metadata is enabled

karthikvetrivel:
The DRA DCGM exporter DaemonSet never sets `KUBERNETES_ENABLE_DRA`, and its ClusterRole has no `resource.k8s.io` read access, so metrics on DRA nodes carry no pod/claim attribution even with `dcgmExporter.enablePodLabels` set.

Render both under the existing `PodMetadataEnabled` gate: the env var on the DaemonSet, and `resourceslices` get/list/watch on the ClusterRole. The grant matches the upstream dcgm-exporter chart's [clusterrole.yaml](https://github.com/NVIDIA/dcgm-exporter/blob/main/deployment/templates/clusterrole.yaml), but as a separate rule: upstream's combined pods+resourceslices rule cross-products into permissions the operator does not hold, so RBAC escalation prevention rejects the create when the operator (rather than an admin running helm) applies it. No `resourceclaims` grant is needed — the exporter's DRA path is a single ResourceSlice informer, and claim attribution comes from the kubelet PodResources socket.

Adds a `gpucluster-dcgm-exporter-pod-metadata` render golden covering the pod-metadata-enabled manifests.

Validated on a T4 node: with `enablePodLabels: true`, the operator renders the ClusterRole without escalation errors and metrics include `pod`, `namespace`, `container`, and `dra_*` labels for DRA-allocated GPUs.

### DIFF
--- a/internal/state/gpucluster_render_test.go
+++ b/internal/state/gpucluster_render_test.go
@@ -131,6 +131,18 @@ func TestGPUClusterRenderGolden(t *testing.T) {
 			},
 		},
 		{
+			// enablePodLabels: pod metadata mounts the ServiceAccount token and adds
+			// the DRA/pod-labels env and the read-pods ClusterRole/Binding.
+			name: "gpucluster-dcgm-exporter-pod-metadata",
+			render: func(t *testing.T) []*unstructured.Unstructured {
+				s := newTestDCGMExporterState(t, false)
+				cr := exporterCR(&nvidiav1.DCGMExporterSpec{EnablePodLabels: ptr.To(true)})
+				objs, err := s.getManifestObjects(context.Background(), cr, draSupportedCatalog())
+				require.NoError(t, err)
+				return objs
+			},
+		},
+		{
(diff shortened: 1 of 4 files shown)


rahulait (approved):
LGTM

rajathagasthya (approved):

tariq1890:
Can we align this ClusterRole manifest with the upstream dcgm-exporter helm chart?

http://github.com/NVIDIA/dcgm-exporter/blob/main/deployment/templates/clusterrole.yaml

tariq1890 (approved):

karthikvetrivel:
@tariq1890 done (but still using the existing formatting of the `.yaml`)

tariq1890 (changes requested):
Thanks @karthikvetrivel. The changes look good to me now. Let's please add a YAML render test case for this. We want a golden file that renders the DRA manifests when `.PodMetadataEnabled` evaluates to true

tariq1890 (approved):

tariq1890:
Thanks @karthikvetrivel !

PR #2707, opened 2026-07-31
dra: fix HOST_ROOT envvar and add missing mounts for PassthroughSupport

cdesiniotis:
## Description

This PR fixes a few bugs encountered when enabling the PassthroughSupport feature gate in the NVIDIA DRA Driver for GPUS. First, the `HOST_ROOT` variable was not set correctly, leading to the following error message in the kubelet-plugin:

```
Error: invalid CLI flags: host root is not mounted at "/host-root"
```

After this issue was fixed, the below issue was encountered as we were missing some volume mounts required for the PassthroughSupport feature.

```
Error: error creating driver: error enumerating all possible devices: error enumerating GPU PCI devices: error getting GPU info from PCI device: error finding best VFIO driver for device "0009:01:00.0": failed to get vfio_pci aliases in modules.alias file: failed to read file /lib/modules/5.14.0-687.20.1.el9_8.aarch64/modules.alias: open /lib/modules/5.14.0-687.20.1.el9_8.aarch64/modules.alias: no such file or directory
```

See upstream DRA driver manifest: https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/50e0e3568f633d87266e3aabf8afa510843d6c7d/deployments/helm/dra-driver-nvidia-gpu/templates/kubeletplugin.yaml#L338-L353

## Checklist

- [X] Test cases are added for new code paths

## Testing

Manual testing on a single node OpenShift cluster. Deployed with and without this change.

### DIFF
--- a/internal/state/dra_driver.go
+++ b/internal/state/dra_driver.go
@@ -19,8 +19,6 @@ package state
 import (
 	"context"
 	"fmt"
-	"sort"
-	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -127,7 +125,7 @@ func (s *stateDRADriver) getManifestObjects(ctx context.Context, cr *nvidiav1alp
 		Namespace:             s.namespace,
 		OpenshiftVersion:      openshiftVersion,
 		DeviceClassAPIVersion: apiVersion,
-		FeatureGates:          renderDRAFeatureGates(cr.Spec.DRADriver.FeatureGates),
+		FeatureGates:          cr.Spec.DRADriver.FeatureGates,
 		GPUsHealthcheckPort: resolveHealthcheckPort(
 			cr.Spec.DRADriver.GPUs.KubeletPlugin.Healthcheck, defaultGPUsHealthcheckPort),
 		ComputeDomainsHealthcheckPort: resolveHealthcheckPort(
@@ -210,21 +208,4 @@ func maxResourceList(resourceLists ...corev1.ResourceList) corev1.ResourceList {
 	}
 
 	return result
-}
-
-// renderDRAFeatureGates renders the feature-gate map as the FEATURE_GATES env value
-// (comma-separated Key=Value). Keys are sorted so the rendered value is a pure
-// function of the input and reconciles do not churn the pod spec. Empty when none.
-func renderDRAFeatureGates(gates map[string]bool) string {
-	names := make([]string, 0, len(gates))
-	for name := range gates {
-		names = append(names, name)
-	}
-	sort.Strings(names)
-
-	pairs := make([]string, 0, len(names))
-	for _, name := range names {
-		pairs = append(pairs, fmt.Sprintf("%s=%t", name, gates[name]))
-	}
-	return strings.Join(pairs, ",")
 }
--- a/manifests/state-dra-driver/0500_daemonset.yaml
+++ b/manifests/state-dra-driver/0500_daemonset.yaml
@@ -168,6 +168,8 @@ spec:
           value: void
         - name: CDI_ROOT
           value: /var/run/cdi
+        - name: HOST_ROOT
+          value: "/host"
         - name: NVIDIA_MIG_CONFIG_DEVICES
           value: all
         - name: NODE_NAME
@@ -188,7 +190,7 @@ spec:
           value: {{ .GPUsHealthcheckPort | quote }}
         {{- if .FeatureGates }}
         - name: FEATURE_GATES
-          value: {{ .FeatureGates | quote }}
+          value: {{ convertStringMapToCommaSeparatedString .FeatureGates | quote }}
         {{- end }}
         {{- range .DRADriver.Spec.GPUs.KubeletPlugin.Env }}
         - name: {{ .Name }}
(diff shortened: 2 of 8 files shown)


internal/state/dra_driver.go
```diff
@@ -127,7 +127,8 @@ func (s *stateDRADriver) getManifestObjects(ctx context.Context, cr *nvidiav1alp
 		Namespace:             s.namespace,
 		OpenshiftVersion:      openshiftVersion,
 		DeviceClassAPIVersion: apiVersion,
-		FeatureGates:          renderDRAFeatureGates(cr.Spec.DRADriver.FeatureGates),
+		FeatureGates:          cr.Spec.DRADriver.FeatureGates,
+		FeatureGatesEnv:       renderDRAFeatureGates(cr.Spec.DRADriver.FeatureGates),
```
tariq1890: Instead of adding an extra data attribute called `FeatureGatesEnv`, you could add `renderDRAFeatureGates` (maybe with a more generic name) to the template FuncMap so that you don't have to pass  `cr.Spec.DRADriver.FeatureGates` twice when constructing renderData.

manifests/state-dra-driver/0500_daemonset.yaml
```diff
@@ -231,6 +233,18 @@ spec:
         - name: driver-install-dir
           mountPath: /driver-root
           mountPropagation: HostToContainer
+        {{- if and .FeatureGates (and (index .FeatureGates "PassthroughSupport") .FeatureGates.PassthroughSupport) }}
```
tariq1890, line 236:
```suggestion
        {{- if and .FeatureGates (index .FeatureGates "PassthroughSupport") }}
```

internal/state/dra_driver.go
cdesiniotis: Fair point. Let me update this.

manifests/state-dra-driver/0500_daemonset.yaml
cdesiniotis: As currently implemented, `(and (index .FeatureGates "PassthroughSupport") .FeatureGates.PassthroughSupport)` returns true if the `PassthroughSupport` key exists and its value is `true`.  This mimics what is done here https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/50e0e3568f633d87266e3aabf8afa510843d6c7d/deployments/helm/dra-driver-nvidia-gpu/templates/kubeletplugin.yaml#L338-L353. Dropping the right-hand-side of the conditional would return true even if `PassthroughSupport=false`.

internal/state/dra_driver.go
cdesiniotis: Updated.

internal/state/dra_driver.go
```diff
@@ -127,14 +129,19 @@ func (s *stateDRADriver) getManifestObjects(ctx context.Context, cr *nvidiav1alp
...
-		FeatureGates:          renderDRAFeatureGates(cr.Spec.DRADriver.FeatureGates),
+		FeatureGates:          cr.Spec.DRADriver.FeatureGates,
 		GPUsHealthcheckPort: resolveHealthcheckPort(
 			cr.Spec.DRADriver.GPUs.KubeletPlugin.Healthcheck, defaultGPUsHealthcheckPort),
 		ComputeDomainsHealthcheckPort: resolveHealthcheckPort(
 			cr.Spec.DRADriver.ComputeDomains.KubeletPlugin.Healthcheck, defaultComputeDomainsHealthcheckPort),
 	}
 
-	return s.renderObjects(ctx, renderData)
+	return s.renderer.RenderObjects(&render.TemplatingData{
```
tariq1890: We don't have to change this line.  We could instead add to the FuncMap [here](https://github.com/NVIDIA/gpu-operator/blob/e72ea5a66908863c210217744660453f0bbbb30b/internal/render/render.go#L114).

internal/state/dra_driver.go
```diff
@@ -127,14 +129,19 @@ func (s *stateDRADriver) getManifestObjects(ctx context.Context, cr *nvidiav1alp
...
+		Funcs: template.FuncMap{
+			"renderDRAFeatureGatesAsEnvVar": renderDRAFeatureGatesAsEnvVar,
```
tariq1890:
```suggestion
			"renderMapAsCommaSeparatedString": renderMapAsCommaSeparatedString,
```

tariq1890: ^ It doesn't have to be that exact name. Basically, anything generic would work

tariq1890: or how about `convertStringMapToEnvVars` ?

internal/state/dra_driver.go
cdesiniotis: You are right. Updated as suggested.

internal/state/dra_driver.go
cdesiniotis: I landed on `convertStringMapToCommaSeparatedString`.

tariq1890:
LGTM. Once `gofmt` is applied, this should be good to merge.

cdesiniotis:
> LGTM. Once `gofmt` is applied, this should be good to merge.

Fixed now.

tariq1890 (approved):

PR #2708, opened 2026-08-01
nvd: update nodeSelector to only use major version for rhel-like distros

cdesiniotis:
## Description

In a prior [commit](https://github.com/NVIDIA/gpu-operator/commit/f49a33978eb03c0a6ed09cbff927f74645274d44), we updated the NVIDIADriver controller so that only one driver daemonset would get created per major version for rhel-like distros (rhel, oracle linux, rocky linux). The image tag and name of the daemonset only contain the major version, for example rhel9 (and not rhel9.x). One item that was not updated was the rendered nodeSelector in the daemonset. The following label is still being set in the nodeSelector:

```
feature.node.kubernetes.io/system-os_release.VERSION_ID
```

This commit updates the nodeSelector so that only the major version is referenced for rhel-like distros. The above label is replaced by the below label in the rendered daemonset:

```
feature.node.kubernetes.io/system-os_release.VERSION_ID.major
```

## Checklist

## Testing

Unit tests

### DIFF
--- a/internal/state/driver.go
+++ b/internal/state/driver.go
@@ -50,8 +50,9 @@ import (
 )
 
 const (
-	nfdOSReleaseIDLabelKey = "feature.node.kubernetes.io/system-os_release.ID"
-	nfdOSVersionIDLabelKey = "feature.node.kubernetes.io/system-os_release.VERSION_ID"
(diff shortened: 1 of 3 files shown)


rahulait (approved):
Thanks @cdesiniotis for fixing it, it was introduced by my change. LGTM

PR #2714, opened 2026-08-04
fix node permissions for computedomain plugin

rahulait:
## Description

These permissions are coming from: https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/9c52a7d50994adbf2fbb5f1ce2f6466fa3f9936f/deployments/helm/dra-driver-nvidia-gpu/templates/rbac-kubeletplugin.yaml#L17-L19
## Checklist

## Testing

Verified using:
```
kubectl auth can-i patch nodes   --as=system:serviceaccount:gpu-operator:nvidia-dra-driver-kubeletplugin
Warning: resource 'nodes' is not namespace scoped

yes
```

### DIFF
--- a/manifests/state-dra-driver/0200_kubeletplugin-clusterrole.yaml
+++ b/manifests/state-dra-driver/0200_kubeletplugin-clusterrole.yaml
@@ -33,15 +33,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - nodes
-  verbs:
-  - get
-  - list
-  - watch
-  - update
-- apiGroups:
-  - ""
-  resources:
   - pods
   verbs:
   - get
--- a/manifests/state-dra-driver/0200_kubeletplugin-clusterrole.yaml
+++ b/manifests/state-dra-driver/0200_kubeletplugin-clusterrole.yaml
@@ -39,6 +39,7 @@ rules:
   - list
   - watch
   - update
+  - patch
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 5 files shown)


manifests/state-dra-driver/0200_kubeletplugin-clusterrole.yaml
```diff
@@ -30,15 +30,6 @@ rules:
   - create
   - update
   - delete
-- apiGroups:
```
rajathagasthya, line 33: Why are these permissions removed? Aren't they required when DRA is deployed in `gpus` (non-ComputeDomain) mode?

rahulait: No, it looks like all node client calls are in the computeDomain components. These components are not started in gpu only mode. cc @shivamerla to confirm as well

rajathagasthya: From offline conversations, it seems like node permissions are only required in ComputeDomain now.

tariq1890:
Don't we need to add the other permissions mentioned in the ClusterRole from the upstream DRA driver helm chart?

rahulait:
Looks like ubuntu/canonical apt servers are having issues, driver installs are taking ages to finish and our e2e tests won't finish on time.

rahulait:
> Don't we need to add the other permissions mentioned in the ClusterRole from the upstream DRA driver helm chart?

Other required permissions remain in the existing general clusterrole https://github.com/NVIDIA/gpu-operator/blob/main/manifests/state-dra-driver/0200_kubeletplugin-clusterrole.yaml. Are you looking for any specific RBAC which is missing?

cdesiniotis (approved):

rajathagasthya (approved):

rahulait:
@rajathagasthya @cdesiniotis re-requesting review as I've updated the PR based on our discussions.

rajathagasthya (approved):

tariq1890 (approved):

PR #2718, opened 2026-08-06
feat(ci): run e2e tests from the runner via Holodeck remoteAccess

abrarshivani:
Runs the Kubernetes-facing parts of the e2e suite on the GitHub Actions runner instead of inside the Holodeck EC2 node. Draft until CI has run against it.

#### What changes

`kubernetes.remoteAccess` (NVIDIA/holodeck#818) gives the runner a usable kubeconfig, so helm, kubectl, the case scripts and log collection all run there. The repo rsync, the `scp` of the values override, the in-VM tooling installs and `pull.sh` are gone.

Two host-mutating operations stay on the node: `sudo modprobe -a i2c_core ipmi_msghandler`, and the operator container kill in `test_restart_operator`. Both go through a new `tests/scripts/node-exec.sh`, which streams the self-contained `node-operations.sh` over SSH. With `NODE_SSH_HOST` unset it runs locally, so the developer path still works. Container-selection logic is moved verbatim.

This is not a revert. Unlike NVIDIA/k8s-device-plugin#1948, the rsync + ssh model here is the original 2021 design for laptop-driven testing, and CI was already using it on Holodeck v0.2.18, before the kubeconfig permission change.

`local.sh`, `ci-run-e2e.sh`, `push.sh`, `pull.sh`, `sync.sh`, `remote.sh` and `prerequisites.sh` are kept, since they are the documented developer workflow in `tests/README.md`. CI just stops calling them.

#### Polling loops, worth a close look

The loops in `checks.sh` and `update-nvidiadriver.sh` bounded themselves by counting sleeps rather than elapsed time. That held when an iteration cost about 5s on the node. From the runner it does not: the 45 minute bound stretches past the job's 90 minute cap, and a cancelled job skipped the `failure()`-gated artifact upload, so a hung run produced no logs at all.

Twelve loops now measure elapsed time with `SECONDS` against a baseline taken at loop entry. Bounds, messages and exit paths are unchanged. In `wait_for_driver_upgrade_done` the `current_time % 30` debug cadence became a due-time check, since irregular elapsed jumps can miss a modulo window entirely. `check_gpu_pod_ready` was re-fetching every container's full log for every pod every 5s; collection now runs on a 30s cadence, forced before both exit paths, while readiness polling stays at 5s. The artifact upload moves to `always()`.

Still not enforceable: a job runs roughly 31 of these loops at 45 minutes each against a 90 minute cap, so two slow ones back to back can exhaust it. That needs shorter budgets or a shared deadline, which felt like a separate discussion.

#### Testing

Both e2e jobs pass against real Holodeck infrastructure. The run exercised the
whole new path end to end: the OCI chart published and installed, the runner
read the remoteAccess kubeconfig and reached the API server, `node-exec.sh`
loaded the kernel modules over SSH, and the operator restart test killed the
container on the node with `crictl` while the runner watched it come back.
Diagnostics, artifact upload and credential cleanup all ran. Artifacts came out
at roughly 735 KB and 336 KB.

Not exercised yet: `use_values_override: true`, which is the one changed branch
the run skipped, and the timeout/cancellation path that `always()` exists for.

Locally, shellcheck reports no new findings on any modified file, and the
wall-clock conversions were checked against stubs — the old code overran a 10s
budget by 3.8x where the new code exits on time.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -156,6 +156,8 @@ jobs:
             echo "NODE_SSH_HOST=ubuntu@${PUBLIC_DNS_NAME}"
             echo "NODE_SSH_KEY=${SSH_DIR}/id_rsa"
             echo "NODE_SSH_KNOWN_HOSTS=${SSH_DIR}/known_hosts"
+            echo "NODE_PUBLIC_DNS_NAME=${PUBLIC_DNS_NAME}"
+            echo "EGRESS_IP_AT_START=$(curl -fsS --max-time 15 https://api.ipify.org || echo unknown)"
           } >> "$GITHUB_ENV"
       - name: Verify cluster access
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -55,6 +55,12 @@ on:
         type: boolean
         default: false
 
+env:
+  HELM_VERSION: v3.19.0
```
rahulait: We should use more recent version of helm

.github/workflows/e2e-tests.yaml
```diff
@@ -55,6 +55,12 @@ on:
...
+  KUBECTL_VERSION: v1.35.4
+  JQ_VERSION: "1.7.1"
```
rahulait: Lets use newer version if we want to use jq

rahulait (commented):
few nits, sorry forgot to submit review before

PR #2721, opened 2026-08-06
Mount ServiceAccount token when dcgm-exporter env needs API access

rajathagasthya:
## Description

The dcgm-exporter DaemonSet sets `automountServiceAccountToken=false` and re-enables it only when pod metadata enrichment is turned on through `enablePodLabels`/`enablePodUID` fields. Users who configure the same features through `dcgmExporter.env` are missed by that condition, and setting `DCGM_EXPORTER_CONFIGMAP_DATA` crashes the exporter on startup since it cannot read the ServiceAccount token. This worked before the token was made opt-in.

Mount the token when the user-provided env reads a ConfigMap or enables pod labels/pod UID. A `DCGM_EXPORTER_CONFIGMAP_DATA` value of `none` is the exporter default meaning no ConfigMap is read, so it does not count. Configurations that make no API calls keep the default.

Apply the same env check to the cluster-scoped RBAC gate. Enrichment through raw env otherwise got a token but no matching permissions, leaving the pod informer with only the namespaced Role, which grants pod reads in the operator namespace without watch. The ConfigMap case keeps a separate condition since it needs no cluster-scoped RBAC: the namespaced Role already grants configmaps get/list and its RoleBinding is ungated.

## Testing

Metrics are available when `DCGM_EXPORTER_CONFIGMAP_DATA` is set:
```shell
kubectl create configmap -n gpu-operator custom-metrics --from-literal=metrics=\
'DCGM_FI_DEV_GPU_TEMP, gauge, GPU temperature (in C).
DCGM_FI_DEV_POWER_USAGE, gauge, Power draw (in W).'

kubectl patch clusterpolicy cluster-policy --type merge -p \
'{"spec":{"dcgmExporter":{"enablePodLabels":false,"enablePodUID":false,"env":[{"name":"DCGM_EXPORTER_CONFIGMAP_DATA","value":"gpu-operator:custom-metrics"}]}}}'
```

Metrics are available when pod enrichment is set via env:
```shell
kubectl patch clusterpolicy cluster-policy --type merge -p \
'{"spec":{"dcgmExporter":{"env":[{"name":"DCGM_EXPORTER_KUBERNETES_ENABLE_POD_LABELS","value":"true"}]}}}'
```

## Checklist

### DIFF
--- a/controllers/object_controls.go
+++ b/controllers/object_controls.go
@@ -116,6 +116,17 @@ const (
 	DCGMRemoteEngineEnvName = "DCGM_REMOTE_HOSTENGINE_INFO"
 	// DCGMDefaultPort indicates default port bound to DCGM host engine
 	DCGMDefaultPort = 5555
+	// DCGMExporterConfigMapDataEnvName is the env name specifying the namespace:name
+	// ConfigMap with custom metrics
+	DCGMExporterConfigMapDataEnvName = "DCGM_EXPORTER_CONFIGMAP_DATA"
+	// DCGMExporterUndefinedConfigMapData is the exporter default meaning no ConfigMap is read
+	DCGMExporterUndefinedConfigMapData = "none"
+	// DCGMExporterPodLabelsEnvName is the env name enabling pod labels on metrics
+	DCGMExporterPodLabelsEnvName = "DCGM_EXPORTER_KUBERNETES_ENABLE_POD_LABELS"
+	// DCGMExporterPodUIDEnvName is the env name enabling the pod UID label on metrics
+	DCGMExporterPodUIDEnvName = "DCGM_EXPORTER_KUBERNETES_ENABLE_POD_UID"
+	// DCGMExporterPodLabelAllowlistEnvName is the env name holding the exported pod label allowlist
+	DCGMExporterPodLabelAllowlistEnvName = "DCGM_EXPORTER_KUBERNETES_POD_LABEL_ALLOWLIST_REGEX"
 	// GPUDirectRDMAEnabledEnvName indicates if GPU direct RDMA is enabled through GPU operator
 	GPUDirectRDMAEnabledEnvName = "GPU_DIRECT_RDMA_ENABLED"
 	// UseHostMOFEDEnvName indicates if MOFED driver is pre-installed on the host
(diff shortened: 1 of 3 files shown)


controllers/object_controls.go
```diff
@@ -1928,6 +1940,36 @@ func TransformDCGMExporter(obj *appsv1.DaemonSet, config *gpuv1.ClusterPolicySpe
...
+	}
+	return false
+}
+
+// dcgmExporterEnvEnablesPodMetadata returns true if a user-provided env var turns
+// on pod metadata enrichment.
+func dcgmExporterEnvEnablesPodMetadata(envs []gpuv1.EnvVar) bool {
+	for _, env := range envs {
+		if env.Name == DCGMExporterPodLabelsEnvName || env.Name == DCGMExporterPodUIDEnvName {
+			if strings.EqualFold(env.Value, "true") {
```
karthikvetrivel, line 1965: nit: do we want to accept alternative boolean representations here, or is "true" fine?

controllers/object_controls.go
```diff
@@ -116,6 +116,17 @@ const (
 	DCGMRemoteEngineEnvName = "DCGM_REMOTE_HOSTENGINE_INFO"
 	// DCGMDefaultPort indicates default port bound to DCGM host engine
 	DCGMDefaultPort = 5555
+	// DCGMExporterConfigMapDataEnvName is the env name specifying the namespace:name
```
karthikvetrivel, line 119: super nit: is the colon between namespace and name intentional?

tariq1890: Isn't `namespace/name` more conventional?

rajathagasthya: > is the colon between namespace and name intentional?

Yes, this is upstream behavior. See https://docs.nvidia.com/datacenter/cloud-native/gpu-telemetry/latest/dcgm-exporter.html#dcgm-exporter-customization

controllers/object_controls.go
rajathagasthya: This is existing behavior. Based on https://github.com/NVIDIA/dcgm-exporter/blob/main/deployment/templates/daemonset.yaml#L147-L154, I don't think we should change it.

karthikvetrivel: Sounds good.

controllers/object_controls.go
karthikvetrivel: Makes sense, thanks.

karthikvetrivel (approved):

rahulait (approved):

PR #2723, opened 2026-08-07
Add THIRD_PARTY_NOTICES.md and tooling to generate it

abrarshivani:
Adds `THIRD_PARTY_NOTICES.md` and `tools/generate-third-party-notices.sh`, the script that
produces it, so the Go deps behind the released artifacts have their licenses
listed in the repo.

- `make third-party-notices` regenerates it
- `make check-third-party-notices` regenerates and diffs it, and runs in CI on every build

## What to review

97% of the diff is generated. Hand-written, 448 lines across 7 files:

| File | Lines |
|---|---:|
| `tools/generate-third-party-notices.sh` | 367 |
| `.github/workflows/third-party-notices-check.yaml` | 53 |
| `Makefile` | 12 |
| `tools/go.mod` | 7 |
| `.github/workflows/ci.yaml` | 6 |
| `.gitignore` | 2 |
| `tools/tools.go` | 1 |

`tools/generate-third-party-notices.sh` is the one to read: named functions behind a `main()`,
43 of them comments. `tools/go.mod` pins `go-licenses v2.0.1`, and
`.gitignore` covers `/.licenses-cache`, the generator's scratch tree. Generated:
`THIRD_PARTY_NOTICES.md` (17,078 lines) and `tools/go.sum` (97). The index
table at the top of the notices file is the summary; the rest is verbatim
license text.

## How it works

Needs Go (from `versions.mk`) and `go-licenses`, which `make install-tools` puts
in `./bin` from the pin in `tools/go.mod`. The script itself reads `vendor/` and needs no network; `make
third-party-notices` depends on `install-tools`, which does fetch on a cold
checkout.

1. Use `./bin/go-licenses` ahead of `$PATH` so the pinned version runs
2. Check `PLATFORMS` against `DOCKER_BUILD_PLATFORM_OPTIONS` in `multi-arch.mk`,
   fail if out of sync
3. Run `go-licenses save` and `csv` once per platform over `./cmd/...` with
   `GOFLAGS=-mod=vendor` and `CGO_ENABLED=0`, then merge. Platforms are
   `linux/amd64` and `linux/arm64`.
4. Sort whole lines, then join all licenses per package instead of picking one
5. Add `module@version` by longest-prefix match against `vendor/modules.txt`
6. Render the header, the index table, then each license verbatim, fenced wide
   enough to survive licenses that are themselves Markdown
7. Write to a temp file, then move it into place

To add a dep: change `go.mod`, `go mod vendor`, `make third-party-notices`, commit. CI fails
the PR if you forget.

## Implementation notes

- **Run per platform and merged.** `go-licenses` resolves only the host's
  graph, and build-tagged sources pull different deps per platform.
- **Only the local module goes to `--ignore`.** It matches raw string prefixes,
  not path segments, so a stdlib list from `go list std | cut -d/ -f1` yields
  the bare token `go` and drops `golang.org/x/*`, `google.golang.org/*`,
  `gopkg.in/*` and `go.uber.org/*`: 18 modules.
  `go-licenses` already excludes the stdlib.
- **Licenses are joined, not picked.** `go-licenses` emits one row per license,
  so collapsing on the package field loses the rest: `filepath-securejoin` would
  show as BSD-3-Clause without its MPL-2.0. Which one survived was unstable too,
  since a keyed `sort -u` compares only that key and BSD and GNU sort break ties
  differently.
- **Output must match on every machine**, since CI diffs it. Whole-line
  `sort -u`, not `sort -t, -k1,1 -u`. `LC_ALL=C` on every sort and grep, or the
  filename match misses `LICENSE` under a Turkish locale. The awk counts instead
  of testing `in`, which mawk and gawk evaluate differently, and mawk is
  `/usr/bin/awk` on Debian and Ubuntu. `grep -a` so a NUL byte cannot change the
  fence width.
- **Vendored deps use `module@version`** from `vendor/modules.txt`, since
  `go-licenses` in vendor mode reports a `HEAD` URL into this repo that stops
  describing released content once `main` moves. No resolved version fails the
  run.
- **An unidentified or empty license fails the run.** `go-licenses` reports a
  license it cannot classify as `Unknown` and still exits 0, so without a guard
  the file would ship an entry that attributes nothing. Anchored so it catches
  `BSD-3-Clause / Unknown` and a bare `Unknown` but not `Unknown-Something`.
- **`bin/go-licenses` is probed by running it**, not with `-x`: a host-built
  binary bind-mounted into a Linux container passes `-x` but cannot exec.
- **`mv`, not `cp`.** `cp` truncates the target first, so an interrupted run
  leaves a half-written file.

The build toolchain is deliberately not covered. `tools/tools.go` pins
`go-licenses`, `gcov2lcov`, `client-gen`, `controller-gen` and `kustomize`; none
of them appears in any Dockerfile, and the committed generated files import only
Kubernetes libraries and other already-listed dependencies, never a generator.
Attribution attaches to what is redistributed, so listing the tools behind the
build overstated the scope. That removed 74 of the previous 198 index rows.

## Scope

Covers `./cmd/...`, not `./...`. `CMDS` is a wildcard over `./cmd/*` and the
Dockerfile builds by running `make cmds`, so this tracks the build rather than a
hand-kept list. Against `./...` the only extras are `ginkgo`, `gomega` and `golang.org/x/tools` from
the never-shipped e2e suite. `gpuop-cfg` is scanned even though it is a build
helper not copied into the image; listing slightly more than ships is the safe
direction.

Not covered: the non-Go contents of the image, meaning the base image packages,
the BusyBox tree, the CUDA sample and the CUDA compat libraries. Base image deps
go through that image's own compliance process, the rest through separate flows
including any source-distribution obligations. The generated file says so in its
header.

## Shipping

An earlier revision of this PR added `release-third-party-notices.yaml`, which
attached the file to each GitHub release. That workflow has been removed, so the
release flow is untouched by this change. Flagging it because the approval on
this PR predates that removal, and the workflow was the subject of one of the
review threads.

Committed to the repository. Not copied into the runtime image, so image size is
unchanged.

The check runs on every build, with no changed-paths filter. The runtime
list is the set of packages `./cmd/...` imports, so it goes stale when a `.go`
file changes its imports, not only when `go.mod` or `vendor/` move. A filter
would pass green while the file rotted, and the failure would surface on `main`
instead of on the PR that caused it.

## Testing

Current output: 124 packages, 908K.
Index rows are per package; one module can own several.

- Same sha256 on macOS (BSD awk, BSD sort) and in `golang:1.26.5` (mawk 1.3.4,
  GNU sort). Two runs on each host also match.
- Completeness against `go list -deps ./cmd/...` for both released platforms,
  mapped to license-owning dirs: 124 expected, 124 present, 0 missing, 0 extra.
- Every index row has a license-text section. 0 `Unknown` licenses, 0 "License
  text unavailable", 0 unresolved module versions.
- Multi-licensed deps checked against the actual license files:
  `filepath-securejoin` (BSD-3-Clause / MPL-2.0), `klauspost/compress`,
  `sigs.k8s.io/yaml`, `sigs.k8s.io/json`.
- `make check-third-party-notices` passes on the committed file, fails when stale, and fails
  when untracked. `git diff` reports nothing for an untracked path, so that gate
  would otherwise pass silently.
- The CI job has run and passed (1m26s), which also confirms a file generated on
  macOS reproduces byte-identically on the Linux runner.
- Error paths exit 1 with useful messages: missing input file, out-of-sync
  platform matrix, a filesystem `replace` in `vendor/modules.txt`, and an
  unreadable module list. A versioned `replace` correctly uses the
  replacement.
- `shellcheck` clean on the generator. `actionlint` and `yamllint` clean on both
  workflows.


.github/workflows/release-third-party-notices.yaml
```diff
@@ -0,0 +1,94 @@
...
+# Future option: nvcr.io already carries per-digest artifacts for this image
+# (sha256-<digest>.{sig,sbom,vex}) via the cosign tag scheme, since the registry
+# does not implement the OCI 1.1 referrers API. Publishing the notices there
+# would bind them to the exact image digest, but those artifacts are attached by
+# the downstream publishing pipeline rather than by this repository, so it is a
+# request to that pipeline rather than a change here.
+
+name: Release Third-Party Notices
+
+on:
```
rahulait: Do we have a sample run where we can verify this action? Like on your forked repo where this action is triggered on a created sample release? Just want to see what it does.

abrarshivani: It creates third-party notices as a release artifact. You can take a look at my fork here: https://github.com/abrarshivani/gpu-operator/releases

rahulait: Thx

rahulait (approved):
LGTM

tools/generate-notices.sh
```diff
@@ -0,0 +1,523 @@
+#!/usr/bin/env bash
```
rahulait: nit: since its a shell script, I would prefer we keep the .sh extension for the file if possible

abrarshivani: Make sense, updated.

.github/workflows/third-party-notices-check.yaml
```diff
@@ -0,0 +1,62 @@
...
+# only when go.mod or vendor/ move. A filter keyed on dependency manifests
+# therefore passes green while the committed file goes stale, and the failure
+# then surfaces on main instead of on the PR that caused it. A few minutes of
+# runner time is worth more than a gate that can silently pass.
+
+name: Third-Party Notices Check
+
+on:
+  workflow_call:
+  workflow_dispatch:
```
tariq1890: Let's just run this as a CI pipeline for now. Please correct me if I am missing anything, but I don't see any actual value in running this as a manually triggered action 
```suggestion
```

THIRD_PARTY_NOTICES.md
```diff
@@ -0,0 +1,17086 @@
...
+components are handled separately, including any source-distribution
+obligations they carry.
+
+Entries identify their dependency as `module@version` rather than a URL.
+`go-licenses` in vendor mode reports a link into this repository at `HEAD`,
+which stops describing the released content once `main` advances;
+`module@version` is immutable and names the upstream project directly.
+
+This file is generated. Run `make third-party-notices` to regenerate it;
+`make check-third-party-notices` verifies it is current.
```
tariq1890: Let's remove this please

.github/workflows/third-party-notices-check.yaml
```diff
@@ -0,0 +1,62 @@
...
+# Regenerates THIRD_PARTY_NOTICES.md and fails if it differs from the committed
+# copy, so a dependency change cannot land without refreshed attribution.
+#
+# This runs on every build rather than behind a changed-paths filter. A filter
+# was tried and dropped: the runtime inventory is the import closure of
+# ./cmd/..., so it changes when ordinary .go files change their imports, not
+# only when go.mod or vendor/ move. A filter keyed on dependency manifests
+# therefore passes green while the committed file goes stale, and the failure
+# then surfaces on main instead of on the PR that caused it. A few minutes of
+# runner time is worth more than a gate that can silently pass.
```
tariq1890: extraneous
```suggestion
```

tariq1890 (approved):

PR #2724, opened 2026-08-07
Grant the dcgm-exporter DRA role read access to ConfigMaps

rajathagasthya:
## Description

`DCGM_EXPORTER_CONFIGMAP_DATA` set through `dcgmExporter.env` points the
exporter at a ConfigMap of custom metrics that it reads from the Kubernetes
API. On the GPUCluster path neither the namespaced Role nor the ClusterRole
granted `configmaps`, so the read failed with a forbidden error and the
exporter fell back to its built-in metrics without crashing. The ClusterPolicy
path has granted this since before the feature existed.

Add `configmaps` get/list to the namespaced Role. The Role is bound in the
operator namespace only, so a `namespace:name` value pointing elsewhere is
still denied.

Verified on a DRA cluster: the ServiceAccount can read ConfigMaps and the
custom metric set is served. Golden fixtures updated to match.

Follow-up to #2721, which fixes the equivalent problem on the ClusterPolicy
path.

## Checklist

### DIFF
--- a/internal/state/testdata/golden/gpucluster-dcgm-exporter-embedded.yaml
+++ b/internal/state/testdata/golden/gpucluster-dcgm-exporter-embedded.yaml
@@ -18,6 +18,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/manifests/state-dcgm-exporter/0200_role.yaml
+++ b/manifests/state-dcgm-exporter/0200_role.yaml
@@ -12,3 +12,10 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
--- a/internal/state/testdata/golden/gpucluster-dcgm-exporter-pod-metadata.yaml
+++ b/internal/state/testdata/golden/gpucluster-dcgm-exporter-pod-metadata.yaml
@@ -18,6 +18,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/internal/state/testdata/golden/gpucluster-dcgm-exporter-remote-engine.yaml
+++ b/internal/state/testdata/golden/gpucluster-dcgm-exporter-remote-engine.yaml
@@ -18,6 +18,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole


internal/state/testdata/golden/gpucluster-dcgm-exporter-embedded.yaml
```diff
@@ -18,6 +18,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
```
tariq1890, line 26: The [role](https://github.com/NVIDIA/dcgm-exporter/blob/main/deployment/templates/role.yaml) in the upstream helm chart is only requesting a `get` privilege and it is scoped to a particular resource name

```
- apiGroups: [""]
  resources: ["configmaps"]
  resourceNames: ["exporter-metrics-config-map"]
  verbs: ["get"]
```

rajathagasthya: I'm not sure why the upstream chart is scoping this to a particular resource name. The ConfigMap name is user supplied, so it doesn't make sense to restrict the name.

As for having both `get` and `list` permissions, this PR matches what is granted to the ClusterPolicy-based DCGM Exporter role: https://github.com/NVIDIA/gpu-operator/blob/cfabcc9a8c4cc071a7120d320d0d8db79984a166/assets/state-dcgm-exporter/0200_role.yaml#L17-L24

We could scope this to only `get` and only create the role when a ConfigMap is provided, but that would be a deviation from a known ClusterPolicy configuration and would require thorough testing.

rajathagasthya: The upstream role is referring to this ConfigMap containing all custom metrics, and is gated behind `.Values.customMetrics`: https://github.com/NVIDIA/dcgm-exporter/blob/main/deployment/templates/metrics-configmap.yaml. It looks like a sample ConfigMap to me.

We also let users create a ConfigMap from the chart, but unlike upstream, we don't restrict the name and hence don't scope the role to one resource name. https://github.com/NVIDIA/gpu-operator/blob/cfabcc9a8c4cc071a7120d320d0d8db79984a166/deployments/gpu-operator/templates/dcgm_exporter_config.yaml#L1-L14

karthikvetrivel: +1 on aligning with the permissions granted to the ClusterPolicy-based DCGM exporter role.

karthikvetrivel (approved):

rahulait (approved):

manifests/state-dcgm-exporter/0200_role.yaml
```diff
@@ -12,3 +12,10 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
tariq1890, line 18: Should we make this change too if we want to align this with the ClusterPolicy-rendered role?

```suggestion
  - configmaps
  - pods
```

rajathagasthya: No, it's already granted in the ClusterRole: https://github.com/NVIDIA/gpu-operator/blob/599e28368646116dc56a8018a068fbac6334cce1/manifests/state-dcgm-exporter/0210_clusterrole.yaml#L10-L15

DCGM Exporter needs both `pods` and `resourceslices` just to attribute basic DRA metrics.

tariq1890 (approved):

PR #2726, opened 2026-08-10
[state-dcgm-exporter][clusterpolicy] remove pods from Role

tariq1890:
The ClusterRole already provides get, list and watch privileges to the dcgm-exporter daemonset. The privileges in the namespace-scoped are therefore redundant and unnecessary.

### DIFF
--- a/assets/state-dcgm-exporter/0200_role.yaml
+++ b/assets/state-dcgm-exporter/0200_role.yaml
@@ -18,7 +18,6 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
   verbs:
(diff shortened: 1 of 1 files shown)


rajathagasthya (commented):
LGTM. 

Technically, the reason this works is because DCGM exporter uses the `pod-resources` socket for GPU -> Pod mapping, so it doesn't require `pods` RBAC. And the ClusterRole is only rendered if pod enrichment is enabled, so with or without enrichment enabled this never did anything.

rajathagasthya (approved):

PR #2727, opened 2026-08-10
[state-driver] check driver daemons startup status through sentinel file

tariq1890:
This change is accompanied by the driver-container PR https://github.com/NVIDIA/gpu-driver-container/pull/913.

We introduce a new sentinel file called `.driver-daemons-status` which will help the driver container ensure that the Fabric Manager and other daemons are all fully up and running before the driver container goes into the `Running` state.

For reasons of backward compatibility, we only enforce this check if the `.driver-daemons-status` file is present. This way, the operator will continue to work with other driver containers that don't have this capability

### DIFF
--- a/assets/state-driver/0400_configmap.yaml
+++ b/assets/state-driver/0400_configmap.yaml
@@ -13,12 +13,21 @@ data:
 
     VALIDATIONS_DIR="/run/nvidia/validations"
     READY_FILE="${VALIDATIONS_DIR}/.driver-ctr-ready"
+    DAEMONS_STATUS_FILE="${VALIDATIONS_DIR}/.driver-daemons-status"
 
     mkdir -p "${VALIDATIONS_DIR}"
 
     if [ ! -f /sys/module/nvidia/refcnt ]; then
       echo "NVIDIA kernel module not loaded"
       exit 1
+    fi
+    
+    if [ -f "$DAEMONS_STATUS_FILE" ]; then
+      daemons_status=$(cat "$DAEMONS_STATUS_FILE")
+      if [ "$daemons_status" != "Ready" ]; then
+        echo "NVIDIA driver daemons are not ready; Current status: $daemons_status"
+        exit 1
+      fi
     fi
 
     if ! nvidia-smi; then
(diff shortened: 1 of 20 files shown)


shivamerla:
looks good to me!

assets/state-driver/0400_configmap.yaml
```diff
@@ -13,13 +13,20 @@ data:
...
 
     mkdir -p "${VALIDATIONS_DIR}"
 
     if [ ! -f /sys/module/nvidia/refcnt ]; then
       echo "NVIDIA kernel module not loaded"
       exit 1
     fi
+    
+    if [ -f "$DAEMONS_STATUS_FILE" ]; then
+      if [ "$(cat $DAEMONS_STATUS_FILE)" != "Ready" ]; then
```
rahulait: Should we echo here why its exiting?

tariq1890: sure

tariq1890: I've added the echo message. Please check now

rahulait (approved):
LGTM

rahulait:
One quick question: what will happen if the user started with newer operator and newer driver image, then driver image wrote NotReady to that status file and fails to start the daemons correctly. If the user has to switch to an older image say which is working fine, will the operator be able to proceed in that case as the status file still exists on node with NotReady state? Or, will it be able to clear it and continue.

tariq1890:
It's a valid concern. I am thinking that it may be best to clear the `.driver-daemons-status` once the driver container does a non-zero exit.

rahulait:
> It's a valid concern. I am thinking that it may be best to clear the .driver-daemons-status once the driver container does a non-zero exit.

Can we add it here? https://github.com/NVIDIA/gpu-operator/blob/144e6c4b7ca1051b989a489297f4ce921a9280ca/assets/state-driver/0500_daemonset.yaml#L153

shivamerla:
> It's a valid concern. I am thinking that it may be best to clear the `.driver-daemons-status` once the driver container does a non-zero exit.

I guess initially your PR was adding it under `/run/nvidia/driver` or the driver root itself which could have avoided this issue.

tariq1890:
> Can we add it here?

Done

PR #2750, opened 2026-08-18
add driver versions 610.57.04, 595.91.07 & 580.178.04

tariq1890:


rst0git:
This will be very helpful. We are using the 610 driver as it provides [CUDA IPC support](https://github.com/NVIDIA/cuda-checkpoint#610-features) for the Dynamo Snapshot functionality.

rahulait (approved):

PR #2753, opened 2026-08-18
chore: pin container base images to digest in Dockerfile

cdesiniotis:
This aligns with best practices and protects us against supply chain attacks.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,7 +14,7 @@
 
 ARG CUDA_SAMPLES_VERSION=13.2
 
-FROM golang:1.26.6 AS builder
+FROM golang:1.26.6@sha256:0d1d3a794be25f809dd2cb3160d8c73276c4056a9f8242a138e908ddeee7b6b6 AS builder
 
 ARG GOPROXY="https://proxy.golang.org,direct"
 ENV GOPROXY=$GOPROXY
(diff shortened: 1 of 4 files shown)


rajathagasthya (approved):

docker/Dockerfile
```diff
@@ -40,6 +40,7 @@ ARG VERSION="unknown"
 ARG GIT_COMMIT="unknown"
 RUN make cmds
 
+# We do not reuse anything from this parent image, so we do not pin to a concrete digest.
```
tariq1890: Probably just better to pin this to an immutable tag and start using that digest. Since this is not the final stage, we wouldn't have to bump it as frequently.

docker/Dockerfile
```diff
@@ -73,6 +74,8 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
 # Build a static busybox layout: one binary plus applet symlinks (sh, rm,
 # ln, sleep, cat, ...) so PATH-resolved commands in init-container wrappers
 # and lifecycle hooks keep working on the non-*-dev* distroless base.
+# We do not reuse anything from this parent image, so we do not pin to
+# a concrete digest.
```
tariq1890: When we last discussed this, we decided that it's better to use an immutable tag of the trixie image as it makes to easier to keep track of the busybox version we end up extracting

cdesiniotis: Updated

docker/Dockerfile
cdesiniotis: Updated

tariq1890 (approved):

tariq1890 (approved):

rahulait (approved):

PR #2761, opened 2026-08-19
Make Renovate update Dockerfile base image digests

rajathagasthya:
## Description

Renovate's custom manager for `docker/Dockerfile` matched only the golang version and resolved it via the `golang-version` datasource, which has no notion of image digests — PRs like #2759 bumped the tag but left the pinned `@sha256` digest unchanged. Since the digest takes precedence over the tag at pull time, docker pull silently uses the old image. The other digest-pinned base images (ubi9, debian, distroless/cc) were not managed at all.

Replace the golang-specific manager with a generic one that captures the image name, tag, and digest from every pinned `FROM` line and resolves updates via the `docker` datasource, so tag and digest are updated together. Non-version tags (`ubi9:latest`, `debian:trixie-slim`) get digest-only updates. The Go toolchain group rule is extended so golang image bumps still land in the same PR as the `versions.mk` `GOLANG_VERSION` bump.

## Checklist

## Testing

Verified the new `matchStrings` regex captures image name, tag, and digest for all five pinned `FROM` lines across `docker/Dockerfile` and `docker/Dockerfile.devel`.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -72,11 +72,9 @@
         "docker/Dockerfile.devel"
       ],
       "matchStrings": [
-        "FROM golang:(?<currentValue>\\d+\\.\\d+\\.\\d+)"
+        "FROM (?<depName>\\S+):(?<currentValue>[^\\s@]+)@(?<currentDigest>sha256:[a-f0-9]{64})"
       ],
-      "depNameTemplate": "go",
-      "datasourceTemplate": "golang-version",
-      "versioningTemplate": "semver"
+      "datasourceTemplate": "docker"
     }
   ],
   "labels": [
@@ -170,8 +168,8 @@
     {
       "description": "Group all Go toolchain version bumps",
       "matchManagers": ["custom.regex"],
-      "matchPackageNames": ["go"],
-      "matchDatasources": ["golang-version"],
+      "matchPackageNames": ["go", "golang"],
+      "matchDatasources": ["golang-version", "docker"],
       "groupName": "Go toolchain",
       "groupSlug": "go-toolchain",
       "separateMajorMinor": false,


.github/renovate.json
```diff
@@ -72,11 +72,9 @@
         "docker/Dockerfile.devel"
       ],
       "matchStrings": [
-        "FROM golang:(?<currentValue>\\d+\\.\\d+\\.\\d+)"
+        "FROM (?<depName>\\S+):(?<currentValue>[^\\s@]+)@(?<currentDigest>sha256:[a-f0-9]{64})"
       ],
-      "depNameTemplate": "go",
-      "datasourceTemplate": "golang-version",
-      "versioningTemplate": "semver"
+      "datasourceTemplate": "docker"
```
cdesiniotis, line 77: Admittedly, I am not too familiar with renovate. But it is not clear to me how an update to `docker/Dockerfile` and `docker/Dockerfile/devel` will be grouped with the update to `GOLANG_VERSION` in `versions.mk` now that there is no mention of `go` / `golang-version` here.

rajathagasthya: That grouping is now done in the other change below:

```json
"description": "Group all Go toolchain version bumps",
"matchPackageNames": ["go", "golang"],
"matchDatasources": ["golang-version", "docker"],
```

The `versions.mk` update (`go/golang-version`) and the `Dockerfile` updates (`golang/docker`) both hit that rule, so they get grouped together. The other base images (`debian`, `ubi9`, `distroless/cc`) don't match it and stay in their per-image groups.

cdesiniotis: Are we missing `depNameTemplate: golang` here?

rajathagasthya: No, `depName` is extracted by the named capture group in the regex (`FROM (?<depName>\S+)`). A hardcoded `depNameTemplate: golang` would break lookups for the non-golang base images. It's the same pattern the CSV manager above uses.

cdesiniotis: Ah, got it. Thank you for explaining!

cdesiniotis (approved):

tariq1890 (approved):

PR #2767, opened 2026-08-20
fix: disable leader election for single replica deployment

tginer:
This change avoids unnecessary leader-election overhead, removing currently lease renewal failures upon API downtime.

### Issue Observed 
In single-node clusters, the gpu-operator enters a CrashLoopBackOff cycle after every node reboot due to leader election lease renewal failures during transient API outages. This adds unnecessary recovery time before ClusterPolicy is reconciled.

The --leader-elect flag is unconditionally hardcoded in the Helm Deployment template, even though the default deployment runs a single replica. With only one replica, leader election provides zero availability benefit — there is no second replica to fail over to. Instead, every transient kube-apiserver disruption (expected during SNO MCP reconciliation) triggers a fatal leader election loss → os.Exit(1) → CrashLoopBackOff.

### Change proposal
This PR makes --leader-elect conditional on the replica count. It is only passed as an argument in case replica count is greater than 1. 

As a precedent, the NVIDIA Network Operator - Node Feature Discovery already implemented this pattern in https://github.com/Mellanox/network-operator/pull/2304 

## Checklist

## Testing

- The operator controller image was built
- API Server downtime was performed 
- Node reboot was performed

The logs demonstrate that from previous ~5 pod restarts, after the fix there is a single restart (caused by node rebot). Leader Election logs attempting to acquire the lease, failing to update lock, failing to renew the lease, leader election lost are no longer present.

### DIFF
--- a/deployments/gpu-operator/templates/operator.yaml
+++ b/deployments/gpu-operator/templates/operator.yaml
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/component: "gpu-operator"
     nvidia.com/gpu-driver-upgrade-drain.skip: "true"
 spec:
-  replicas: 1
+  replicas: {{ .Values.operator.replicas | default 1 }}
   selector:
     matchLabels:
       app.kubernetes.io/component: "gpu-operator"
@@ -39,7 +39,9 @@ spec:
         imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
         command: ["gpu-operator"]
         args:
+        {{- if gt (int (.Values.operator.replicas | default 1)) 1 }}
         - --leader-elect
+        {{- end }}
       {{- if .Values.operator.logging.develMode }}
         - --zap-devel
       {{- else }}


deployments/gpu-operator/templates/operator.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     app.kubernetes.io/component: "gpu-operator"
     nvidia.com/gpu-driver-upgrade-drain.skip: "true"
 spec:
-  replicas: 1
+  replicas: {{ .Values.operator.replicas | default 1 }}
```
rahulait, line 11: This also needs to be added to values.yaml (maybe commented) so that users can find this.

deployments/gpu-operator/templates/operator.yaml
```diff
@@ -39,7 +39,9 @@ spec:
         imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
         command: ["gpu-operator"]
         args:
+        {{- if gt (int (.Values.operator.replicas | default 1)) 1 }}
         - --leader-elect
+        {{- end }}
```
rahulait, line 44: What happens when two gpu-operator controllers come up at same time during upgrade (old controller and new controller)?

tginer: In my opinion leader election is critical under multiple replicas deployed. With a single replica and a very brief overlap period, acting without a leader seems fairly safe. 

The tests I performed resulted in:

- The overlap window was between 10 - 15 seconds
- Kubernetes reconciliation is idempotent
- Kubernetes `resourceVersion` would prevent conflicts as pod B trying to update old version would result in a re-try and read new resourceVersion


Basically the two pods were briefly trying to do the same idempotent work and the old pod was terminated in a 10-15 second-window.

I replicated the idea implemented by NVIDIA in https://github.com/Mellanox/network-operator/pull/2304 but the network operator AI agent bot also raised your concern in my identical PR in the NNO, see https://github.com/Mellanox/network-operator/pull/3056

rajathagasthya: Concurrent reconciles by different versions of the operator are almost never a good thing, even if they are idempotent in theory. It can be a source of unexpected bugs and we are sensitive to unnecessary daemonset rollouts (especially triggering unnecessary driver upgrades). So this change does not look safe to me. 

Instead of disabling leader election, how about setting a larger `--leader-lease-renew-deadline=60s` value? That will make the operator tolerate any API outages up to 60s while still holding the lease. That way the operator doesn't enter a crash loop if it can't renew the lease in 10s (controller-runtime default) and doesn't delay ClusterPolicy processing due to exponential backoff. The OLM bundle already does this: https://github.com/NVIDIA/gpu-operator/blob/10ee5b3638b89e11e949412aafa5ba99279c3721/bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml#L1081-L1082

rahulait: @tginer let us know if this suggestion fixes the issue you are having.

rahulait: @tginer gentle reminder: let us know if the above suggestion works for you.

tginer: hi @rahulait  I can confirm that after multiple rests, the 60s renew deadline did make a difference and avoid the CrashLoopBackOff from the controller. Thank you very much for your inputs

rahulait:
As discussed, the workaround works without disabling leader election. Closing this PR for now, please feel free to re-open if you still see issues with it.

deployments/gpu-operator/templates/operator.yaml
rajathagasthya: Awesome, thanks for trying out @tginer!

@rahulait Should we match OLM bundle's `--leader-lease-renew-deadline=60s` in helm too?

rahulait: Yes, I think that makes sense. I'll open a PR for the same.

PR #2779, opened 2026-08-21
Structured YAML unmarshalling based manifest kind detection

kvalliyurnatt:
Replace Regex based Kubernetes manifest kind detection with structured YAML unmarshalling. This feels a little more safe than regex based yaml parsing. Also added a couple of unit tests to test this.


tariq1890 (approved):
Thank you @kvalliyurnatt!

PR #2784, opened 2026-08-21
Mount host /lib/modules for precompiled drivers on SUSE

### LINKED ISSUE #2776: [Bug]: Unreachable code prevents lib-modules volume mount when using precompiled drivers.
**Description**
When deploying an NVIDIA driver Custom Resource with `usePrecompiledDrivers: true`, the `lib-modules` volume is not being mounted inside the driver pod.

This happens because of conflicting conditional checks in `internal/state/driver_volumes.go` within the `getDriverAdditionalConfigs` function, which creates an unreachable execution path.

**Code Reference**

* **Repository:** NVIDIA/gpu-operator
* **File:** `internal/state/driver_volumes.go` (commit `de69cf78225fa0b3f7bea871e91783630fa104f5`)
* **Function:** `getDriverAdditionalConfigs` (line 140)

**Details**
At line 145, the code checks if precompiled drivers are *disabled* before entering the block:

```go
if !cr.Spec.UsePrecompiledDrivers() {
    // ...

```

However, inside this exact same block at line 226, there is a contradictory condition checking if precompiled drivers are *enabled*:

```go
if cr.Spec.UsePrecompiledDrivers() {
    // ...

```

Since the outer `if` statement enforces `UsePrecompiledDrivers()` as `false`, the inner condition at line 226 evaluates to `false` in every scenario. This results in dead code, meaning the volume mount below is never executed or appended to the pod:

```go
libModulesVolMount := corev1.VolumeMount{
	Name:      "lib-modules",
	MountPath: "/run/host/lib/modules",
	ReadOnly:  true,
}

```

**Environment**

* **GPU Operator Version:** v24.9.2

**Expected Behavior**
When `usePrecompiledDrivers: true` is set, the `lib-modules` volume mount should be successfully evaluated and mounted to the pod. The logic should be refactored so that the `lib-modules` mount configuration is not trapped inside a contradictory `!cr.Spec.UsePrecompiledDrivers()` block.

**Actual Behavior**
The logic silently skips the volume mount configuration, resulting in a pod deployed without the necessary `/run/host/lib/modules` mount.

abrarshivani:
## Description

Fixes #2776.

`getDriverAdditionalConfigs` builds the `lib-modules` volume and mount inside the
`if !cr.Spec.UsePrecompiledDrivers()` branch, and the block itself is guarded by
`if cr.Spec.UsePrecompiledDrivers() && (pool.osRelease == "sles" || pool.osRelease == "sl-micro")`.
The inner condition can never hold in that branch, so an NVIDIADriver with
`usePrecompiled: true` on sles or sl-micro renders a driver pod without
`/run/host/lib/modules`.

Rather than repeat the predicate, the mount now hangs off the non-precompiled
branch as an `else`. The two paths are mutually exclusive, and saying so once
gives precompiled-only logic an obvious home, which is what went wrong here in
the first place. The comment above the mount goes with it: the `else` already
says precompiled, the guard already says SUSE, and the volume literals already
say host module tree.

This only affects the NVIDIADriver path. The equivalent code in
`controllers/object_controls.go` sits before the early return for precompiled
drivers and works as intended. That is also why it went unnoticed, since the
tests that came in with a7323f51 only covered the ClusterPolicy path.

One note for triage: the issue lists v24.9.2 under Environment, but v24.9.2 has
no `lib-modules` handling in either driver path, so the unreachable code
described in the report cannot exist there. The line numbers quoted in the issue
match current main. A v24.9.2 user on precompiled SUSE does see the same missing
mount, but because the feature had not landed yet rather than because of this
bug.

## Checklist

## Testing

Added `TestDriverPrecompiledLibModules`, a table-driven test covering sles and
sl-micro with precompiled drivers, plus sles without them and ubuntu with them
as negative cases. Each case renders the DaemonSet and checks the pod spec for
the `lib-modules` hostPath volume and `nvidia-driver-ctr` for the mount, so it
covers both the helper and the template threading `AdditionalConfigs` into the
pod.

The two SUSE cases fail against the current code with "expected a lib-modules
volume on the driver pod spec" and pass with the fix. Removing the
`AdditionalConfigs` mounts from the driver container in
`manifests/state-driver/0500_daemonset.yaml` also fails the test, which the
helper-level assertions alone would not have caught.

`go test ./internal/... ./controllers/...` passes, along with `go vet`, gofmt
and `-race`. Lint and the generated-asset check were run with `GOOS=linux`,
since the vendored `filepath-securejoin/pathrs-lite` is linux-only and breaks
the typecheck of `cmd/nvidia-validator` on macOS. Regenerating the CRDs produces
no diff.

### DIFF
--- a/internal/state/driver_volumes.go
+++ b/internal/state/driver_volumes.go
@@ -220,29 +220,27 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
 				additionalCfgs.Volumes = append(additionalCfgs.Volumes, subscriptionVol)
 			}
 		}
+	}
 
-		// Mount /lib/modules for precompiled drivers on SUSE distributions.
-		// Those containers need access to host /lib/modules at runtime.
-		if cr.Spec.UsePrecompiledDrivers() && (pool.osRelease == "sles" || pool.osRelease == "sl-micro") {
-			logger.Info("Mounting /lib/modules into the driver container")
-			libModulesVolMount := corev1.VolumeMount{
-				Name:      "lib-modules",
-				MountPath: "/run/host/lib/modules",
-				ReadOnly:  true,
-			}
-			additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, libModulesVolMount)
-
-			libModulesVol := corev1.Volume{
-				Name: "lib-modules",
-				VolumeSource: corev1.VolumeSource{
-					HostPath: &corev1.HostPathVolumeSource{
-						Path: "/lib/modules",
-						Type: ptr.To(corev1.HostPathDirectory),
-					},
-				},
-			}
-			additionalCfgs.Volumes = append(additionalCfgs.Volumes, libModulesVol)
+	if cr.Spec.UsePrecompiledDrivers() && (pool.osRelease == "sles" || pool.osRelease == "sl-micro") {
+		logger.Info("Mounting /lib/modules into the driver pod")
+		libModulesVolMount := corev1.VolumeMount{
+			Name:      "lib-modules",
+			MountPath: "/run/host/lib/modules",
+			ReadOnly:  true,
 		}
+		additionalCfgs.VolumeMounts = append(additionalCfgs.VolumeMounts, libModulesVolMount)
+
+		libModulesVol := corev1.Volume{
+			Name: "lib-modules",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/lib/modules",
+					Type: ptr.To(corev1.HostPathDirectory),
+				},
+			},
+		}
+		additionalCfgs.Volumes = append(additionalCfgs.Volumes, libModulesVol)
 	}
 
 	// mount any custom kernel module configuration parameters at /drivers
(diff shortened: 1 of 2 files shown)


internal/state/driver_volumes.go
```diff
@@ -220,29 +220,25 @@ func (s *stateDriver) getDriverAdditionalConfigs(ctx context.Context, cr *v1alph
 				additionalCfgs.Volumes = append(additionalCfgs.Volumes, subscriptionVol)
 			}
 		}
+	} else if pool.osRelease == "sles" || pool.osRelease == "sl-micro" {
```
tariq1890: This is technically correct, but for readability, let us add this to its own `if` block as opposed to an `else if`. The new `if` will still be outside the `if !cr.Spec.UsePrecompiledDrivers() {` block.

tariq1890:
Thanks for working on this @abrarshivani !

This will need to be backported to `release-26.7`

abrarshivani:
@tariq1890 Sure, will backport this change once this PR is checked in.

internal/state/driver_volumes.go
abrarshivani: Updated

tariq1890 (approved):

PR #2794, opened 2026-08-24
Link third-party notices to their licence files, and add a per-release notices document

abrarshivani:
Adds licence links to the notices document on `main`, and a manually
triggered workflow that builds a richer per-release document.

## Problem

- The index's third column repeated the module path and carried no information; nothing pointed at a licence file.
- Secondary licence files such as `PATENTS` and `NOTICE` were not reproduced alongside the licence they accompany, because `go-licenses save` copies only the file it classified.
- Two identifiers were wrong as a result: `go.yaml.in/yaml/v2` and `v3` are both `Apache-2.0 AND MIT`.
- Nothing outside the Go module graph was attributed, including the GPL-2.0 busybox binary the Dockerfile copies in.

## Changes

**Repo document** (`THIRD_PARTY_NOTICES.md`, gated by CI)
- Licence text read from `vendor/`, which is what we actually ship.
- `Location` links to each licence file in this repository.
- Secondary files (`PATENTS`, `NOTICE`, `AUTHORS`, `LICENSE.*`) reproduced alongside the main licence.
- `tools/license-overrides.tsv` carries curated SPDX identifiers for files bundling two licences, joined with `AND`.

**Release artifact** (workflow only, never committed)
- States each dependency's version, with links pinned to the ref the release was cut from.
- Reports the base image and its version, read from the final `FROM`.
- Lists what the Dockerfile adds on top — currently busybox — with licence, version and notices link. NVIDIA's own components are first-party and out of scope.

**Workflow**
- Manual trigger only; never runs as part of a release.
- Publishes nothing: the document is a workflow artifact, so the job stays read-only.
- Ref must resolve to a tag or the checked-out commit.

## Guards

The generator fails rather than emitting a document that is quietly wrong:

- a package with no licence file, or an unresolvable licence location
- an `Unknown` or empty licence identifier
- a stale override row, or one with an empty identifier
- a Dockerfile `COPY` into the final image with no catalogue row
- a base image that is not a digest-pinned literal, or not NVIDIA distroless
- a hand-recorded version whose source image is no longer built from
- a malformed or duplicated catalogue row

## Verification

- packages: go-licenses reports 123, the document lists 123
- links: all 166 in the repo document, all 167 in a release document, HTTP 200
- 53 assertions; shellcheck and actionlint clean; `make check-third-party-notices` green
- output byte-identical across runs

## Not included

Base-layer packages of the image itself. The base image is named with its version and linked to NVIDIA's per-version OSS source index, which is not a substitute for that image's own licence texts.

The last commit adds a generated sample for review and should be dropped before merge.

---

<a href="https://nvidia-cloud.devinenterprise.com/review/nvidia/gpu-operator/pull/2794" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


.github/workflows/third-party-notices-release.yaml
```diff
@@ -0,0 +1,120 @@
...
+          INPUT_REF: ${{ inputs.ref }}
+        run: |
+          if [[ "${INPUT_REF}" =~ ^[0-9a-fA-F]{40}$ ]]; then
+            if [[ "$(git rev-parse HEAD)" != "$(echo "${INPUT_REF}" | tr '[:upper:]' '[:lower:]')" ]]; then
+              echo "::error::${INPUT_REF} is not the checked-out commit."
+              exit 1
+            fi
+            exit 0
+          fi
+          git rev-parse --verify --quiet "refs/tags/${INPUT_REF}^{commit}" >/dev/null \
```
kvalliyurnatt: should we also verify the tag's commit matches head 
```
[[ "$(git rev-parse HEAD)" == \
   "$(git rev-parse "refs/tags/${INPUT_REF}^{commit}")" ]]
```

tools/generate-third-party-notices.sh
```diff
@@ -103,218 +103,501 @@ verify_platform_matrix() {
...
+        [[ -z "${source_path}" ]] && continue
+        expanded="$(expand_dockerfile_args "${source_path}")" || exit 1
+        bundled_component_row "${source_path}" >/dev/null \
+            || bundled_component_row "${expanded}" >/dev/null \
+            || die "${DOCKERFILE} copies ${source_path} into the released image, and ${BUNDLED_COMPONENTS} has no row for it." \
+                   "Add one saying what it is and under what terms it is redistributed."
+    done < <(final_stage_copies "${base_repository}")
+
+    local component version license_file
+    while IFS=$'\t' read -r component version _ license_file _ _ _; do
```
kvalliyurnatt: should we check the other fields as well here
```
while IFS=$'\t' read -r component version license license_file notices_url _ _; do
    [[ "${version}" != "-" ]] || die "${component} has no version."
    [[ "${license}" != "-" ]] || die "${component} has no license identifier."
    [[ "${license_file}" != "-" ]] || die "${component} has no license file."
    [[ "${notices_url}" != "-" ]] || die "${component} has no notices URL."
    [[ -f "${LICENSE_TEXTS_DIR}/${license_file}" ]] \
        || die "License file for ${component} does not exist."
done < "${BUNDLED_ROWS}"
```

why are we skipping them currently ?

abrarshivani: Good catch! I've added checks for license and notices_url as well.

.github/workflows/third-party-notices-release.yaml
abrarshivani: Yes, it is good to have this check. Updated.

tools/generate-third-party-notices.sh
```diff
@@ -103,218 +103,513 @@ verify_platform_matrix() {
...
+}
 
-    while IFS=, read -r pkg _ license module; do
-        [[ -z "${pkg}" ]] && continue
+check_version_provenance() {
+    local source_path component provenance_digest
+    while IFS=$'\t' read -r source_path component _ _ _ _ _ provenance_digest _; do
+        case "${source_path}" in ''|'#'*) continue ;; esac
+        [[ "${provenance_digest}" == "-" ]] && continue
+        LC_ALL=C grep -qF "${provenance_digest}" "${DOCKERFILE}" \
```
kvalliyurnatt, line 451: I think this is a little fragile, we check if the provenance_digest is present somewhere in the Dockerfile. I think for a component we should verify what the source stage is and what the image digest used there is and whether that matches the provenance_digest

example if we expect BusyBox to have A as the digest but we have this in the Dockerfile

```
FROM debian@sha256:A AS old-shell
FROM alpine@sha256:B AS new-shell
COPY --from=new-shell /busybox /busybox
```

This will pass currently even though busybox is not being built from A, I think this is not an issue for any of the Dockerfiles currently, but this might break in the future.

tools/bundled-components.tsv
```diff
@@ -0,0 +1,19 @@
+# disposition: project | first-party | third-party. Only third-party is listed;
+# every copied path needs a row so nothing reaches the image unattributed.
+# Write "-" for an absent value; an empty field is rejected.
+# source_path	component	disposition	version	license_identifier	license_file	notices_url	provenance_digest	attribution_note
+/busybox	busybox	third-party	1:1.37.0-6	GPL-2.0-only	busybox/LICENSE	https://sources.debian.org/data/main/b/busybox/1%3A1.37.0-6/debian/copyright	sha256:3a39a0592364683e6bab97937b72cad5a8fa6dcbbee90edb3bb48c7f8e94f258	apt resolves the package at build time, so the version is the one Debian trixie ships for this builder digest
```
rajathagasthya: The actual version of busybox is `1:1.37.0-6+b9`. Should we include that or is this good enough?

.github/workflows/third-party-notices-release.yaml
```diff
@@ -0,0 +1,212 @@
...
+          } >> "${GITHUB_OUTPUT}"
+
+      - name: Upload the notices artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: ${{ steps.generate.outputs.artifact_name }}
+          path: ${{ steps.generate.outputs.artifact_path }}
+          if-no-files-found: error
+
+      - name: Summarise
```
rahulait:
```suggestion
      - name: Summarize
```

.github/workflows/third-party-notices-release.yaml
```diff
@@ -0,0 +1,212 @@
...
+
+      - name: Attach the notices to the GitHub Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          INPUT_REF: ${{ inputs.ref }}
+          ARTIFACT_PATH: ${{ runner.temp }}/notices/${{ needs.generate.outputs.artifact_name }}
+        run: gh release upload "${INPUT_REF}" "${ARTIFACT_PATH}"
+
+      - name: Summarise
```
rahulait:
```suggestion
      - name: Summarize
```

.github/workflows/third-party-notices-release.yaml
```diff
@@ -0,0 +1,212 @@
...
+        env:
+          ARTIFACT_NAME: ${{ steps.generate.outputs.artifact_name }}
+          RESOLVED_SHA: ${{ steps.resolve.outputs.resolved_sha }}
+        run: |
+          {
+            echo "### Third-party notices generated"
+            echo
+            echo "- Artifact: \`${ARTIFACT_NAME}\`"
+            echo "- Commit: \`${RESOLVED_SHA}\`"
```
rajathagasthya, line 152: nit: let's also add `Tag: <tag>` here in addition to commit SHA.

rajathagasthya (approved):
Didn't really review the shell scripts too closely, but the generated results look good.

rahulait (approved):
Other than few nits, LGTM. Didn't run or test the scripts, but assuming its tested, this looks good to me.

tools/generate-third-party-notices.sh
```diff
@@ -103,218 +103,549 @@ verify_platform_matrix() {
...
+    done
+    printf '%s' "${text}"
+}
+
+final_stage_copies() {
+    local base_repository="$1"
+    [[ -n "${base_repository}" ]] \
+        || die "final_stage_copies needs the repository of the image the final stage is built from."
+
+    LC_ALL=C awk -v base_repository="${base_repository}" '
```
kvalliyurnatt: Dockerfile commands are case insensitive, but here we are loking for specific commands in a specific case, I think we should make this case insensitive as well
Also we have not considered ADD here ?

tools/bundled-components.tsv
abrarshivani: Good catch, will update this.

tools/generate-third-party-notices.sh
abrarshivani: Updated

tools/bundled-components.tsv
```diff
@@ -0,0 +1,19 @@
...
+/workspace/cleanup-gpuclusters	cleanup-gpuclusters	project	-	-	-	-	-	-
+/workspace/nvidia-validator	nvidia-validator	project	-	-	-	-	-	-
+assets	assets	project	-	-	-	-	-	-
+manifests	manifests	project	-	-	-	-	-	-
+validator/manifests	validator-manifests	project	-	-	-	-	-	-
+deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml	clusterpolicies-crd	project	-	-	-	-	-	-
+deployments/gpu-operator/crds/nvidia.com_nvidiadrivers.yaml	nvidiadrivers-crd	project	-	-	-	-	-	-
+deployments/gpu-operator/crds/nvidia.com_gpuclusters.yaml	gpuclusters-crd	project	-	-	-	-	-	-
+LICENSE	gpu-operator-license	project	-	-	-	-	-	-
+deployments/gpu-operator/charts/node-feature-discovery/crds/nfd-api-crds.yaml	node-feature-discovery CRDs	third-party	v0.19.0	Apache-2.0	node-feature-discovery/LICENSE	https://github.com/kubernetes-sigs/node-feature-discovery/tree/v0.19.0	-	generated CRD schema from the vendored Helm chart; the NFD programs themselves are not shipped in this image
```
kvalliyurnatt: How would this work when the NFD version gets updated in the helm chart ? do we have to ensure we update here as well ?

gpu-operator-26.7.0-THIRD_PARTY_NOTICES.md
```diff
@@ -10,153 +10,175 @@ resolved as the union across every released image platform. The `gpu-operator`,
...
+Software the Dockerfile adds on top of the base image is listed under Bundled
+Components below, with its license and corresponding source. NVIDIA's own
+components are not third party to NVIDIA and are out of scope here.
+
+The base image's own sources are published per version. A `-dev` variant is
+built from the sources published under the corresponding release version.
+
+| Image | Version | Role | Notices and source |
+|-------|---------|------|--------------------|
+| `nvcr.io/nvidia/distroless/cc` | `v4.1.4` | final runtime base | [NVIDIA Distroless OSS source index](https://developer.download.nvidia.com/distroless-oss/cc/v4.1.4/index.html) |
```
rahulait: This doesn't look right. This document states its for v26.7.0, however, v26.7.0 release had older version of distroless/cc (v4.1.1 and not v4.1.4)

tariq1890:
(Let's not forget to squash the commit history)

I'll review the changes in the afternoon.

tools/generate-third-party-notices.sh
abrarshivani: Makes sense. I'll update this. Yes, we should consider the `ADD` case here.

tools/bundled-components.tsv
abrarshivani: Yes, we need to update the `bundled-components.tsv` as well.

.github/workflows/third-party-notices-release.yaml
abrarshivani: Will add this.

tools/bundled-components.tsv
kvalliyurnatt: I am just a little worried about a possible drift here, is there any way to add a check or something to make sure the versions line up with what we have in the helm chart ?

gpu-operator-26.7.0-THIRD_PARTY_NOTICES.md
abrarshivani: This was initially generated for v26.7.0 but later was updated to this branch. Will fix this.

tools/bundled-components.tsv
abrarshivani: I'll add a check for version drift.

.github/workflows/third-party-notices-release.yaml
```diff
@@ -0,0 +1,216 @@
...
+      - name: Summarize
+        env:
+          ARTIFACT_NAME: ${{ steps.generate.outputs.artifact_name }}
+          INPUT_REF: ${{ inputs.ref }}
+          RESOLVED_SHA: ${{ steps.resolve.outputs.resolved_sha }}
+        run: |
+          {
+            echo "### Third-party notices generated"
+            echo
+            echo "- Artifact: \`${ARTIFACT_NAME}\`"
```
tariq1890, line 150: Why use backticks here?

docker/Dockerfile
```diff
@@ -78,7 +78,7 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
 # and lifecycle hooks keep working on the non-*-dev* distroless base.
 FROM debian:trixie-slim@sha256:3a39a0592364683e6bab97937b72cad5a8fa6dcbbee90edb3bb48c7f8e94f258 AS shell
 RUN apt-get update \
- && apt-get install -y --no-install-recommends busybox-static \
+ && apt-get install -y --no-install-recommends busybox-static=1:1.37.0-6+b9 \
```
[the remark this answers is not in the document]
tariq1890: @abrarshivani  This comment is correct. Usually, it's a good idea to pin the package version, but the debian trixie repositories don't freeze the package repositories. This means that once a new busybox version is out, the old version of the busybox is no longer available thus causing builds to break.

@rajathagasthya  FYI, this was the reason I had looked into using the official busybox images to grab the busybox binary instead.

tools/licenses/busybox/LICENSE
tariq1890, line 1: How is this file generated?

docker/Dockerfile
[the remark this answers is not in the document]
abrarshivani: Agreed. I was actually thinking along the same lines that is using the official busybox image as the source for the busybox binary. But that felt out of scope for this PR, so I added this version pin as a temporary measure. I can revert the pin for now if needed, since I'm planning to handle the busybox source properly in a follow up PR anyway.

.github/workflows/third-party-notices-release.yaml
abrarshivani: Just to format the artifact name as inline code in the github actions job summary. It is not necessary though, so I can remove the backticks.

tools/licenses/busybox/LICENSE
abrarshivani: This file isn't generated. It is a checked in copy of the upstream busybox license file, which the TPN generator references for the busybox entry.

docker/Dockerfile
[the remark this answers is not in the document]
rajathagasthya: Let's revert the version pin. There's unfortunately no way to make it work due to how to Debian package repo behavior. 

We've tried using official busybox binary from its official image, but it that binary is dynamically linked compared to the static one we install currently. Details here: https://github.com/NVIDIA/gpu-operator/pull/2853#discussion_r3930322386

PR #2804, opened 2026-08-25
Add CodeRabbit configuration to scope automated PR reviews

karthikvetrivel:
## Description

I added a `.coderabbit.yaml` so that CodeRabbit reviews this repo usefully.

Two things about this repo would make a default install noisy.
-  Roughly half of all pull requests are machine generated by Dependabot, Renovate, and the cherry-pick workflow. 
-  Vendored and generated files also make up about 90% of tracked files.

The config addresses both:

- It uses the `quiet` profile, so only critical and major findings are posted inline.
- It skips pull requests authored by `dependabot[bot]` and `github-actions[bot]`. A backport only replays commits that were already reviewed on main.
- It filters `vendor/`, generated deepcopy and clientset code, the CRD copies that `make sync-crds` produces, and the intentionally malformed fixtures under `testdata/`.
- It disables `golangci-lint`, which CI already runs with our own config, and `yamllint`, which cannot parse our Helm and Go templates.

## Checklist

## Testing

This change adds no Go code, so there are no new code paths to cover and no Go artifacts to regenerate.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,177 @@
+# SPDX-FileCopyrightText: Copyright (c) NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+
+tone_instructions: "Be direct, technical and brief. No praise or emoji. Skip style and formatting, which CI already gates. Without evidence in the diff, ask a question rather than assert a defect. Name the concrete failure each finding prevents."
+
+reviews:
+  # Only critical and major findings are posted inline. Everything else is
+  # collapsed into the summary instead of becoming its own review thread.
+  profile: quiet
+
+  # The pull request template already asks for a description and a testing note,
+  # so a generated walkthrough on top of it is noise.
+  high_level_summary: false
+  changed_files_summary: false
+  sequence_diagrams: false
+  estimate_code_review_effort: false
+  assess_linked_issues: false
+  related_issues: false
+  related_prs: false
+  review_status: false
+  in_progress_fortune: false
+  # This strips the "Prompt for AI Agents" block appended to every inline comment.
+  enable_prompt_for_ai_agents: false
+
+  # CI owns the merge gate, so we do not need another check run or commit status.
+  # Both keys are required because commit_status only applies while
+  # review_progress is disabled.
+  review_progress: false
+  commit_status: false
+
+  # CODEOWNERS already assigns reviewers, and cherry-pick/* labels drive real
+  # backport automation, so CodeRabbit must not guess at either.
+  suggested_reviewers: false
+  suggested_labels: false
+
+  auto_review:
+    # Each pull request is reviewed once. Pushes do not re-trigger a review, so
+    # request another pass with "@coderabbitai review" after addressing feedback.
+    auto_incremental_review: false
+    # These are regular expressions rather than globs. ".*" reviews pull requests
+    # into any base branch, which covers main, release-<calver> hotfixes, and
+    # stacked work merging one feature branch into another.
+    base_branches:
+      - ".*"
+    # About half of all pull requests are machine-generated. Renovate and the
+    # cherry-pick workflow both author as github-actions[bot], and a backport
+    # only replays commits that were already reviewed on main.
+    ignore_usernames:
+      - "dependabot[bot]"
+      - "github-actions[bot]"
+    ignore_title_keywords:
+      - "WIP"
+      
+  path_filters:
+    - "!vendor/**"
+    - "!**/zz_generated.*.go"
+    - "!api/versioned/**"
+    - "!config/crd/bases/**"
+    - "!config/rbac/role.yaml"
+    # make sync-crds copies config/crd/bases verbatim into both of these.
+    - "!deployments/gpu-operator/crds/**"
+    - "!bundle/manifests/nvidia.com_*.yaml"
+    - "!bundle/manifests/resource.nvidia.com_*.yaml"
+    # This is the upstream node-feature-discovery subchart, vendored as-is.
+    - "!deployments/gpu-operator/charts/**"
+    # These are golden snapshots and deliberately malformed parser fixtures.
+    - "!**/testdata/**"
+    - "!go.sum"
+    - "!deployments/gpu-operator/Chart.lock"
+    # These are the packaged chart and generated index published to the gh-pages
+    # branch. 
+    - "!stable/**"
+
+  path_instructions:
+    - path: "{controllers,internal,cmd}/**/*.go"
+      instructions: |
+        Reconciliation code for a shipped operator. Below is a prioritized
+        list of failures to look out for:
+        - Flag reconcile logic that is not safe to run twice.
+        - Flag an error that is logged and then returned as success. That
+          suppresses the retry, so the object keeps its broken state until the
+          next watch event or resync.
+        - Flag an optional CRD field dereferenced without a nil check. These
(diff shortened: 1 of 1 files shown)


kvalliyurnatt (approved):

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+
+tone_instructions: "Be direct, technical and brief. No praise or emoji. Skip style and formatting, which CI already gates. Without evidence in the diff, ask a question rather than assert a defect."
```
rajathagasthya:
```suggestion
tone_instructions: "Be direct, technical and brief. No praise or emoji. Skip style and formatting, which CI already gates. Without evidence in the diff, ask a question rather than assert a defect. Name the concrete failure each finding prevents."
```

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
...
+    title:
+      mode: "off"
+    description:
+      mode: "off"
+    issue_assessment:
+      mode: "off"
+
+knowledge_base:
+  code_guidelines:
+    # This replaces the default patterns such as AGENTS.md and CLAUDE.md, none of
```
rajathagasthya: non-blocking: maybe we should consider having an `AGENTS.md` and pointing this config to it. I think all the "path instructions" could go there instead and it can be read by any agentic coding tool.

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
...
+        added or renamed without a matching template reference (or the reverse),
+        and any default whose change alters behaviour for existing installs on
+        helm upgrade. Image repository/version pairs must stay consistent with
+        bundle/manifests/gpu-operator-certified.clusterserviceversion.yaml.
+
+    - path: "**/*_test.go"
+      instructions: |
+        Flag assertions that would still pass if the behaviour under test were
+        broken, and new operand or CRD behaviour that ships with no regression
+        test. Do not ask for tests covering generated or vendored code.
```
rajathagasthya:
```suggestion
        Flag assertions that would still pass if the behavior under test were
        broken, and new operand or CRD behaviour that ships with no regression
        test. Flag tests that mirror the implementation instead of checking
        behavior — the expected value is computed the same way as the code
        under test, or the test only checks that a mock was called. These break
        on refactors without catching bugs.
        Do not ask for tests covering generated or vendored code.
```

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
...
+
+    - path: "api/**/*.go"
+      instructions: |
+        ClusterPolicy, GPUCluster and NVIDIADriver are released CRDs that users
+        have already applied. Treat as major any removed or renamed field,
+        narrowed +kubebuilder:validation, changed +kubebuilder:default, or
+        changed Go type: each breaks existing custom resources on upgrade.
+        New fields must be optional and carry omitempty.
+        Editing these types requires regenerating the deepcopy and CRD assets,
+        so say so if the PR does not also update them.
```
rajathagasthya, line 104:
```suggestion
        ClusterPolicy, GPUCluster and NVIDIADriver are released CRDs that users
        have already applied. Treat as major any removed or renamed field or
        json tag, a field made required, narrowed +kubebuilder:validation,
        changed +kubebuilder:default, or changed Go type: each breaks existing
        custom resources on upgrade.
        New fields must be optional and carry omitempty.
        Every new field is permanent API surface that must be supported across
        upgrades. Ask for justification when a new field duplicates an existing
        knob, could be derived from existing fields, or configures an
        implementation detail that could live in a ConfigMap or annotation
        instead. Spec holds user intent; status holds observed state; reject
        fields that blur that line.
        Editing these types requires regenerating the deepcopy and CRD assets,
        so say so if the PR does not also update them.
```

rajathagasthya (commented):
Added some suggestions.

rahulait (approved):

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
...
+tone_instructions: "Be direct, technical and brief. No praise or emoji. Skip style and formatting, which CI already gates. Without evidence in the diff, ask a question rather than assert a defect."
+
+reviews:
+  # Only critical and major findings are posted inline. Everything else is
+  # collapsed into the summary instead of becoming its own review thread.
+  profile: quiet
+
+  # The pull request template already asks for a description and a testing note,
+  # so a generated walkthrough on top of it is noise.
+  high_level_summary: false
```
abrarshivani: Can we enable this for large PRs?

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
...
+  changed_files_summary: false
```
abrarshivani: Can we enable this for large PRs?

.coderabbit.yaml
```diff
@@ -0,0 +1,165 @@
...
+  sequence_diagrams: false
```
abrarshivani: Can we enable this for large PRs?

.coderabbit.yaml
```diff
@@ -0,0 +1,176 @@
...
+    base_branches:
+      - ".*"
+    # About half of all pull requests are machine-generated. Renovate and the
+    # cherry-pick workflow both author as github-actions[bot], and a backport
+    # only replays commits that were already reviewed on main.
+    ignore_usernames:
+      - "dependabot[bot]"
+      - "github-actions[bot]"
+    ignore_title_keywords:
+      - "WIP"
```
cdesiniotis, line 54: Question -- does coderabbit also skip reviewing draft PRs? Or do we need to explicitly configure that behavior?

.coderabbit.yaml
```diff
@@ -0,0 +1,176 @@
...
+    - "!bundle/manifests/resource.nvidia.com_*.yaml"
+    # This is the upstream node-feature-discovery subchart, vendored as-is.
+    - "!deployments/gpu-operator/charts/**"
+    # These are golden snapshots and deliberately malformed parser fixtures.
+    - "!**/testdata/**"
+    - "!go.sum"
+    - "!deployments/gpu-operator/Chart.lock"
+    # These are the packaged chart and generated index published to the gh-pages
+    # branch. 
+    - "!stable/**"
```
cdesiniotis, line 74: Not a strong opinion from me, but I slightly prefer disabling reviews where the base branch is `gh-pages` instead of having this `path_filter` + comment explaining that this is specific to the gh-pages branch.

.coderabbit.yaml
```diff
@@ -0,0 +1,176 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cdesiniotis: nit: let's remove the year from the license header

.coderabbit.yaml
karthikvetrivel: Yep, they are skipped.

.coderabbit.yaml
karthikvetrivel: From everything I've read, this is hard to do because `base_branches` is a positive allowlist with no negation so this would be hard to implement...

.coderabbit.yaml
karthikvetrivel: Done.

.coderabbit.yaml
cdesiniotis: Ack, just a suggestion. Will resolve since it is not straightforward to implement.

cdesiniotis (approved):

PR #2805, opened 2026-08-25
Add pprof endpoint flag to the operator

rajathagasthya:
## Description

Add a `--pprof-bind-address` flag wired to controller-runtime's built-in
pprof server (`ctrl.Options.PprofBindAddress`), for profiling the
operator's memory and CPU at large node counts. The endpoint is disabled
by default and only serves when an address (e.g. `":6060"`) is given.

The address is exposed as the `operator.pprof.bindAddress` chart value,
passed to the operator only when set.

## Checklist

## Testing

`go build ./cmd/gpu-operator`; `helm template` with and without
`operator.pprof.bindAddress` set (flag rendered only when set); `helm lint`.

### DIFF
--- a/cmd/gpu-operator/main.go
+++ b/cmd/gpu-operator/main.go
@@ -78,9 +78,12 @@ func main() {
 	var enableLeaderElection bool
 	var leaderElectionNamespace string
 	var probeAddr string
+	var pprofAddr string
 	var renewDeadline time.Duration
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&pprofAddr, "pprof-bind-address", "",
+		"The address the pprof endpoint binds to (e.g. \":6060\"). Disabled when empty.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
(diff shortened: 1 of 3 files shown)


rahulait (approved):

kvalliyurnatt (approved):

cmd/gpu-operator/main.go
```diff
@@ -78,9 +78,12 @@ func main() {
 	var enableLeaderElection bool
 	var leaderElectionNamespace string
 	var probeAddr string
+	var pprofAddr string
 	var renewDeadline time.Duration
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&pprofAddr, "pprof-bind-address", "",
+		"The address the pprof endpoint binds to (e.g. \":6060\"). Disabled when empty.")
```
tariq1890, line 86: Why disabled when empty? Why not just expose the pprof endpoint on port 6060 by default? It's not too costly too host a pprof endpoint, yes?

rajathagasthya: No particular reason really. I was just partially matching controller-runtime behavior of `""` to disable serving. We can enable by default, although [the doc](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#Options) says it may contain sensitive information. 

```go
// PprofBindAddress is the TCP address that the controller should bind 
// for serving 
// It can be set to "" or "0" to disable the pprof 
// Since pprof may contain sensitive information, make sure to protect 
// before exposing it to public.
PprofBindAddress string
```

tariq1890: Got it makes sense. If this is aligned with controller-runtime, no objections from me. Thanks!

PR #2808, opened 2026-08-26
fix(ci): use least-privilege token permissions in GHA workflows

cdesiniotis:
## Description

OpenSSF Scorecard's Token-Permissions check flagged several reusable GHA workflows for not adhering to the principle of least-privilege for their GITHUB_TOKEN usage. This commit makes it so that the permissions definitions in each workflow's yaml file are set as read-only at the top level and the required write permissions are declared at the run-level.

Reference https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

Changes in this PR were made with the assistance of Claude Code.
## Checklist

## Testing

I ran https://github.com/ossf/scorecard with and without this change. With this change, our score for the `Token-Permissions` category jumps from 0/10 to 10/10

### DIFF
--- a/.github/workflows/image-builds.yaml
+++ b/.github/workflows/image-builds.yaml
@@ -30,6 +30,9 @@ on:
         required: true
         type: string
 
+permissions:
+  contents: read
+
 jobs:
   variables:
     runs-on: ubuntu-latest
@@ -147,6 +150,9 @@ jobs:
     needs: [variables, build-gpu-operator-arm64, build-gpu-operator-amd64]
     runs-on: ubuntu-latest
     timeout-minutes: 15
+    permissions:
+      contents: read
+      packages: write
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         name: Check out code
--- a/.github/workflows/cherrypick.yml
+++ b/.github/workflows/cherrypick.yml
@@ -21,9 +21,7 @@ on:
     types: [closed]
 
 permissions:
-  contents: write
-  pull-requests: write
-  issues: write
+  contents: read
 
 jobs:
   add-labels:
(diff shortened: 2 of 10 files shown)


.github/workflows/image-builds.yaml
rahulait, line 152: This also needs write permissions

rahulait (approved):

cdesiniotis:
Converting to draft while I figure out why the CI is not running.

rahulait:
> Converting to draft while I figure out why the CI is not running.
I see this in logs:

The workflow is not valid. .github/workflows/ci.yaml (Line: 37, Col: 3): Error calling workflow 'NVIDIA/gpu-operator/.github/workflows/code-scanning.yaml@439d1da5037074434d66ad06dfd181906f7e0904'. The workflow is requesting 'contents: read', but is only allowed 'contents: none'.

cdesiniotis:
> > Converting to draft while I figure out why the CI is not running.
> > I see this in logs:
> 
> The workflow is not valid. .github/workflows/ci.yaml (Line: 37, Col: 3): Error calling workflow 'NVIDIA/gpu-operator/.github/workflows/code-scanning.yaml@[439d1da](https://github.com/NVIDIA/gpu-operator/commit/439d1da5037074434d66ad06dfd181906f7e0904)'. The workflow is requesting 'contents: read', but is only allowed 'contents: none'.

I believe I fixed this issue now. But the CI is not running after I force-pushed. The CI appears to be working if I manually invoke it though... https://github.com/NVIDIA/gpu-operator/actions/runs/32986504647

.github/workflows/image-builds.yaml
cdesiniotis: Thanks for pointing this out. Updated.

cdesiniotis:
It got triggered now https://github.com/NVIDIA/gpu-operator/actions/runs/32987573904

PR #2840, opened 2026-08-31
Add per-state duration metric to reconciliation

rajathagasthya:
## Description

A reconcile processes many component states, but existing metrics only
expose whole-reconcile totals, so a slow state cannot be identified
from metrics alone. Add a `gpu_operator_state_seconds` histogram
labeled by controller and state, covering the ClusterPolicy,
GPUCluster, and NVIDIADriver paths.

## Checklist

## Testing

`go test ./controllers/ ./internal/state/`; `go build`; `gofmt`.

### DIFF
--- a/internal/state/metrics.go
+++ b/internal/state/metrics.go
@@ -1,0 +1,39 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package state
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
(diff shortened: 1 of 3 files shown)


internal/state/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// StateDurationSeconds records the time spent per reconcile in each state,
+// labeled by the owning controller's CRD kind and the state name.
+var StateDurationSeconds = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Namespace: "gpu_operator",
```
cdesiniotis, line 29: Question -- is this related to the namespace that the operator is installed in? Or is this completely independent of that? (I believe it is the latter).

cdesiniotis (commented):
Overall, lgtm. One question -- should we also add per-controller duration metrics? E.g. time spent per reconcile for X controller? (not specific to a single state)

internal/state/metrics.go
rajathagasthya: Yeah, it's independent of the k8s namespace. This is the namespace for the metrics in Prometheus, so the `state_seconds` in the next line becomes in `gpu_operator_state_seconds`.

rajathagasthya:
> One question -- should we also add per-controller duration metrics? E.g. time spent per reconcile for X controller? (not specific to a single state)

@cdesiniotis controller-runtime [already exports](https://github.com/kubernetes-sigs/controller-runtime/blob/5f8d96b043ef8ccadc544535415395be336630e8/pkg/internal/controller/metrics/metrics.go#L49-L56) a metric called `controller_runtime_reconcile_time_seconds` for every controller registered. The metric in this PR just breaks it down by "state" within a reconcile.

cdesiniotis (approved):

PR #2843, opened 2026-09-01
Add configurable securityContext and dnsConfig for the operator Deployment

### LINKED ISSUE #2533: [Feature]: Provide good securityContext by default
## Describe the current problem

Currently the problem is twofold:
1. Some `securityContext` are not configurable by the helm chart.
2. Several default values for `securityContext` are not as secure as they should be

For example, the deployment for the `gpu-operator` has no configurable `securityContext` and, by default, it is left empty.
**It is doubtful that this deployment needs full capabilities, privilege escalation, etc.** Hence it would make sense to provide a good security posture out of the box.

## Describe the solution you'd like
Allow configuring securityContexts of the different pods that can be deployed through this helm chart. (partially discussed here https://github.com/NVIDIA/gpu-operator/issues/2070)
Provide good default for the various `securityContext` (aka stricted possible while allowing the container to fulfill its role), e.g.:
```yaml
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    readOnlyRootFilesystem: true
    runAsGroup: 2000
    runAsNonRoot: true
    runAsUser: 2000
    seccompProfile:
      type: RuntimeDefault

```

## What is the added value?
This would improve the default security out of the box for helm chart users. If the containers currently support it, there isn't much downside to improving the default security.

It also helps clarifying to the end-user that these values are officially supported and will not cause any problem with the containers (e.g.: `readOnlyRootFilesystem: true` could for example cause problems for containers expecting to write to specific directories). By having them already specified, the end-user does not need to do extensive testing to ensure it does not cause any issue.

dundysm:
## Description

The `gpu-operator` Deployment currently has no pod or container `securityContext`, so it runs with the cluster default (often unrestricted). This PR sets restricted defaults that still work on vanilla Kubernetes and OpenShift, makes the pod `securityContext` configurable through Helm values, and adds `dnsPolicy` / `dnsConfig` values.

This also covers the Helm hook Jobs that run the same operator image (`upgrade-crd`, `cleanup-crd`, `cleanup-gpucluster`). Those were previously empty like the Deployment.

**Not in this PR:** operand DaemonSets (driver, device plugin, toolkit, MIG, and so on). Those are created later by the operator, not by this chart, and they need host privileges. ClusterPolicy already has `spec.daemonsets.podSecurityContext` for that path.

Defaults:
- pod (`operator.securityContext`, configurable): `runAsNonRoot: true`, `seccompProfile: RuntimeDefault`
- container (fixed in the templates, not a values key): `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, `capabilities.drop: [ALL]`
- `dnsPolicy: ""` (cluster default ClusterFirst) and `dnsConfig: {}`

The container-level fields are not exposed as a values key to keep the chart's API surface small, per review. They have to live at container level because `allowPrivilegeEscalation`, `readOnlyRootFilesystem`, and `capabilities` are not fields of `PodSecurityContext`.

No writable `/tmp` is added. The Deployment is already annotated with `openshift.io/scc: restricted-readonly`, and that SCC sets `readOnlyRootFilesystem: true`, so it has been running read-only on OpenShift without one. Neither `manage-crds` nor `cleanup-gpuclusters` creates temporary files.

UID/GID/`fsGroup` are not pinned. The operator image already uses `USER 1000:1000`, so vanilla Kubernetes still runs non-root. OpenShift `restricted-readonly` SCC uses `MustRunAsRange`; a hardcoded UID fails namespaces whose allocated range does not include that ID. Pin a UID with `operator.securityContext.runAsUser` if you need one.

Fixes #2533

## Checklist

- [ ] Lint checks passing (`make lint`) — n/a, no Go changes
- [ ] Generated assets in-sync (`make validate-generated-assets`) — n/a, no API/CRD/generated changes
- [ ] Go mod artifacts in-sync (`make validate-modules`) — n/a, no dependency changes
- [ ] Test cases are added for new code paths — n/a, Helm template-only change; verified with `helm template` below

## Testing

Helm-only change (no Go, CRD, or generated-asset edits). The diff against `main` is additions only (+79 / −0). Ran against this branch:

- `helm lint deployments/gpu-operator/` — pass
- `helm template` with chart defaults and `operator.cleanupCRD=true` / `operator.upgradeCRD=true` (both default to `false`), diffed against `main`: the only change is the added pod and container `securityContext` on the Deployment and the three hook Jobs. `dnsPolicy`/`dnsConfig` are omitted by default
- `helm template` with `operator.dnsPolicy=None` and `operator.dnsConfig.nameservers=[8.8.8.8]` — both render on the Deployment and hook Jobs
- `helm template` with `operator.securityContext=null` — pod `securityContext` is omitted
- `helm template` with `operator.securityContext.runAsUser=1000` — explicit UID still overrides
- `helm template` with `operator.pprof.bindAddress=:6060` — `--pprof-bind-address` still renders, same as `main`
- Confirmed `docker/Dockerfile` ends with `USER 1000:1000`

Did not apply this to a live GPU cluster.

### DIFF
--- a/deployments/gpu-operator/templates/cleanup_crd.yaml
+++ b/deployments/gpu-operator/templates/cleanup_crd.yaml
@@ -32,10 +32,27 @@ spec:
       {{- end }}
       nodeSelector:
         {{- toYaml .Values.operator.nodeSelector | nindent 8 }}
+      {{- with .Values.operator.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.operator.dnsPolicy }}
+      dnsPolicy: {{ .Values.operator.dnsPolicy }}
+      {{- end }}
+      {{- with .Values.operator.dnsConfig }}
+      dnsConfig:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: cleanup-crd
           image: {{ include "gpu-operator.fullimage" . }}
           imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
           command:
             - /usr/bin/manage-crds
           args:
--- a/deployments/gpu-operator/values.yaml
+++ b/deployments/gpu-operator/values.yaml
@@ -121,6 +121,17 @@ operator:
     requests:
       cpu: 200m
       memory: 100Mi
+  # Restricted defaults for the operator Deployment (not operand DaemonSets).
+  # Do not pin runAsUser/runAsGroup/fsGroup: the image already uses USER 1000:1000
+  # on vanilla Kubernetes, and a hardcoded UID fails OpenShift namespaces whose
+  # allocated range does not include 1000 (restricted-readonly SCC is MustRunAsRange).
+  securityContext:
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
+  # Empty dnsPolicy keeps the cluster default (ClusterFirst).
+  dnsPolicy: ""
+  dnsConfig: {}
   # metrics:
   #   serviceMonitor:
   #     interval: 15s
--- a/deployments/gpu-operator/templates/cleanup_gpucluster.yaml
+++ b/deployments/gpu-operator/templates/cleanup_gpucluster.yaml
@@ -37,10 +37,27 @@ spec:
       {{- end }}
       nodeSelector:
         {{- toYaml .Values.operator.nodeSelector | nindent 8 }}
+      {{- with .Values.operator.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.operator.dnsPolicy }}
+      dnsPolicy: {{ .Values.operator.dnsPolicy }}
+      {{- end }}
+      {{- with .Values.operator.dnsConfig }}
+      dnsConfig:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: cleanup-gpucluster
           image: {{ include "gpu-operator.fullimage" . }}
           imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
           command:
             - /usr/bin/cleanup-gpuclusters
             - --gpucluster-name
--- a/deployments/gpu-operator/templates/operator.yaml
+++ b/deployments/gpu-operator/templates/operator.yaml
@@ -33,6 +33,17 @@ spec:
       {{- if .Values.operator.priorityClassName }}
       priorityClassName: {{ .Values.operator.priorityClassName }}
       {{- end }}
+      {{- with .Values.operator.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.operator.dnsPolicy }}
+      dnsPolicy: {{ .Values.operator.dnsPolicy }}
+      {{- end }}
+      {{- with .Values.operator.dnsConfig }}
+      dnsConfig:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
       - name: gpu-operator
         image: {{ include "gpu-operator.fullimage" . }}
@@ -83,6 +94,12 @@ spec:
         resources:
           {{- toYaml . | nindent 10 }}
       {{- end }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+              - ALL
         ports:
           - name: metrics
             containerPort: 8080
--- a/deployments/gpu-operator/templates/upgrade_crd.yaml
+++ b/deployments/gpu-operator/templates/upgrade_crd.yaml
@@ -81,10 +81,27 @@ spec:
       {{- end }}
       nodeSelector:
         {{- toYaml .Values.operator.nodeSelector | nindent 8 }}
+      {{- with .Values.operator.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.operator.dnsPolicy }}
+      dnsPolicy: {{ .Values.operator.dnsPolicy }}
+      {{- end }}
+      {{- with .Values.operator.dnsConfig }}
+      dnsConfig:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: upgrade-crd
           image: {{ include "gpu-operator.fullimage" . }}
           imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - ALL
           command:
             - /usr/bin/manage-crds
           args:


deployments/gpu-operator/values.yaml
```diff
@@ -118,6 +118,23 @@ operator:
     requests:
       cpu: 200m
       memory: 100Mi
+  # Restricted defaults for the operator Deployment (not operand DaemonSets).
+  # UID/GID 1000 matches the operator image USER.
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 1000
+    runAsGroup: 1000
+    fsGroup: 1000
```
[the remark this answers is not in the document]
dundysm: Addressed in 0849d04. Dropped the pinned UID/GID/fsGroup so OpenShift MustRunAsRange can allocate. The image still uses USER 1000:1000, so vanilla Kubernetes keeps running non-root. Callers who need a pinned UID can set operator.securityContext.runAsUser.

dundysm:
@coderabbitai review

dundysm:
@rajathagasthya picking this up from #2533, including `dnsConfig` as you asked there.

restricted defaults for the operator deployment and the helm hook Jobs that run the same image. no pinned UID, so OpenShift `MustRunAsRange` can still allocate. `dnsPolicy` / `dnsConfig` are configurable (`dnsPolicy` stays empty so ClusterFirst remains the default). operand DaemonSets are unchanged; they still need host privileges.

this is my first PR here, so copy-pr-bot is blocking CI. If the approach looks right, would you mind an `/ok-to-test`?

rajathagasthya:
Thanks @dundysm! Will take a look at this PR soon.

rajathagasthya:
@coderabbitai review

rahulait (approved):
Overall, LGTM

deployments/gpu-operator/templates/cleanup_crd.yaml
```diff
@@ -32,10 +32,28 @@ spec:
...
       containers:
         - name: cleanup-crd
           image: {{ include "gpu-operator.fullimage" . }}
           imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
+        {{- with .Values.operator.containerSecurityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+        {{- end }}
+          volumeMounts:
+            - name: tmp
```
tariq1890: why is this needed?

deployments/gpu-operator/templates/cleanup_crd.yaml
```diff
@@ -32,10 +32,28 @@ spec:
...
+      {{- end }}
+      {{- with .Values.operator.dnsConfig }}
+      dnsConfig:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: cleanup-crd
           image: {{ include "gpu-operator.fullimage" . }}
           imagePullPolicy: {{ .Values.operator.imagePullPolicy }}
+        {{- with .Values.operator.containerSecurityContext }}
```
tariq1890: Is a containerSecurityContext really needed? Shouldn't the pod Security Context suffice? I am not sure what the advantages are with being able to customise the container-level security context.

dundysm: _**on whether it's needed:**_ allowPrivilegeEscalation, readOnlyRootFilesystem, and capabilities exist only on v1.SecurityContext; they aren't fields on v1.PodSecurityContext, so pod-level can't express them at any version. pss restricted requires the first two per-container, so the pod block alone can't get us admitted into a restricted namespace. pod-level here carries only runAsNonRoot and seccompProfile, which do accept pod-level.

**_on customisability:_** that's a fair challenge. i exposed it because helm create convention splits pod vs container securityContext, and because relaxing readOnlyRootFilesystem or adding back a capability is the kind of thing a downstream consumer hits on odd runtimes. 

but, i take the point that clusterpolicy.spec.daemonsets only exposes podSecurityContext, so a container-level knob is new surface for this project. happy to hardcode the three restricted values and drop operator.containerSecurityContext from values.yaml if you'd prefer less api surface, the fields still have to live at container level either way.

deployments/gpu-operator/templates/cleanup_crd.yaml
dundysm: you're right, dropping it. i'd added it as insurance for `readOnlyRootFilesystem: true`, but checking properly: the operator deployment is already annotated with `openshift.io/scc: restricted-readonly` (values.yaml), and that SCC sets `readOnlyRootFilesystem: true`, so it has been running with a read-only root filesystem on OpenShift without a writable `/tmp`.

`manage-crds` and `cleanup-gpuclusters` only read manifests from `/opt/gpu-operator` and apply them through the API server, no temp writes in either cmd or in `crdutil`.

removed the emptyDir and the mounts from all four templates in faf30f8. `readOnlyRootFilesystem: true` stays on the container securityContext.

dundysm:
@coderabbitai review the updates

rahulait:
@dundysm can you squash your commits into one?

dundysm:
@rahulait squashed into one commit: 19ed550

assets/state-mig-manager/0420_configmap.yaml
```diff
@@ -14,13 +14,13 @@ data:
       echo "waiting for the driver validations to be ready..."
       sleep 5
     done
-
+
```
tariq1890: Let's remove the extraneous white spaces here

bundle/manifests/resource.nvidia.com_computedomains.yaml
```diff
@@ -66,13 +66,12 @@ spec:
...
                 type: object
               numNodes:
-                default: 0
                 description: |-
                   Intended number of IMEX daemons (i.e., individual compute nodes) in the
                   ComputeDomain. Must be zero or greater.
 
                   With `featureGates.IMEXDaemonsWithDNSNames=true` (the default), this is
-                  recommended to be set to zero (default). Workload must implement and consult its
+                  recommended to be set to zero. Workload must implement and consult its
```
tariq1890: This shouldn't be a part of this PR's diff

cmd/gpu-operator/main.go
```diff
@@ -78,12 +78,9 @@ func main() {
 	var enableLeaderElection bool
 	var leaderElectionNamespace string
 	var probeAddr string
-	var pprofAddr string
```
tariq1890, line 81: This shouldn't be a part of the PR diff either. Please make sure the PR is rebased correctly

deployments/gpu-operator/templates/cleanup_crd.yaml
tariq1890: Yes, let's limit the api surface for now. If we find the need to specify container-level security contexts, we can revisit that in a future PR.

dundysm:
rebased that squash onto current `main` so the diff is only the helm chart changes again (+77 / −6). tip is now 09022df.

.github/workflows/image-builds.yaml
```diff
@@ -148,7 +148,6 @@ jobs:
 
   build-multi-arch-images:
     needs: [variables, build-gpu-operator-arm64, build-gpu-operator-amd64]
-    if: needs.variables.outputs.push_on_build == 'true'
```
tariq1890, line 151: This shouldn't be a part of the PR diff.

deployments/gpu-operator/templates/cleanup_crd.yaml
dundysm: Done in acf5b99. `operator.containerSecurityContext` is removed from values.yaml; `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, and `capabilities.drop: [ALL]` are now fixed in the Deployment and the three hook Job templates.

dundysm:
pushed acf5b99 to address the latest review:

- removed `operator.containerSecurityContext`. the container securityContext is now fixed in the deployment and hook Job templates.
- fixed the rebase. the previous squash had reverted the chart side of 41c6c1a (`operator.pprof.bindAddress` / `--pprof-bind-address`). that's restored, and the diff against `main` is now additions only (+79 / −0).

the description is updated to match.

cmd/gpu-operator/main.go
dundysm: thanks for catching this. these came from my first squash (19ed550), which was rebased incorrectly. the rebase onto current `main` removed them, but it also reverted the chart side of 41c6c1a (the `operator.pprof.bindAddress` value and the `--pprof-bind-address` arg). that's restored in acf5b99, and the diff against `main` is now additions only (+79 / −0) in the five helm chart files.

assets/state-mig-manager/0420_configmap.yaml
dundysm: out of the diff now. this came in with the bad squash (19ed550); acf5b99 only touches the five helm chart files.

bundle/manifests/resource.nvidia.com_computedomains.yaml
dundysm: out of the diff now. this came in with the bad squash (19ed550); acf5b99 only touches the five helm chart files.

.github/workflows/image-builds.yaml
dundysm: out of the diff now. this came in with the bad squash (19ed550); acf5b99 only touches the five helm chart files.

deployments/gpu-operator/values.yaml
```diff
@@ -121,6 +121,17 @@ operator:
...
       cpu: 200m
       memory: 100Mi
+  # Restricted defaults for the operator Deployment (not operand DaemonSets).
+  # Do not pin runAsUser/runAsGroup/fsGroup: the image already uses USER 1000:1000
+  # on vanilla Kubernetes, and a hardcoded UID fails OpenShift namespaces whose
+  # allocated range does not include 1000 (restricted-readonly SCC is MustRunAsRange).
+  securityContext:
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
```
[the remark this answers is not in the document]
rahulait: We don't support such old k8s versions. This should be fine.

tariq1890 (approved):

PR #2853, opened 2026-09-03
build: use the official docker image to install busybox

tariq1890:
Instead of using an immutable tag of the `debian:trixie-slim` image and an unpinned busybox apt package, we use the busybox binary from the official dockerhub busybox image.

This has two advantages

- We version-pin the extracted busybox binary
- Using the official busybox image helps us stay up-to-date with the latest versions faster as the debian-trixie apt package repos don't host the latest busybox versions immediately. For e.g., at the time of filing this PR, the latest busybox version available in the debian trixie package repo is still 1.37, whereas the actual latest version of busybox is 1.38

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -76,11 +76,8 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
 # Build a static busybox layout: one binary plus applet symlinks (sh, rm,
 # ln, sleep, cat, ...) so PATH-resolved commands in init-container wrappers
 # and lifecycle hooks keep working on the non-*-dev* distroless base.
-FROM debian:trixie-slim@sha256:3a39a0592364683e6bab97937b72cad5a8fa6dcbbee90edb3bb48c7f8e94f258 AS shell
-RUN apt-get update \
- && apt-get install -y --no-install-recommends busybox-static \
- && rm -rf /var/lib/apt/lists/* \
- && mkdir /busybox \
+FROM busybox:1.38.0@sha256:dc2d74b28e4cf8984fa52af1f39bc7c3d9c73760b41a74d629f5d11b1ab28616 AS shell
+RUN mkdir /busybox \
  && cp /bin/busybox /busybox/busybox \
  && /busybox/busybox --install -s /busybox
 


rajathagasthya:
> This helps keep track of the busybox versions used in the gpu-operator container image.

I don't think this is true. Busybox is installed via `apt` without version pinning, so its version will be whatever is served by the apt repository at build time. Pinning the base image doesn't help with that, although it's probably fine to do.

tariq1890:
Completely agree, thanks for pointing that out @rajathagasthya !

I can make sure we pin the busybox version when apt-get'ing the package.

We pin everything :)

tariq1890:
@rajathagasthya I've switched to a different implementation; the commit message and PR descriptions have also been updated to reflect this. Requesting another round of review.

docker/Dockerfile
```diff
@@ -76,11 +76,8 @@ RUN curl -L https://codeload.github.com/NVIDIA/cuda-samples/tar.gz/refs/tags/v${
 # Build a static busybox layout: one binary plus applet symlinks (sh, rm,
 # ln, sleep, cat, ...) so PATH-resolved commands in init-container wrappers
 # and lifecycle hooks keep working on the non-*-dev* distroless base.
-FROM debian:trixie-slim@sha256:3a39a0592364683e6bab97937b72cad5a8fa6dcbbee90edb3bb48c7f8e94f258 AS shell
-RUN apt-get update \
- && apt-get install -y --no-install-recommends busybox-static \
- && rm -rf /var/lib/apt/lists/* \
- && mkdir /busybox \
+FROM busybox:1.38.0@sha256:dc2d74b28e4cf8984fa52af1f39bc7c3d9c73760b41a74d629f5d11b1ab28616 AS shell
```
rajathagasthya, line 79: FYI: Claude says busybox from this image is dynamically linked and depends on `glibc` in `distroless/cc`. We were previously installing `busybox-static` package. 

I think a mismatch between `glibc` version needed by busybox vs what's in `distroless/cc` would fail here at build time, so I think it's okay? https://github.com/NVIDIA/gpu-operator/blob/df32e7e61ad2bc6677c302ec4d805ddaf43a13c4/docker/Dockerfile#L120

It's a problem if this line is ever removed though because then we wouldn't be able to detect failure at build time. But it's highly unlikely this line will ever be removed. 

Another option is to keep the existing `apt` install, but pin the [latest version](https://packages.debian.org/trixie/busybox-static) and configure Renovate to bump it automatically. 

```
apt-get install -y --no-install-recommends busybox-static=1:1.37.0-6+b8
```

rajathagasthya: Apparently, `busybox:1.38.0-musl` gives a fully static binary, but both that and `busybox:1.38.0` ship 400+ "applets" (programs like `rm`, `sleep`) vs 272 from the Debian package. 

It will increase CVE surface area on paper because scanners don't know the applets are just symlinks to `/busybox/busybox` itself since it's a [multi-call binary](https://github.com/NVIDIA/gpu-operator/pull/2434#discussion_r3275682282) and not separate programs. Something to consider.

tariq1890:
Thanks for the reviews @rajathagasthya. We definitely don't want to use dynamically linked binary. I also agree that CVE surface area increase with statically built busybox-musl binary is not worth it.

Closing this PR.

Issue #2854, opened 2026-09-04
[Bug]: nvidia-dcgm hostengine can start without NVML during a driver restart and never recover


gengwg:
**Describe the bug**

With the standalone DCGM hostengine enabled (`dcgm.enabled: true`), the `nvidia-dcgm` pod can start while the nvidia kernel module is not loaded and then run indefinitely with no NVML. Both it and `nvidia-dcgm-exporter` stay `1/1 Running` and Ready while exporting zero `DCGM_FI_*` series for that node, so GPU health is silently invisible until someone notices.

The mechanism:

1. The `nvidia-dcgm` DaemonSet's only init container is `toolkit-validation`, which checks `/run/nvidia/validations/toolkit-ready`. That file survives an in-place driver container restart, so the gate passes while the driver is mid-reload. Only `nvidia-container-toolkit-daemonset` and `nvidia-operator-validator` carry `driver-validation`.
2. The toolkit-installed host wrapper (`/usr/local/nvidia/toolkit/nvidia-container-runtime`, from `nvidia-ctk-installer` with `CheckModules`) does `grep "^nvidia " /proc/modules || exec runc "$@"`. When the module is unloaded at container create time the container gets plain runc with no library or device injection. The hostengine pod logs:
   ```
   nvidia driver modules are not yet loaded, invoking /usr/local/bin/runc directly
   Started host engine version 4.4.2 using port number: 5555
   ```
3. nv-hostengine has no `libnvidia-ml.so` in its filesystem, starts anyway, and never re-initialises. The container has no liveness probe, so nothing restarts it.
4. `nvidia-dcgm-exporter` (started a little later, after the module loaded; it even hit `failed to stat CDI host device /run/nvidia/driver/dev/nvidia-modeset` a few times until the driver finished) connects via `DCGM_REMOTE_HOSTENGINE_INFO=nvidia-dcgm:5555` and logs:
   ```
   Not collecting GPU metrics; error getting devices count: Cannot perform the requested operation because NVML doesn't exist on this system.
   ```
   It then serves an empty metrics page while Ready. Restarting only the exporter does not help; it caches the empty device list from the still-deaf hostengine.

[...]

PR #2855, opened 2026-09-04
Probe DCGM hostengine GPU visibility in the liveness check

gengwg:
## Description

The `nvidia-dcgm` hostengine can start while the nvidia kernel module is not loaded (an operator rollout that recreates the dcgm pod while the driver container restarts, for example). The toolkit's host runtime wrapper then execs plain runc with no injection, `nv-hostengine` starts with no NVML, and nothing ever restarts it. The pod stays `1/1 Running` while exporting zero `DCGM_FI_*` series.

The `tcpSocket` probe on 5555 passes on a deaf hostengine because it still listens. This replaces it with an exec probe that requires `dcgmi discovery -l` to report at least one active GPU, so kubelet restarts the hostengine once the driver is back. The check is skipped while no driver is present (no `nvidia` module and no WSL2 `/dev/dxg`), so a planned driver restart does not loop the hostengine. The readiness probe is unchanged.

The init-container gate from the original version of this PR is no longer needed: #2881 requires a live nvidia module, or the WSL2 `/dev/dxg` path, before every operand that uses the NVIDIA runtime, so this PR now carries only the probe change.

Behaviour worth stating: once the driver is present, a node that genuinely loses every GPU sees its hostengine restart every ~3 minutes instead of sitting silently at 0 GPUs. That is intended; it is visible where the previous state was not. Nodes with at least one GPU visible are unaffected. When the hostengine restarts, `dcgm-exporter` hits `DCGM_ST_CONNECTION_NOT_VALID`, which it already treats as fatal and exits, so it reconnects on its own restart.

Refs #2854

## Checklist

- [ ] Lint checks passing (`make lint`) - YAML-only change under `assets/`
- [x] Generated assets in-sync (`make validate-generated-assets`) - no generated files touched
- [x] Go mod artifacts in-sync (`make validate-modules`) - go.mod untouched
- [ ] Test cases are added for new code paths - no Go code paths changed

## Testing

- Probe pattern checked against the DCGM source: `dcgmi discovery -l` prints `N GPUs found (Active).` on stdout, and `0 GPUs found (Active).` when no GPU is visible, so the anchored count matches only the visible case.
- `go test ./controllers/... ./internal/...` passes, including `TestOperandToolkitReadinessWaitsForLiveDriver` from #2881.
- Not end-to-end tested with a rebuilt operator image on a live cluster.


myeolenv:
Thanks for the contribution. We will review the PR soon.

rajathagasthya:
Thanks for the contribution!

As mentioned in https://github.com/NVIDIA/gpu-operator/issues/2854#issuecomment-5611605401, any fix for #2874 should also fix this. Although, I think the enhanced liveness probe could be valuable on its own.

Let's keep this open and we'll evaluate the probe changes.

rahulait:
Hi @gengwg , we have merged changes for option 1 in https://github.com/NVIDIA/gpu-operator/pull/2881

Can you rebase your changes?

rajathagasthya:
@gengwg As Rahul said, the issue is fixed by #2881. If you want to keep the liveness probe change, we can evaluate. Otherwise, close this PR. Let us know what you'd prefer.

gengwg:
Hi @rahulait @rajathagasthya — rebased onto `main` and kept only the liveness probe change, so the diff is now one file.

The init-container change is dropped: #2881 already gates every NVIDIA-runtime operand on a live module, and it also covers the WSL2 `/dev/dxg` path that my version did not. Upstream already carries that gate, so this PR reduces to one file again and `assets/state-dcgm-exporter/0800_daemonset.yaml` matches upstream. (Note: the fork's `main` predates #2881, so a diff against the fork shows the gate as if it were ours — the PR is against NVIDIA's `main`.)

What remains is `assets/state-dcgm/0400_dcgm.yml`: the `nvidia-dcgm` hostengine liveness probe becomes an exec check on `dcgmi discovery -l` instead of `tcpSocket: 5555`, so a hostengine that came up without NVML is restarted by kubelet once the driver is back. Readiness is unchanged.

Two notes for the evaluation:

- The pattern `^[1-9][0-9]* GPUs? found` is deliberate. `dcgmi discovery -l` prints `8 GPUs found (Active).` (see `dcgmi/Query.cpp`) and `0 GPUs found (Active).` when nothing is visible, so an anchored non-zero count accepts the first and rejects the second.
- The check is skipped while no driver is present (no `nvidia` module and no WSL2 `/dev/dxg`), so a planned driver restart does not loop the hostengine. Once the driver is back the probe fails and kubelet restarts the deaf hostengine within ~3 minutes. If you'd rather keep the probe unconditional, I can drop the guard — the tradeoff is a restart loop for the whole driver outage.
- With the guard, a node that genuinely loses every GPU *while the driver is present* still restarts its hostengine roughly every 3 minutes instead of sitting silently at 0 GPUs. That is intended — visible where the previous state was not.

Verified: `TestOperandToolkitReadinessWaitsForLiveDriver` and `go test ./controllers/... ./internal/...` pass on the rebase. I have not run this end to end on a cluster — my original image check was against `dcgm:4.4.2` and `values.yaml` now defaults to `4.6.0`, so I can re-verify against 4.6.0 if that helps.

Happy to keep it or drop it — your call.

PR #2860, opened 2026-09-04
Propagate image pull secrets to compute-domain-daemon pod specs

karthikvetrivel:


manifests/state-dra-driver/0600_controller-deployment.yaml
```diff
@@ -108,6 +108,12 @@ spec:
               fieldPath: metadata.namespace
         - name: IMAGE_NAME
           value: {{ .DRADriver.ImagePath | quote }}
+        {{- if .DRADriver.Spec.ImagePullSecrets }}
+        # Comma-separated secret names propagated to the compute-domain-daemon
+        # DaemonSet pod specs created by the controller.
```
tariq1890, line 113: Can we remove these comments?

Issue #2795, opened 2026-08-25
[Bug]: Deprecated `cdi.default` CRD default causes persistent Argo CD drift


dentiny:
**Describe the bug**
Hi team, I found there's inconsistency between cluster policy definition and helm-rendered cluster policy, which leads to ArgoCD keep reporting out-of-sync.

The issue is

The deprecated `spec.cdi.default` field is no longer rendered by the [ClusterPolicy Helm template](https://github.com/NVIDIA/gpu-operator/blob/7b3eb8393ef506a343cc3990aa14edf3c0d991c4/deployments/gpu-operator/templates/clusterpolicy.yaml#L145-L149).
The rendered resource therefore contains:
```yaml
spec:
  cdi:
    enabled: true
```
However, the [ClusterPolicy CRD](https://github.com/NVIDIA/gpu-operator/blob/7b3eb8393ef506a343cc3990aa14edf3c0d991c4/deployments/gpu-operator/crds/nvidia.com_clusterpolicies.yaml#L137-L159) still defines `spec.cdi.default` with a schema default of `false`.
The corresponding [Go API definition](https://github.com/NVIDIA/gpu-operator/blob/7b3eb8393ef506a343cc3990aa14edf3c0d991c4/api/nvidia/v1/clusterpolicy_types.go#L1967-L1973) also marks the field as deprecated and unused while retaining `+kubebuilder:default=false`.
After the rendered resource is applied, the Kubernetes API server adds the default:
```yaml
spec:
  cdi:
    enabled: true
    default: false
```
This creates a persistent difference:
```text
Desired: spec.cdi.default is absent
Live:    spec.cdi.default is false
```
Argo CD reports the `ClusterPolicy` as `OutOfSync`. Re-syncing cannot resolve the difference because the API server adds the field again after every apply.

**To Reproduce**
N/A

**Expected behavior**
These two files are kept in sync.

**Environment (please provide the following information):**
 - GPU Operator Version: v26.3.3
 - OS: N/A, unrelated
 - Kernel Version: N/A, unrelated
 - Container Runtime Version: N/A, unrelated
 - Kubernetes Distro and Version: N/A, unrelated

**Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/)** (optional if deemed irrelevant)

 - [ ] kubernetes pods status: `kubectl get pods -n OPERATOR_NAMESPACE`
[...]

PR #2865, opened 2026-09-05
Revert "Stop defaulting deprecated cdi.default in ClusterPolicy"

tariq1890:
This reverts commit 23464863fc4491f2fb6f6ef9b7a5b47bd41842c8.

- We do not typically make non-additive changes to a CRD
- This change was made in response to an ArgoCD behaviour, which is just one among several tools for Kubernetes application deployment
- The [issue](https://github.com/NVIDIA/gpu-operator/issues/2795) is well-known and there is a well-documented solution. Enabling `server-side diff strategy` in ArgoCD fixes the issue for Argo users who face this problem. That is a better solution than making an upstream CRD schema change


tariq1890:
Closing as this Revert PR targets the cherrypick commit as opposed to the original commit.

PR #2881, opened 2026-09-10
Wait for the driver before starting GPU operands

### LINKED ISSUE #2874: Operands can start without driver libraries after a driver restart and fail (MIG manager errors out with `ERROR_LIBRARY_NOT_FOUND`)
## What happens

After the GPU driver is unloaded and reinstalled on a node, for example during a Kubernetes or CRI-O upgrade, the MIG manager pod can be recreated and its main container started while the NVIDIA kernel module is not loaded. The container is created through the `nvidia` runtime class, which falls back to the plain low-level runtime with no driver libraries injected. Once the driver is back, the MIG manager runs `nvidia-mig-parted` and fails:

```
Error applying MIG configuration with hooks: error initializing NVML: ERROR_LIBRARY_NOT_FOUND
```

The node is left at `nvidia.com/mig.config.state=failed` and stays there. There is no retry. Deleting the MIG manager pod fixes it, because the replacement container is created after the module is loaded and gets the driver libraries injected.

## Scope

This is not specific to the MIG manager. The underlying condition, where an operand init container that gates only on the `toolkit-ready` sentinel file so its main container can be created through the `nvidia` runtime class when the module is unloaded and receive no driver libraries, applies to every operand that uses the `nvidia` runtime class. 

This exact issue is seen in the DCGM operands in #2854. 

## Environment

- GPU Operator 25.10.0, also seen on 26.3.x
- CRI-O
- CDI enabled
- Containerized driver, not a pre-installed host driver
- MIG strategy mixed

## Root cause

The `nvidia` runtime injects driver libraries into a container only if the kernel module is loaded at container-create time. The bug is that the MIG manager container gets created during the window when the module is unloaded:

1. A CRI-O major or minor upgrade removes every pod sandbox on the node, which wipes all the containers.
2. The kubelet recreates all the pods together.
3. The driver pod comes back and its `k8s-driver-manager` init evicts the operands, then re-enables the operand labels in a single node update. The DaemonSet controllers reschedule all the operand pods at once, before the driver container reinstalls the driver.
4. The sentinel files under `/run/nvidia/validations` live on the host tmpfs, so they survive the CRI-O upgrade and are now stale for the freshly created pods.
5. The operator-validator preStop hook, which would have removed the `toolkit-ready` file, never runs: CRI-O wiped that container outside kubelet, and the kubelet runs preStop only for containers it still sees running.
6. The new MIG manager init gate checks only that `toolkit-ready` exists, finds the stale file, and passes immediately. Its main container starts before the driver and toolkit are ready, with no driver libraries injected, and eventually fails at NVML with `ERROR_LIBRARY_NOT_FOUND`.

## Logs

Driver pod init container (`k8s-driver-manager`) evicts and reschedules operands before the driver reinstalls:

```
Shutting down all GPU clients on the current node by disabling their component-specific nodeSelector labels
Waiting for nvidia-operator-validator to shutdown
Unloading NVIDIA driver kernel modules
Rescheduling all GPU clients on the current node by enabling their component-specific nodeSelector labels
```

MIG manager, created without libraries, fails once the driver is installed:

```
nvidia driver modules are not yet loaded, invoking /usr/libexec/crio/runc directly
waiting for the driver validations to be ready...        (repeats until driver is ready)
IS_HOST_DRIVER=false
Starting nvidia-mig-manager
level=fatal msg="Error applying MIG configuration with hooks: error initializing NVML: ERROR_LIBRARY_NOT_FOUND"
level=info msg="Changing the 'nvidia.com/mig.config.state' node label to 'failed'"
level=info msg="Waiting for change to 'nvidia.com/mig.config' label"
```

## How to reproduce

1. Bring up a MIG-capable node with the GPU Operator, `cdi.enabled=true`, and a containerized driver, and get it to `nvidia.com/mig.config.state=success`.
2. Recreate all pod sandboxes on the node without a reboot while the module is still loaded. A CRI-O minor version upgrade does this: CRI-O wipes all sandboxes when the major or minor version changes, and `/run` survives the upgrade so the sentinel files persist. Removing the sandboxes with `crictl` (`crictl pods --namespace gpu-operator -q | xargs -r crictl stopp` then `crictl pods --namespace gpu-operator -q | xargs -r crictl rmp`) should also work.
3. The driver pod's `k8s-driver-manager` init re-runs, and the toolkit, validator, and MIG manager pods are recreated together, before the driver reinstalls.
4. The MIG manager init passes on the stale `toolkit-ready`, its main container is created while the module is unloaded, it waits for `driver-ready`, and once the driver is back it fails with `ERROR_LIBRARY_NOT_FOUND`. The node stays at `mig.config.state=failed`.

> Note: reproduce with container-toolkit v1.18.1 or later. With v1.18.0 the runtime wrapper writes its "modules not yet loaded" message to stdout, which corrupts CRI-O's parsing of the runtime state output and leaves the operands in `Init:RunContainerError` until the driver returns, masking this bug as a transient. v1.18.1 and later write it to stderr, so the container starts and the failure surfaces. GPU Operator 25.10.0 defaults to v1.18.0.

## Suggested fixes

1. Gate the operand init containers on the module as well as the file, matching the runtime wrapper's own check:

   ```sh
   until [ -f /run/nvidia/validations/toolkit-ready ] && grep -q '^nvidia ' /proc/modules; do sleep 5; done
   ```

   Applies to mig-manager, device-plugin, gpu-feature-discovery, dcgm, dcgm-exporter, and mps-control-daemon.

2. Have `k8s-driver-manager` remove the sentinel files under `/run/nvidia/validations` before it re-enables the operand labels, including on its failure path. The validator recreates them once the driver is up. Removing them by hand after the operand container already exists does not help, because injection is fixed at creation time.

3. Have the MIG manager treat `ERROR_LIBRARY_NOT_FOUND` as unrecoverable and exit non-zero instead of waiting for a label change, so the kubelet recreates the container and the next one gets the libraries.

## Workaround

Delete the MIG manager pod on the affected node.

sylvesterkaczmarek:
Fixes #2874

## Description

Require the live NVIDIA kernel module as well as `toolkit-ready` before creating GPU Operator operand containers that use the NVIDIA runtime. This prevents a stale validation file from admitting an operand while the driver is temporarily unloaded, when the runtime would create the container without driver libraries.

The guard is applied consistently to GPU Feature Discovery, Device Plugin, DCGM, DCGM Exporter, MIG Manager, and MPS Control Daemon.

## Checklist

## Testing

- `make unit-test` - passed
- `go test ./controllers -run TestOperandToolkitReadinessWaitsForDriverModule -count=1` - passed
- `make fmt` - passed
- `make validate-modules` - passed
- `make lint` reports three existing `SA4023` findings in `cmd/nvidia-validator/main.go`; the same three findings reproduce on pristine `main`


rajathagasthya:
@sylvesterkaczmarek Thanks for the contribution! Can you please squash your commits and make sure it's [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)?

sylvesterkaczmarek:
Done. Squashed to one commit and re-signed it; the new head is 41de0bd6.

rajathagasthya:
@coderabbitai review

assets/gpu-feature-discovery/0500_daemonset.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         - name: toolkit-validation
           image: "FILLED BY THE OPERATOR"
           command: ['sh', '-c']
-          args: ["until [ -f /run/nvidia/validations/toolkit-ready ]; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
+          args: ["until [ -f /run/nvidia/validations/toolkit-ready ] && { grep -q '^nvidia ' /proc/modules || grep -qi microsoft /proc/sys/kernel/osrelease; }; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
```
cdesiniotis, line 32: I suggest we check for the presence of `/dev/dxg` which aligns with our WSL2 platform check in other components.
```suggestion
          args: ["until [ -f /run/nvidia/validations/toolkit-ready ] && { grep -q '^nvidia ' /proc/modules || [ -e /dev/dxg ] }; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
```

cdesiniotis: E.g. https://github.com/NVIDIA/nvidia-container-toolkit/blob/341598c1c36ad7fc4a80cb7a00239fdd5a59d053/deployments/systemd/nvidia-cdi-refresh.service#L30

sylvesterkaczmarek:
Updated the existing squashed commit for the WSL2 case found in review. Standard Linux still requires the live `nvidia` module, while WSL2 is accepted via the Microsoft kernel path once `toolkit-ready` exists. The new signed/verified head is `a721f488`; the targeted controller test and shell syntax check pass.

assets/gpu-feature-discovery/0500_daemonset.yaml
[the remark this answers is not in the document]
rajathagasthya: @sylvesterkaczmarek Looks like Chris' suggestion above of checking `[ -e /dev/dxg ]` is the right thing to do here.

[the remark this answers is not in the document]
sylvesterkaczmarek: Applied. The WSL2 alternative now checks the live `/dev/dxg` device instead of the static Microsoft kernel string, consistently across all six operand manifests. I updated the regression gate to require the same predicate. `go test ./controllers -run TestOperandToolkitReadinessWaitsForLiveDriver -count=1`, shell syntax validation, and `git diff --check` pass. I amended the existing single commit and re-signed it; GitHub verifies the new head `891d8e30`.

sylvesterkaczmarek:
@rajathagasthya The `/dev/dxg` review change is now on signed/verified head `891d8e30`. The targeted controller test and shell syntax check pass locally. NVIDIA runner checks need re-authorization for the new head when convenient.

controllers/object_controls_test.go
```diff
@@ -1259,6 +1259,45 @@ func TestVGPUManagerAssets(t *testing.T) {
...
+
+			var args string
+			for _, initContainer := range ds.Spec.Template.Spec.InitContainers {
+				if initContainer.Name == "toolkit-validation" {
+					args = strings.Join(initContainer.Args, " ")
+					break
+				}
+			}
+			require.NotEmpty(t, args, "toolkit-validation init container not found")
+			expectedGate := "until [ -f /run/nvidia/validations/toolkit-ready ] && { grep -q '^nvidia ' /proc/modules || [ -e /dev/dxg ]; }; do"
```
rajathagasthya, line 1294: non-blocking: I'm not really sure asserting a string instead of behavior adds any value. This test just becomes change detection.

assets/state-mig-manager/0600_daemonset.yaml
```diff
@@ -28,7 +28,7 @@ spec:
         - name: toolkit-validation
           image: "FILLED BY THE OPERATOR"
           command: ['sh', '-c']
-          args: ["until [ -f /run/nvidia/validations/toolkit-ready ]; do echo waiting for nvidia container toolkit to be setup; sleep 5; done"]
+          args: ["until [ -f /run/nvidia/validations/toolkit-ready ] && { grep -q '^nvidia ' /proc/modules || [ -e /dev/dxg ]; }; do echo waiting for nvidia container toolkit to be setup; sleep 5; done"]
```
rajathagasthya:
```suggestion
          args: ["until [ -f /run/nvidia/validations/toolkit-ready ] && { grep -q '^nvidia ' /proc/modules || [ -e /dev/dxg ]; }; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
```
Consistent with others.

controllers/object_controls_test.go
rajathagasthya: Let's keep it for now.

rajathagasthya (approved):

tariq1890 (approved):

assets/gpu-feature-discovery/0500_daemonset.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         - name: toolkit-validation
           image: "FILLED BY THE OPERATOR"
           command: ['sh', '-c']
-          args: ["until [ -f /run/nvidia/validations/toolkit-ready ]; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
+          args: ["until [ -f /run/nvidia/validations/toolkit-ready ] && { grep -q '^nvidia ' /proc/modules || [ -e /dev/dxg ]; }; do echo waiting for nvidia container stack to be setup; sleep 5; done"]
```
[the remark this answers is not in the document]
rajathagasthya: This doesn't change anything in practice. The failure described needs the module to disappear between init container exiting and main container being created. That gap exists for any check irrespective of the state of the module.

In the operator flow, the module is never unloaded while an operand pod exists. k8s-driver-manager ensures operands are evicted first before unloading.

Issue #428, opened 2022-10-27
entitlement free system documentation out of date to work with RHCOS on OCP 4.11 (need to do manual tag) 


relyt0925:
When following the steps defined at:
https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/openshift/install-gpu-ocp.html#create-the-clusterpolicy-instance

I don't ultimately get an entitlement free build system. I do see the driver container
```
Tylers-MacBook-Pro:~ tylerlisowski$ oc get imagestream -n openshift driver-toolkit
NAME             IMAGE REPOSITORY                                                            TAGS                           UPDATED
driver-toolkit   image-registry.openshift-image-registry.svc:5000/openshift/driver-toolkit   411.86.202210032349-0,latest   55 minutes ago
```

I use Nvidia GPU operator 22.9
![Screen Shot 2022-10-26 at 8 31 14 PM](https://user-images.githubusercontent.com/7716924/198170117-37ed3b99-e98d-4039-8453-db2867363bb4.png)

And then follow the following cluster policy steps for 4.11
```
oc get csv -n nvidia-gpu-operator gpu-operator-certified.v22.9.0 -ojsonpath={.metadata.annotations.alm-examples} | jq .[0] > clusterpolicy.json
oc apply -f clusterpolicy.json 
clusterpolicy.nvidia.com/gpu-cluster-policy configured
```

However the driver container still tries to fallback and pull from yum repos:
```
error: a container name must be specified for pod nvidia-driver-daemonset-411.86.202210072320-0-lvwjl, choose one of: [nvidia-driver-ctr openshift-driver-toolkit-ctr] or one of the init containers: [k8s-driver-manager]
Tylers-MacBook-Pro:~ tylerlisowski$ oc logs -n nvidia-gpu-operator nvidia-driver-daemonset-411.86.202210072320-0-lvwjl -c nvidia-driver-ctr -f
Running nv-ctr-run-with-dtk
+ [[ true == \t\r\u\e ]]
+ echo 'WARNING: RHCOS '\''411.86.202210072320-0'\'' imagetag missing, using entitlement-based fallback'
WARNING: RHCOS '411.86.202210072320-0' imagetag missing, using entitlement-based fallback
+ exec bash -x nvidia-driver init
+ set -eu
```

Does a manual image tag need to be created to use this path?

shivamerla:
[...]

PR #2882, opened 2026-09-11
Split API into a standalone Go module

aryangorwade:
## Description
This is a PR splitting `/api` into a standalone module, as described in this [design doc](https://docs.google.com/document/d/1Op5REEaLEaWMXkl12DkkAkvz5B2u-XWi4vCJkpJx87I/edit?tab=t.s2mtit3nkz4i), this would serve to fix GPU Operator versioning issues (Go supports `v0.x.y` and `v1.x.y`). 

This architectural split involves:
- Moving dependencies external to `api/` into `/api` to avoid circular imports
- Splitting API into  its standalone module with its own `go.mod` and `go.sum`; root `go.mod` references this module now

CI checks introduced: 
- PR check ensuring shared module dependencies stay in sync
- Post-release (after tagging) check ensuring compiled binary versions match tags
- PR check ensuring version bump PRs do not reference preexisting API tags (checks for "operator", "version", "bump" in PR title)

Changes to CI pipeline: 
- Modified `.nvidia-ci.yaml` and `.common-ci.yaml` to ensure that tagging a release commit with operator and api tags does not cause a separate release process to trigger for the `api` tag in addition to the operator tag

Dependabot changes: 
- Modified `dependabot.yaml` using `group-by: dependency-name`  and a `cross-directory` pattern to update each shared dependency between both modules in one PR (one PR per dependency; for both modules). Note that groups such as `k8s.io/*` can be created, resulting in one cross-directory PR for all deps under `k8s.io/*`. 

## Checklist

## Testing

Tested building & deploying the operator on a cluster with some test workloads. Tested Makefile commands.


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.26.3
 require (
 	github.com/Masterminds/sprig/v3 v3.3.0
 	github.com/NVIDIA/go-nvlib v0.12.0
-	github.com/NVIDIA/k8s-kata-manager v0.2.3
+	github.com/NVIDIA/gpu-operator/api v0.2603.3
```
aryangorwade: Should this be changed to a psuedo-verison `v0.0.0-00010101000000-000000000000` that is not updated in any version-bump PRs? If changes were made to `/api` after say release `26.03.3`, but those changes are not part of any current release; this could incorrectly point to a stale older API version.

~~A problem is that this would result in a built image's embedded binary reporting a psuedo-version and failing CVE scans. That is what currently happens with the `gpu-operator` and most of our Go products anyways~~

rajathagasthya: > A problem is that this would result in a built image's embedded binary reporting a psuedo-version and failing CVE scans.

I'm not sure setting a version for `github.com/NVIDIA/gpu-operator/api` has any effect on the version embedded for the `gpu-operator` binary. I think a pseudo-version will be embedded for the main binary regardless of what version a dependency is declared as.

aryangorwade: Makes sense. [#428](https://github.com/NVIDIA/cloud-native-team/issues/428) would only apply to the main binary which is the `gpu operator`.

PR #2889, opened 2026-09-11
Add unit tests for internal/conditions/gpucluster

abrarshivani:
## Description

`internal/conditions/gpucluster.go` had no unit tests. This adds them, taking the
`conditions` package from 64.6% to 100.0% of statements. Tests only, no production
code changed.

They follow the shape of the existing `nvidiadriver_test.go` and
`clusterpolicy_test.go`: fake client with a status subresource for the happy paths,
`interceptor.Funcs` to inject a conflict on the first status update and a plain
failure on every one.

`gpucluster.go` largely mirrors `nvidiadriver.go`. The one real difference is how the
error branch resolves `status.state`:

```go
// nvidiadriver.go                          // gpucluster.go
instance.Status.State = cr.Status.State     if instance.Status.State == "" {
                                                instance.Status.State = cr.Status.State
                                            }
```

`NVIDIADriverReconciler` sets `status.state` in memory without persisting it and then
writes conditions, so its updater has to copy the caller's value over.
`GPUClusterReconciler.updateCRStatus` persists the desired state first, so the stored
value is already authoritative and is only filled in when empty. Both are correct for
their own controller, and neither was covered. `SetConditionsErrorStateResolution`
covers the four combinations of stored and caller state;
`SetConditionsReadyLeavesStateAlone` covers the ready branch, which does not touch
`status.state` at all.

Not addressed here because it needs a production change: a typed-nil
`*nvidiav1alpha1.GPUCluster` satisfies the type assertion in `SetConditionsReady` and
`SetConditionsError` and then panics on `cr.Name`. All three updaters in this package
share that gap.

## Checklist

- [x] Lint checks passing (`make lint`) — no new findings. `make lint` does report 3
  pre-existing `staticcheck` deprecation warnings in `controllers/object_controls.go`
  and `internal/state/driver_volumes.go`; both files are untouched by this PR and the
  warnings are present on `main`.

## Testing

Unit tests only.

- 12 tests, 16 subtests, passing under `-race`
- `gpucluster.go` at 100% of statements, every function; package 64.6% to 100.0%
- Every assertion was checked for liveness by mutating `gpucluster.go` and confirming
  a test fails: 11 mutants, all killed. One assertion did not start out live.
  `GetError` matched only the wrapper text, so a `%w` to `%v` regression survived; it
  now asserts `apierrors.IsNotFound` through the wrapper. `RetryOnConflict` counts
  `Get` calls as well as update calls, which is what distinguishes a refetch inside
  the retry loop from one hoisted out of it.

### DIFF
--- a/internal/conditions/gpucluster_test.go
+++ b/internal/conditions/gpucluster_test.go
@@ -1,0 +1,306 @@
+/**
+# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+**/
+
+package conditions
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


rahulait:
@coderabbitai approve

rahulait (approved):

PR #2891, opened 2026-09-11
use a direct reader to check for resource claim for DRA

### LINKED ISSUE #2888: [Bug]: DRA teardown misses GPU claim Pods outside operator namespaces
**Describe the bug**

During a DRA-to-device-plugin transition, node labeling can remove `nvidia.com/gpu.deploy.dra-driver` while Pods in application namespaces still hold NVIDIA GPU ResourceClaims. Removing the DRA kubelet plugin too early can leave those Pods stuck in `Terminating` because their claims cannot be unprepared.

`nodeHasDRAClaimPods` lists Pods and resolves ResourceClaims through the manager's cached client. The cache only covers the operator and `openshift` namespaces, so claim-bearing Pods in other namespaces are invisible to the teardown guard. The node-name index does not expand that namespace coverage.

- [Cache configuration](https://github.com/NVIDIA/gpu-operator/blob/9f85a2ad08afc43b220cf368543fc2735244df13/cmd/gpu-operator/main.go#L128-L137)
- [Teardown guard](https://github.com/NVIDIA/gpu-operator/blob/9f85a2ad08afc43b220cf368543fc2735244df13/controllers/nodelabeling_controller.go#L394-L417)

**To Reproduce**

Scenario to validate on a DRA-capable cluster:

1. Configure the DRA stack with a `GPUCluster`.
2. Run a Pod in an application namespace using a ResourceClaim allocated by `gpu.nvidia.com`.
3. Transition to a `ClusterPolicy` while the claim-bearing Pod is still present on the node.
4. Check whether node labeling removes `nvidia.com/gpu.deploy.dra-driver` before the Pod has drained.

The namespace visibility gap is confirmed by source inspection. A real-cluster reproduction of the stuck-Pod symptom is still pending.

**Expected behavior**

DRA teardown should account for GPU claim-bearing Pods in every namespace, including terminating Pods and Pods with admin-access claims. The kubelet plugin should remain available until those Pods no longer require it.

The fix should preserve the operator's memory-efficiency goals: caching complete Pods and ResourceClaims across a large cluster could substantially increase its memory footprint.

**Possible approaches to evaluate**

- Broaden the existing Pod and ResourceClaim caches to all namespaces, using transforms to reduce retained fields where safe.
- Keep operand caches namespace-scoped and maintain a separate, compact informer-backed view of DRA workload dependencies.

kvalliyurnatt:
## Description

Currently we only cache Pod and Resource claims for the operator namespaces, so during the DRA to device plugin switch when we check to see if any pods on the node have resource claims we are only checking Pods that are on the operator namespaces, other Pods are ignored which can lead to issues if the DRA driver is removed with Pods still using the DRA driver. In this PR we attempt to fix it by using a direct API read instead of using the cached client, this does add an additional read to the API server but prevents having to cache all Pods and Resource claims of all namespaces, that can lead to large memory consumption on large clusters, we could look at optimizing what fields of Pod and Resource claims we store in the cache and optimize on the cache space, but doing the direct read in this case seemed simpler since this specific code path is only called very narrowly, I am open to rethinking that strategy as well, if we feel modifying the caching is a better solution.

## Checklist

## Testing

I have added unit tests, will test this and verify the bug is fixed, I was able to repro the bug.

fixes: https://github.com/NVIDIA/gpu-operator/issues/2888


kvalliyurnatt:
Closing this as this is known limitation that we don't plan to address at this moment

PR #2893, opened 2026-09-13
Map centos/SCOS OS tags to published rhel driver images

### LINKED ISSUE #2892: [Feature]: Allow overriding the derived OS-tag suffix on driver image versions - OKD/SCOS nodes (os_release ID=centos) resolve to nonexistent -centos10 tags
In OKD deployments ([okd-project/okd](https://github.com/okd-project/okd)) the nodes run CentOS Stream CoreOS (SCOS), whose NFD labels report `feature.node.kubernetes.io/system-os_release.ID=centos` and `…VERSION_ID=10`. `getOSTag()` in [internal/state/nodepool.go](https://github.com/NVIDIA/gpu-operator/blob/main/internal/state/nodepool.go#L155) therefore renders the driver image suffix as `centos10` → `nvcr.io/nvidia/driver:<version>-centos10`, a tag that is not published, and the driver pod fails with ImagePullBackOff. RHEL and its clones are already special-cased there (`ol`, `rocky`, `rhel` → major-only suffix); `centos` falls through to the default branch.

The documented workaround (confirmed in [#1533](https://github.com/NVIDIA/gpu-operator/issues/1533#issuecomment-3070802732)) is pinning `spec.driver.version` to a digest, which is used verbatim. That works, but freezes the image generation: rebuilds never flow under the same pin, leaving no tag-based upgrade path for the pinned reference. In practice we also hit the pinned tag being removed from nvcr.io ([gpu-driver-container#985](https://github.com/NVIDIA/gpu-driver-container/issues/985)), and the newer images we would otherwise move to were blocked by a separate issue ([gpu-driver-container#984](https://github.com/NVIDIA/gpu-driver-container/issues/984)).

**Expected behavior:** users running OKD/SCOS should be able to select a published, compatible driver image suffix (or otherwise prevent the operator from constructing an unavailable `-centos10` tag).

One shape that would solve it: a CRD-level override of the derived OS tag — e.g. `spec.driver.osTag: rhel10.0` — taking precedence over the os_release-derived suffix. A smaller alternative in the same spirit: extending `getOSTag`'s existing `ol`/`rocky`/`rhel` normalization to map `centos` → `rhel`. For context on compatibility: OKD is the community OpenShift distribution and the operator's OpenShift/DTK integration is active on these nodes (not a standalone CentOS host), and we build against the published `-rhel10` driver images on SCOS today (digest-pinned `580.126.20-rhel10.0`, kernel `6.12.0-233.el10`) with the kmod building and loading fine. Happy with either shape — whichever better fits the project's support model.

Related: [#1533](https://github.com/NVIDIA/gpu-operator/issues/1533) (original ask, resolved via the digest workaround, closed stale) · [#2811](https://github.com/NVIDIA/gpu-operator/issues/2811) (downstream: repoConfig in DTK container) · [gpu-driver-container#984](https://github.com/NVIDIA/gpu-driver-container/issues/984) / [#985](https://github.com/NVIDIA/gpu-driver-container/issues/985)

**Environment:** GPU Operator v26.3.3 · OKD 4.22.0-okd-scos.6 · CentOS Stream CoreOS 10, kernel 6.12.0-233.el10.x86_64 · NVIDIA L4 · amd64

dundysm:
## Description

OKD/SCOS nodes report NFD `feature.node.kubernetes.io/system-os_release.ID=centos` (often with `VERSION_ID=10`). `getOSTag()` / `getGPUNodeOSInfo()` previously built a `-centos10` driver image suffix, which is not published on nvcr.io, causing ImagePullBackOff.

RHEL clones (`ol`, `rocky`, `rhel`) already normalize to a major-only suffix. This change extends that path for `centos` and maps the image ID to `rhel`, matching the published driver tag family used successfully on SCOS today (for example `-rhel10`).

The NFD-reported release ID remains `centos` for repo/cert/subscription path selection; only the driver image OS tag is remapped.

A CRD-level `spec.driver.osTag` override was considered but deferred as a larger follow-up; this keeps the fix small and aligned with existing rhel-family handling.

Fixes #2892

## Checklist

## Testing

```bash
go test ./internal/state/ -count=1
go test ./controllers/ -count=1
```

Both packages passed locally.

## Notes for reviewers

cc @cdesiniotis @karthikvetrivel @rahulait @rajathagasthya @shivamerla @tariq1890


dundysm:
@cdesiniotis @tariq1890 @shivamerla friendly ping when you have a moment; maps centos/OKD/SCOS driver os tags to published rhelN (fixes #2892). go test ./internal/state and ./controllers green; no crd field in this pr (override can be a follow-up).

controllers/state_manager.go
```diff
@@ -600,12 +600,17 @@ func (n *ClusterPolicyController) getGPUNodeOSInfo() (string, string, error) {
 	if !ok {
 		return "", "", fmt.Errorf("unable to retrieve OS version from label %s", nfdOSVersionIDLabelKey)
 	}
-	// If the OS is RockyLinux, Oracle Linux or RHEL, we will omit the minor version when constructing the os image tag
+	// If the OS is RockyLinux, Oracle Linux or RHEL, we will omit the minor version when constructing the os image tag.
+	// CentOS Stream / SCOS (OKD) reports ID=centos, but published driver images use the rhel tag family.
+	osTagName := osName
 	switch osName {
+	case "centos":
```
kvalliyurnatt, line 607: @dundysm thanks for your contribution!

I think we don't want to map centos to rhel in our code this way ( while this may work fine in the case of OKD)
1. We never supported CentOS
2. There is no guarantee that a rhel image will always work smoothly on centos

So doing this mapping in the code feels wrong, if you do want to do this you can duplicate images, put it in a public repository and override the driver image to point to yours, I don't think we want to embed this logic in code

dundysm: thanks @kvalliyurnatt; that makes sense. i'll close this pr rather than embed a centos→rhel mapping.

for okd/scos the path is overriding the driver image (or digest pin) instead of teaching the operator that centos is rhel. appreciate the clarification on the support model.

dundysm:
closing per @kvalliyurnatt's review: we shouldn't map centos→rhel inside the operator. okd/scos users should override the driver image (or pin a digest) rather than rely on an auto alias.

thanks for the look.

controllers/state_manager.go
aqeelat: @kvalliyurnatt would you be open to allowing the user to choose the suffix of the image?

kvalliyurnatt: @aqeelat what would that look like, also can you elaborate the use case for that ? what would that solve that can't be solved with just overriding the whole image ?

aqeelat:
Both good questions.

**What it would look like:** one optional field on the driver spec, e.g. `spec.driver.osTag: rhel10.0`. When non-empty, the operator uses it in place of the node-derived OS tag when rendering the driver image reference (`<repository>/<image>:<version>-<osTag>`); when empty, behavior is exactly as today. No per-distro mapping table — just precedence of explicit field over derived value. (On mixed-OS clusters a single string would collapse node pools — a per-pool shape could be a follow-up if that matters; OKD/SCOS clusters are single-OS, so the simple form covers the case.)

**What it solves vs whole-image override:** the suffix is appended to `version` regardless of `repository`/`image` — `GetImagePath()` only skips the `-<osVersion>` append when the spec contains a digest (`api/nvidia/v1alpha1/nvidiadriver_types.go`). So the mirror-and-override path doesn't actually escape the suffix: we'd have to duplicate every driver release into our own registry *re-tagged as `<version>-centos10`* and keep that mirror in sync forever. The other escape — digest pinning — freezes the reference: rebuilds never flow under a digest, and we've since been dead-ended by a tag removal from nvcr.io plus a separate bug blocking the newer images (the chain is documented in #2892).

With an `osTag` override we'd consume `nvcr.io/nvidia/driver:<version>-rhel10.0` directly — no mirror, no re-tagging, and version bumps become a one-field edit as NVIDIA publishes tags.

From a support-model perspective it's also the same trust level as the two existing escape hatches — digest pinning and repository override are both admin-owned choices — and it adds no in-tree claim that "centos is rhel" (the concern on this PR): the operator stays agnostic, the admin opts into a suffix and owns compatibility. For what it's worth as a data point: we run a digest-pinned `-rhel10.0` image on SCOS kernel `6.12.0-233.el10` today and the kmod builds and loads fine.

Happy to hear if a different shape would fit better.

kvalliyurnatt:
@aqeelat thanks for the detailed response. I understand that updating needs a manual digest update when you go down the digest pinning route, but that is the officially recommended method, plus since you are running on an unsupported version of OS, it feels driver updates should be validated than having them be automatic, so I think this does not justify enough for us to adding a new osTag field for now. I think I would recommend using the digest pinning method for your use case.

PR #2918, opened 2026-09-17
fix: preserve DCGM service compatibility

### LINKED ISSUE #2917: [Bug] Preserve stable DCGM discovery across ClusterPolicy and GPUCluster deployments
## Problem

The `GPUCluster` DRA stack uses different Service names and pod labels from the existing `ClusterPolicy` stack.

ClusterPolicy uses:

- DCGM Service: `nvidia-dcgm`
- DCGM pod label: `app=nvidia-dcgm`
- Exporter Service: `nvidia-dcgm-exporter`
- Exporter pod label: `app=nvidia-dcgm-exporter`

GPUCluster uses:

- DCGM Service: `nvidia-dcgm-dra`
- DCGM pod label: `app=nvidia-dcgm-dra`
- Exporter Service: `nvidia-dcgm-exporter-dra`
- Exporter pod label: `app=nvidia-dcgm-exporter-dra`

The established DNS names and labels are used by downstream services, runbooks, readiness checks, and monitoring configurations. Consumers must now know which allocation stack is active or probe both identities.

This will become more complicated when DRA and non-DRA nodes are supported in the same cluster.

## Expected behavior

GPU Operator should provide allocation-mode-independent discovery for DCGM and DCGM Exporter.

Consumers should be able to:

- Use a stable DNS name without detecting the allocation mode.
- Discover either implementation with a common component label.
- Continue using the DRA-specific DNS names where they have already been adopted.
- Avoid duplicate Prometheus scraping when both Service aliases exist.

DaemonSet names, selectors, and `app` labels can remain implementation-specific.

## Proposed common labels

Add the following label to both DCGM DaemonSet metadata and pod-template metadata:

`nvidia.com/gpu-operator.dcgm: "true"`

Add the following label to both DCGM Exporter DaemonSet metadata and pod-template metadata:

`nvidia.com/gpu-operator.dcgm-exporter: "true"`

Apply these labels to both:

- The ClusterPolicy manifests under `assets/state-dcgm*`
- The GPUCluster manifests under `manifests/state-dcgm*`

These labels should be reserved so user-provided DaemonSet labels cannot override them.

## Proposed DNS compatibility

For the initial implementation:

The legacy ClusterPolicy stack should continue exposing:

- `nvidia-dcgm`
- `nvidia-dcgm-exporter`

The DRA GPUCluster stack should expose:

- `nvidia-dcgm`
- `nvidia-dcgm-dra`
- `nvidia-dcgm-exporter`
- `nvidia-dcgm-exporter-dra`

This restores the established DNS names for DRA deployments without unnecessarily exposing DRA-specific names from the legacy stack.

## Upgrade safety

The initial implementation should not immediately change existing Service selectors to the new common labels.

During an upgrade, existing pods do not receive the new labels until their DaemonSets roll out. If a Service selector changes before that rollout completes, the Service can temporarily lose all or some endpoints.

This is particularly disruptive for DCGM because its Service uses `internalTrafficPolicy: Local`.

For the initial phase:

- Existing ClusterPolicy Services should retain their existing `app` selectors.
- Existing DRA Services should retain their DRA-specific `app` selectors.
- New compatibility aliases in the DRA path should initially select the DRA-specific `app` labels.
- Common Service selectors should be introduced only after labeled pods have rolled out.

## Future coexistence and shared ownership

When ClusterPolicy and GPUCluster are allowed to coexist, the stable Services will become shared resources.

The current reconciliation paths cannot safely co-own these Services:

- Both use controller owner references, but Kubernetes permits only one controller owner.
- Both perform whole-object updates that can overwrite owner references.
- ClusterPolicy directly deletes disabled Services.
- GPUCluster cleanup deletes objects by state label.
- Deleting either CR could remove a Service still needed by the other stack.

Simply replacing `SetControllerReference` with `SetOwnerReference` is insufficient.

A future phase should introduce a single shared-Service reconciler that:

- Computes desired Services from both CRs.
- Maintains deterministic non-controller owner references.
- Removes only the disabled or deleted CR's ownership.
- Deletes a Service only when no configuration requires it.
- Preserves API-assigned Service fields.
- Resolves conflicting exporter Service configurations deterministically.

Shared ownership must be implemented before ClusterPolicy and GPUCluster coexistence is enabled.

## Monitoring considerations

If both exporter Service aliases select the same pods, monitoring must avoid scraping every exporter twice.

The implementation should define one canonical scrape Service, preferably `nvidia-dcgm-exporter`, and ensure:

- Only the canonical Service participates in annotation-based discovery.
- Operator-managed ServiceMonitors select only the canonical Service.
- Conflicting ServiceMonitor configurations are reported instead of using last-writer-wins behavior.

## Acceptance criteria

- [ ] Both DCGM implementations carry `nvidia.com/gpu-operator.dcgm="true"`.
- [ ] Both exporter implementations carry `nvidia.com/gpu-operator.dcgm-exporter="true"`.
- [ ] Common labels exist on DaemonSet metadata and pod templates.
- [ ] User-provided labels cannot override the common labels.
- [ ] A DRA deployment exposes both established and DRA-specific DNS names.
- [ ] The legacy stack does not expose DRA-specific DNS names.
- [ ] Existing Service selectors are not changed before labeled pods roll out.
- [ ] Both DCGM DNS names preserve `internalTrafficPolicy: Local`.
- [ ] Exporter aliases preserve supported Service configuration.
- [ ] Exporter pods are not scraped twice.
- [ ] Regression tests cover both stacks and the upgrade transition.
- [ ] Shared ownership is implemented before ClusterPolicy and GPUCluster coexistence is enabled.

rahulait:
## Description

Partially fixes https://github.com/NVIDIA/gpu-operator/issues/2917

Restore stable DCGM and DCGM Exporter identities for GPUCluster/DRA deployments while retaining the released DRA-specific Services.

## Changes

- Add DRA compatibility Services:
  - `nvidia-dcgm`
  - `nvidia-dcgm-exporter`
- Retain:
  - `nvidia-dcgm-dra`
  - `nvidia-dcgm-exporter-dra`
- Add common labels to DaemonSets and pod templates:
  - `nvidia.com/gpu-operator.dcgm: "true"`
  - `nvidia.com/gpu-operator.dcgm-exporter: "true"`
- Prevent ClusterPolicy custom labels from overriding the common labels.
- Preserve Kubernetes-assigned Service fields during GPUCluster updates, including ClusterIPs and NodePorts.
- Annotate only `nvidia-dcgm-exporter` for Prometheus scraping to avoid duplicate targets.
- Add tests and update golden manifests.

## Checklist


internal/state/state_skel.go
```diff
@@ -442,9 +443,86 @@ func (s *stateSkel) mergeObjects(updated, current *unstructured.Unstructured) er
...
 	}
+	if gvk.Group == "" && gvk.Kind == "Service" {
+		return s.mergeService(updated, current)
+	}
+	return nil
+}
+
+// mergeService preserves fields allocated by the API server. Clearing these fields in a full
+// update is rejected because their values are immutable or must remain allocated.
+func (s *stateSkel) mergeService(updated, current *unstructured.Unstructured) error {
```
cdesiniotis, line 454: Question -- why is this needed now when it wasn't needed before? Is it because this commit is introducing a diff in the `nvidia-dcgm-exporter-dra` Service object?

rahulait: This was flagged by LLMs. I believe we had this bug introduced recently but never caught it.

non-dra path uses:
https://github.com/NVIDIA/gpu-operator/blob/93c0ff560ca71e761d6c1782d904765bb88de3db/controllers/object_controls.go#L4952-L4956
It sets resourceVersion and clusterip allocated before calling update().

However, GPUCluster path uses generic unstructured reconciler which only preserves resourceVersion but not the clusterIP. Kubernetes can reject the update if clusterIP is omitted. This might be an existing bug in v26.7.0 not yet reported. Once service is created and ip is assigned, future reconcile updates might be failing but its not getting affected as it exists, just not getting updated.

assets/state-dcgm-exporter/0800_daemonset.yaml
```diff
@@ -3,6 +3,7 @@ kind: DaemonSet
 metadata:
   labels:
     app: nvidia-dcgm-exporter
+    nvidia.com/gpu-operator.dcgm-exporter: "true"
```
tariq1890, line 6: Can we look at alternate names for this label key?

The format of `nvidia.com/gpu-operator.dcgm-exporter: "true"` seems non-standard for labels.

cdesiniotis: Maybe `app.kubernetes.io/component: nvidia-dcgm-exporter`?

tariq1890: Yes, that could work. It is a standardised label

rahulait: Sure, we can. This label was selected so that we use the same scheme we used before like here: https://github.com/NVIDIA/gpu-operator/blob/93c0ff560ca71e761d6c1782d904765bb88de3db/internal/consts/consts.go#L71

tariq1890: Looks like `app.kubernetes.io/component` has a specific meaning as per this [K8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)

How about `app.kubernetes.io/name` instead?

tariq1890: @rahulait I missed to mention this earlier, using a boolean as the label value typed seemed a bit off to me.

PR #2922, opened 2026-09-18
Honor GPUCluster priority classes in monitoring and validation

ajavanma:
## Description

Setting `GPUCluster.spec.daemonsets.priorityClassName` left DCGM, DCGM exporter, and DRA validator at `system-node-critical`, although their render data already carried the requested value. Honor the configured name in these three DaemonSets, preserving `system-node-critical` when unset.

Quote the resolved value so names such as `on` and `null` remain strings through YAML decoding. Cover ordinary custom names, an unset value, a YAML boolean keyword, and a YAML null keyword through the real renderer for all three operands (12 cases). Existing golden manifests remain unchanged.

## Checklist

## Testing

- Original custom-name regression: all three custom cases failed on the upstream base and passed with the initial fix.
- Quoting regression: before quoting, all six `on`/`null` cases failed while the six existing custom/default cases passed. With quoting, all 12 cases pass through real rendering and typed DaemonSet conversion.
- `make fmt license-check`: passed.
- `make coverage`: passed, including build and the full unit suite (21 tested packages); existing golden manifests unchanged.
- `make lint`: passed with zero issues.
- `make validate-modules`: passed for the source module, vendor tree, and tools module.
- `make validate-generated-assets`: passed; no generated-file drift.
- `make check-third-party-notices`: passed for all 123 packages on Linux AMD64 and ARM64; notices unchanged.
- Current local environment: Linux ARM64, Go 1.27.1 and golangci-lint 2.13.1 matching CI pins.
- No live Kubernetes/GPU or end-to-end scheduling validation performed.


myeolenv:
Thanks for the contribution. We will review this PR soon.

PR #2941, opened 2026-09-22
chore: [CODEOWNERS] use gh teams instead of usernames

tariq1890:
Instead of listing every GitHub username as an authorised code owner, we use a GitHub team instead. This has two advantages:

- No need to update CODEOWNERS when members are added or removed
- It keeps the added reviewer list for PRs minimal


rahulait:
~Does it auto-detect who the users are in that mailing list?~
nvm, I see the team now

rahulait (approved):

